PR #56, opened 2022-08-07
Spell checked the public pages

niktek:
Hi,

Decided to do some spell checking while reading the docs!  There's also an add to the .gitignore for pnpm users.

Cheers,
Nik


endigo9740:
This is spectacular @niktek thanks so much for doing this! This was one of like a hundred items on my agenda, so thanks for making my life a bit easier :D

PR #62, opened 2022-08-09
docs: update create command

oekazuma:
`npm init svelte@next my-app` has been changed to `npm create svelte@latest my-app`.

Reference: https://kit.svelte.dev/


endigo9740:
@oekazuma great catch, thanks for this!

PR #66, opened 2022-08-09
Added copy code button to CodeBlock - not for merging

niktek:
Added 'hidden' prop to Button to assist with auto show/hide
Fixed 'undefined' being output to the class attribute on Button when no class was set on original component.

It shows on rollover of the CodeBlock, updates to reflect user has clicked it and then reverts after 3 seconds.

Mobile is being a bit pesky at the moment - on iOS, the click will not change the colour, yet it will copy the code....  It also causes the page to reload and go to the top....  so if I could get some pointers on this one, much appreciated.

Also, despite everything looking beautiful in VS Code, it clearly has munged up the tab/spacing from that of the project.  If you could let me know what settings/extensions/tools the project is using, then I will update and re-send the PR properly.

### DIFF
--- a/src/lib/Button/Button.svelte
+++ b/src/lib/Button/Button.svelte
@@ -11,6 +11,7 @@
     export let weight: string = 'ring-1';
     export let width: string = 'w-auto';
     export let rounded: string = 'rounded-lg';
+	export let hidden: boolean = false;
 
     // Set tag and href values
     const tag: string = $$props.href ? 'a' : 'button';
@@ -97,8 +98,9 @@
 
 <svelte:element
     this={tag}
-    class="comp-button {classesButton} {$$props.class}"
+	class="comp-button {classesButton} {$$props.class ? $$props.class : ''}"
     {href}
+	class:hidden={hidden}
     data-testid="comp-button"
     on:click
     {...prunedRestProps()}
--- a/src/lib/CodeBlock/CodeBlock.svelte
+++ b/src/lib/CodeBlock/CodeBlock.svelte
@@ -1,22 +1,52 @@
 <script lang="ts">
     import hljs from 'highlight.js';
-    
+	import Button from '$lib/Button/Button.svelte';
+
     export let language: string = 'plaintext';
     export let code: string = '';
+	export let variant:string = 'ghost';
+
+	let hideCopyButton: boolean = true;
+	let toggleCopyButtonVisibility = () => { hideCopyButton = !hideCopyButton}
+	let computedVariant:string = variant
+	let showCopiedMessage:boolean = false;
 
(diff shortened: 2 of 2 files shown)


endigo9740:
Thanks for sending this as a proof of concept Nik. I think there's some value here but my concern is changes may be a bit far reaching.

For example, the addition of the `hidden` prop on the Button component probably isn't required. All Skeleton components provide a built-in `class` prop for arbitrary class styles. Something like this:

```
<Button variant="ghost" class="hidden">Skeleton</Button>
```

Honestly I have mixed feeing about the copy button using reveal on hover. We always want to strive for responsive support and accessibility by default, so I might suggest the button become static element so that it supports devices without a mouse (mobile, etc).

Additionally, we have a very specific style for how we define and apply classes using "base" and "reactive" patterns. It's kind of unique, so there's a learning curve, but I'd definitely suggest you review other components to see what I mean. See also the "Component Styling" section on the contribution guide here:

Finally, I have some concerns with the size of the button. I might suggest we use the `variant="text"` style to make the background transparent by default. Then use an embedded SVG icon. Something like what Mantine does here:

https://mantine.dev/core/accordion/

Screen Shot 2022-08-09 at 10 49 09 AM

The only thing I have mixed feelings on would be the introduction of the tooltip popover. It seems like unnecessary flashiness where a standard `title` attribute would probably suffice.

However, I do like how you're using the native copy to clipboard with `navigator.clipboard`. Native is always best!

Per your request, I'll go ahead and reject the PR. Let me know if I can expand on any of the points mentioned above. But all the same I appreciate the effort! Thanks!

niktek:
I knew there was going to be design language and code patterns to get up to speed with, so I really appreciate the time taking to provide the feedback!

endigo9740:
No worries, it took Thomas and myself a while to settled into things as well. When we started there really wasn't a precedence for handling Tailwind styling within Svelte, so we've defined that as we've gone. If I can clarify anything just let me know.

Also, per our roadmap discussion today, we noted that libraries like Mantine implement a general purpose "copy button" component. Mantine has been a HUGE source of inspiration when designing the current components. This might be a better way to approach this feature:

https://mantine.dev/core/copy-button/

Things to note:

* Naming schemes - props, verbiage, etc
* Features - what options do they provide?
* Simplicity - let's start simple and build up, don't need every feature out of the gate

General idea is this is created as a dedicated component, then we turn around and embed it into the CodeBlock.

thomasbjespersen:
It literally took us months to decide on certain conventions, with a lot of back and fourth for some component development, so yeah, definitely still appreciated. We have plans for expanding our contribution guide(s), as it is currently a little lean in certain areas.

PR #73, opened 2022-08-11
Added custom typescript exports on package+docs

niktek:


endigo9740:
@niktek Hey, so I pulled down the PR and tested locally. I felt you were on the right track with this but there had to be a simpler way to do this. Sure enough I've found one!

Basically instead of having to create a dedicated types file (custom.d.ts) and having to create a script to do al this merging and such, we can actually just rename the `index.js` file to `index.ts` (typescript format) and then drop our type exports straight in:

> NOTE: ignore all the red squiggles here, they don't impact how this works. The paths are relative to the package, not where the file exists as is.

Screen Shot 2022-08-13 at 2 02 47 PM

The result is EXACTLY the same! Even tested within a local project and it worked perfectly.

Given your PR comes from a fork, I cannot unfortunately modify it directly afaik, so I'm going to reject this PR and request you instead follow my lead and make this same change. Just make sure to update the Toast/Dialog docs like before!

It's almost silly how simple the fix is in hindsight, but I think it's due to gaps in the SvelteKit docs, not on us. I might even suggest we put in a PR to the SvelteKit docs to make this process more clear! Help them and everyone else out!

PR #74, opened 2022-08-11
Removed deprecated and updated vitest

niktek:
This seems to be clearing all the fresh install issues for me.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,10 +18,10 @@
 		"coverage": "vitest run --coverage"
 	},
 	"devDependencies": {
-		"@sveltejs/adapter-auto": "^1.0.0-next.39",
-		"@sveltejs/adapter-static": "^1.0.0-next.29",
+		"@sveltejs/adapter-static": "1.0.0-next.39",
 		"@sveltejs/kit": "^1.0.0-next.322",
 		"@tailwindcss/forms": "^0.5.2",
+		"@testing-library/dom": "^8.17.1",
 		"@testing-library/svelte": "^3.1.1",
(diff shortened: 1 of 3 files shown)


endigo9740:
I had suspected the same on this, but unfortunately your changes causes a lot of errors in the test runner. Almost half the tests fail to load after this change:

Screen Shot 2022-08-13 at 2 14 33 PM

Screen Shot 2022-08-13 at 2 15 44 PM

Again, I do believe it should be possible to drop this extra dep. In fact, when we implemented Vitest it was early days, and very rough around the edges (back in February). But they've made a lot of progress since then and I think we need to ensure Vitest a conforms to the latest configuration requirements.

I've included this on the Roadmap, but one thing I've noted is we may be making extra work for ourselves by having to declare JSDOM in every single test file. It looks like you can now configure this globally:
https://vitest.dev/guide/features.html#mocking

I'm sure there's other issues like this, so my goal is to almost reboot and reinstall Vitest from the ground up. Go test-by-test and ensure there's no extra "fat". This may be a heavy lift, so no worries if you want to leave this for Thomas and I post-a11y update!

endigo9740:
If you can find a solution for the issues above I'll welcome a follow up PR though.

niktek:
Pretty sure this is a case of not doing an `npm i` or `pnpm I` after merging in.  Tested with both, all tests run fine, no regressions on docs/components

niktek:
This is being rolled up into another PR that upgrades to Svelte Kit 408

PR #75, opened 2022-08-11
Unset class prop will emit undefined

niktek:
Most buttons currently emit undefined as the last class if they do not have a class explicitly defined on them in the source.

### DIFF
--- a/src/lib/Button/Button.svelte
+++ b/src/lib/Button/Button.svelte
@@ -97,7 +97,7 @@
 
 <svelte:element
     this={tag}
-    class="comp-button {classesButton} {$$props.class}"
+    class="comp-button {classesButton} {$$props.class ? $$props.class : '' }"
     {href}
     data-testid="comp-button"
     on:click


thomasbjespersen:
@endigo9740 - If I am not mistaken, this has happened for other components?

thomasbjespersen:
> Most buttons currently emit undefined as the last class if they do not have a class explicitly defined on them in the source.

@niktek If you could provide more information on this issue, that would be great. Possibly a video or a few screenshot?

niktek:
Sorry, should have checked more widely.  Yep it comes through on most components - just right click->inspect and you will see the last class is undefined.

image

thomasbjespersen:
You're all good @niktek! Thanks for bringing it to our attention. @endigo9740 - I doubt we have any tests in place for this as well? Along with the inconsistency with restProps, the fix(es) should probably be consolidated?

endigo9740:
@niktek @thomasbjespersen 

I've experienced this and fixed it in a few places. It is something that needs to be addressed globally though. It's due to `$$props.class`. We utilize this to pass through custom/arbitrary styles on every component. However, if a component is not providing a `class` attribute, then the result appears to be a string of "undefined" rather than a true `undefined` symbol.

@niktek I'm going to reject, for two reasons:

1. We need to address this in a global pass through all components. We'll need to create an issue to track progress on this.
2. We should opt for a simpler syntax rather than an inline ternary expression.

Per item 2, I have a couple proposed ideas:
- Use an inline "or" operator: `{$$props.class||undefined}`
- Strip and empty class prop in the `<script>` tag:
```
if ($$props.class === undefined)  { delete $$props.class }
```

Personally I like this better. It's more verbose, but doesn't need to be crammed into the HTML template! It also helps set a proper convention for handling/stripping $$prop values as needed (we're doing something similar for conflicts with $$restProps in 4 components currently.

thomasbjespersen:
@endigo9740 - Is this tested to be able to close this issue yet for all components? It is widespread...

endigo9740:
Close = rejected. It won't be merged. I've made my suggestion above. Let's create a Issue for this, document the issue. Then we can discuss proposals for solutions there. Will be more apt than here!

thomasbjespersen:
@endigo9740 - Lets create a new issue for the suggestions..

PR #77, opened 2022-08-11
Pastable functional demo

niktek:
I had an error in the previous docs.  This example is now fully copy pastable with a note about the breakpoint.  Also removed the Dividers component to shrink the code up a bit.

### DIFF
--- a/src/routes/components/drawers.svelte
+++ b/src/routes/components/drawers.svelte
@@ -57,29 +57,32 @@
     <section class="space-y-4">
         <h2>Usage</h2>
         <CodeBlock language="typescript" code={`
-import type { Writable } from 'svelte/store';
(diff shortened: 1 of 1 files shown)


endigo9740:
Looks good Nik, a quick stylistic note for the docs though - when we have imports and declarations let's try to separate them in separate code blocks like this:

```
import { writable, type Writable } from 'svelte/store';
import { Drawer, Button } from '@brainandbones/skeleton';
```

```
const drawer: Writable<boolean> = writable(false);
const drawerOpen: any = () => {drawer.set(true);};
const drawerClose: any = () => {drawer.set(false);};
```
We've used this pattern in other places and I think it helps make things visually less cluttered. Easier for the end user to read and parse! Completely understand they all go into the same `<script>` tag though!

PR #81, opened 2022-08-13
Added NavItem component export

mstibbard:
Following along with the [Menu](https://skeleton.brainandbonesllc.com/components/menus) documentation and I noticed the `NavItem` component wasn't being exported by default.


endigo9740:
This is a great catch @mstibbard but unfortunately ill timed! The next major version that is nearing release will actually drop NavItem completely! Everything is getting merged into ListItem to simply things. This will come as part of the a11y improvements detailed here:

https://github.com/Brain-Bones/skeleton/issues/41

See the notes regarding updates to the List component, and by all means, feel free to test this out yourself on the `feat/a11y` branch!

So sorry for the trouble, but I appreciate the effort all the same!

PR #84, opened 2022-08-14
Scheme and theme changer

KryptoPX:
Idea (It's working, but i never documented nothing :sweat_smile:) Maybe i can pull request and any of us can maybe use them??
It's not well documented because my english is very poor sorry (I will try)

If there has errors please tellme with comments and i will try to fix them. I'm learning github

chrome_UFkYY2tJ7O


endigo9740:
Hey Krypto, first off thanks for the PR. I appreciate any contribution big or small. A lot of your ideas here align with what we want to do per the roadmap:

https://github.com/Brain-Bones/skeleton/wiki/%F0%9F%9B%A3%EF%B8%8F-The-Skeleton-Roadmap-(proposal)

Unfortunately I don't think this one I can merge in the current state. Without discussing some of these ideas with us beforehand, you've taken some liberties that may not align with a few of the project's goals. Specifically:

* The LightSwitch is due for a refactor and cleanup, but your version changes from providing three options (light/dark/system) to a binary (light/dark). Users lose a very important mechanic this way.
* I would really like to introduce a theme picker like this, but it should be separate from the LightSwitch. I've provided a wireframe mock of the best position in the image below this list.
* We have plans to migrate the themes to the `/src/lib/` directory so that users may import the styles directly into their projects from the package, rather having to import these themselves, so the `/static` directory is not ideal. However we need update the documented instructions in Guides -> Themes alongside this change, so users know how this works. Probably best for this to be a dedicated update on a dedicated branch/PR.
* I've also noted at least one `console.log` remaining in the code. Logs should always be pruned from the code before merged into a production environment.

Screen Shot 2022-08-14 at 11 45 57 AM

We've outlined the process and provided a place to [discuss larger updates here](https://github.com/Brain-Bones/skeleton/discussions/79). Going forward, it may be best to reach out and review your ideas here before you spend the time implementing. Just to help save everyone's time and ensures the requirements are set before you write your first line of code! This ensures we can and will merge proposed features and updates! :)

If you have any questions, please feel free to follow up on this page. I'll see your replies and can respond asap. Thanks!

PR #94, opened 2022-08-15
Bugfix/import of highlightjs as esm module - preview only

niktek:


niktek:
Decided to remove this from the exported package and only use as an dev-dependency for now

PR #95, opened 2022-08-15
Shell script for quickly creating a Skeleton app

niktek:
More for contributors/testers - but makes it much simpler to setup a base Skeleton app for testing.

I know the end goal would be more along the lines of a plugin for svelte-add, but there is a lot of complexity involved in doing so as it uses a full on AST to properly insert it's changes rather than the naive string manipulation this script does.  But this definitely helps in the short term.

### DIFF
--- a/scripts/skel
+++ b/scripts/skel
@@ -1,0 +1,36 @@
+#!/bin/sh
+# Usage: skel my-new-app
+# Auto-runs Svelte-Kit + Tailwind + Skeleton setup to make it a one liner for setting a up a fresh Skeleton app.
+# Either call this script directly, or preferably drop it into a bin directory that is in your $PATH. You need jq, pnpm installed as well
+# Most likely you should tweak the options depending on what you are testing.
+
+npm init @svelte-add/kit@latest $1 -y -- --with typescript+eslint+prettier+tailwindcss --install --tailwindcss-forms --tailwindcss-typography
+cd $1/
+echo "$( jq '.devDependencies += {"@brainandbones/skeleton":"*"}' package.json )" > package.json
+cat <<EOT > tailwind.config.cjs
+const typography = require('@tailwindcss/typography');
+const forms = require('@tailwindcss/forms');
+const skeleton = require('@brainandbones/skeleton/tailwind.cjs');
+
+const config = {
+	content: ['./src/**/*.{html,js,svelte,ts}', './node_modules/@brainandbones/skeleton/**/*.{html,js,svelte,ts}'],
+
+	theme: {
+		extend: {}
+	},
+
+	plugins: [forms, typography, skeleton]
+};
+
+module.exports = config;
+EOT
+curl -s -o src/theme.css https://raw.githubusercontent.com/Brain-Bones/skeleton/master/src/themes/theme-skeleton.css
+cat <<EOT > src/routes/__layout.svelte
+<script>
+	import '../theme.css'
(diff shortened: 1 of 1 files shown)


endigo9740:
Whoa, missed this earlier! Really, really dig the concept! Automation is always a value add. Some feedback assuming we evolve this a bit:

- I have mixed feelings about embedding this inside the project right now. If it's meant for contributors/testers perhaps we host externally and distribute on-demand? We may create extra support cases for ourselves if it gets stale.
- I'm sure you're aware but it's a bit rigid. I don't typically need/use Tailwind Forms/Typography if I'm just wanting to spin up a quick Skeleton project for testing purposes. May be worth cutting those?
- I have two onboarding-focused ideas that need to be formalized asap:
    1. I think we should migrate our current `/src/themes` into the `/lib` directly, then provide them as a built-in package exports. So users can import them directly from the package rather than having to go search online. At least for the defaults. This would need a doc update obviously.
    2. It may be possible to add our Tailwind config > Content changes into [our plugin](https://github.com/Brain-Bones/skeleton/blob/master/src/lib/tailwind.cjs). Even if it's a small step, it's annoying to do over and over.

Another recommendation would be to look into what tooling Svelte/Astro/etc use to prompt and onboard via the CLI. Not that we need that YET, but that's the sort of frictionless onboarding I am for us to have by v1.0!

niktek:
1) yeah, I was thinking the same, but:  it doesn't get exported, it's not a package that can be obtained and most consumers of Skeleton will be using the package, so really, it's only contributors that will see this from the repo directly. I can definitely see it moving up a directory if things evolve to a mono repo with multiple sub packages and things like CI/CD, change sets and other tooling sit in the top level.  But that's for later IMHO.
2) I put in the kitchen sink example, comment line 3 explains that it should be altered depending on what you are testing
3.1) agreed
3.2) yeah worth looking into

4) This was the main reason why I had to write this as is.  The svelte-create package has rejected providing command line options to make it installable in a single line with no interactions.  They consider this to be exposing a public API and therefore only allow the guided interface, which is useless for testing.  Which then of course lead to the svelte-add route, but creating a proper plugin there is a considerable task as it parses the typescript and respects previous plugin insertions as well, in an attempt to allow you to add almost all permutations of the plugins it supports.

endigo9740:
1. True, this may be fine then.
2. Also true, commenting is simply enough, quick and painless
3. Ahh so you've looked into some of these other tools. I know Svelte used to use [degit](https://github.com/Rich-Harris/degit) but not sure what they are doing now. Astro's project scaffolding tool was also great testing that today. Would love to know what those are just out of pure curiosity!

endigo9740:
FYI I test 3.2 from above, and it's not working. The content settings for the plugin don't seem to affect the base Tailwind config, unless I'm doing something wrong.

Tested with both:
```
module.exports = {
    content: [
        './node_modules/@brainandbones/skeleton/**/*.{html,js,svelte,ts}'
    ],
    // ...
}
```

As well as:
```
module.exports = {
    extend: {
        content: [
            './node_modules/@brainandbones/skeleton/**/*.{html,js,svelte,ts}'
        ],
        // ...
    }
}
```
With either approach the Skeleton component styles were not being detected, so all I see is unstyled HTML markup.

endigo9740:
@niktek Merging this, just FYI this will be the last PR I accept targeting `master`.

Issue #100, opened 2022-08-16
Update package.json to use default Vite dev server port


endigo9740:
The old default port for SvelteKit's dev server was specified as `3000`:

```json
"scripts": {
    "dev": "vite dev --port 3000",
}
```

However, in new SvelteKit projects the port is no longer provided, meaning it uses Vite's default of `5173`

```json
"scripts": {
    "dev": "vite dev",
 }
```

You can then open the local dev site at:

```
http://localhost:5173/
```

I don't believe we document the port anywhere, so this should just be a small one-line change!

PR #105, opened 2022-08-17
Feat/use the new dynamic port of vite

niktek:
Just FYI to people, if you have multiple instances, it will auto port hunt upwards to a non used port


niktek:
This closes #100

PR #106, opened 2022-08-17
Closes #100

### LINKED ISSUE #100: Update package.json to use default Vite dev server port

niktek:


endigo9740:
@niktek Looks like this was targeted at Master branch. Gonna take some getting used to to aim for dev, almost did that myself a time or two. Maybe there's a way we can specify a default?

I'll go ahead and accept since it's such a minor change though. Won't affect anyone but contributors.

PR #109, opened 2022-08-17
Chore/migrate to post sk.405

niktek:
Tested with both npm and pnpm.
npm run test works
no visual regressions found on docs site


endigo9740:
Super excited to try this. Will be first thing on my list to test in the morning! Docs will be due for an update alongside this. Off the top of my head:

### Internal/Contributor-Facing:

* Contribution Guide - several path and file references here

### External/Package

* Pretty much every guide section mentions some form of `app.css` or `__layout.svelte` for SvelteKit now
* Some of the Utility Docs did the same prior, but should be more vague now (ex: use your root component)

Good news is the new tabbed UI allows now allows us to differentiate the Vite format from SvelteKit. They matched prior, but now they they don't! We're now prepped for this!

endigo9740:
Working great. Props to the group that built the migration tool. It's working so well!

PR #111, opened 2022-08-17
Chore/formatting codebase

### LINKED ISSUE #70: Linting Pass
As reported here, we're well overdue for a linting pass: https://github.com/Brain-Bones/skeleton/issues/67

I'd advise we handle this immediately following completion and merge of the a11y branch per this ticket: https://github.com/Brain-Bones/skeleton/issues/41

I'm anticipating this to be a lot of small issues (missing semicolon, etc.), so general process should go something like this

* Create the branch following [convention](https://skeleton.brainandbonesllc.com/docs/contributions) (ex: `bugfix/linting`)
* Run the Prettier lint command `npm run lint`
* Address each issue provided one-by-one
* Run automated test via Vitest to check for regressions `npm run test`
* Do a manual pass through the project to check for regressions

> Contact either @thomasbjespersen or myself if you have questions or need assistance

In the future we'll look to make this a regular process and/or automate with a CI/CD pipeline.

niktek:
Finally got there!!

Good thing - there is close to nothing we can configure, due to their philosophy https://prettier.io/docs/en/option-philosophy.html

The main one is `printWidth` in `.prettierrc` which I've set at 200 as it fits two files side by side with spare space and seems to give the nicest formatting as a result.  Easy to change.

Have sprinkled some ignores throughout - the critical one is on the `<pre>` element in the CodeBlock - which in hindsight was really obvious for causing the formatting issues I was seeing.

This will close #70


endigo9740:
Looks good as well. I think I may end up adding/removing some of the prettier formatting in place. The body gradient in the app.css is a prime candidate for example. But overall I think it's good. I'll have to adjust my style to align with this. I tend to like more white space between parent/child HTML elements, but I can live with this. Overall a net gain.

We'll all need to be more diligent about running Prettier often though, since we do not have a dedicated CI/CD system yet. But worse case we can handle this each release or something like that.

Great job Nik!

PR #112, opened 2022-08-17
Feat/alias for docs site to catch import issues

### LINKED ISSUE #103: Add missing component exports in index.js
Looks like we have at least one or two missing exports in `/lib/index.js`, per: https://github.com/Brain-Bones/skeleton/issues/102

So far the only component I know needs to be added is:

* `Step` - a child component for the Stepper component

Please do not add or modify the following:

* `NavItem` - reported missing, but it's being dropped in the next release
* `CodeBlock` - is being removed in the next release.
* `ConicGradient` -a work in progress, not ready for end users
* `ThemeGenerator` - this is only for use in the Docs.

If you're unsure of any others just let me know. I'll clarify asap!

niktek:
Aliases actually do work!

For all of the docs site, we can now use Skeleton exactly as an end developer would via:

```import {Card} from '@braindandbones/skeleton```
Any internal ones like CodeBlock can just be imported from the $lib.  Also comes with intellisense too.

This closes #103 and should go a long way to making sure it doesn't happen for other new components.


endigo9740:
Still a bit on the fence about this one, but I suppose having the option to go relative or alias is valid. We'll give it a go!

Issue #68, opened 2022-08-09
Decouple Skeleton from SvelteKit


endigo9740:
Expand Skeleton's support and instruction for any framework where Svelte and Tailwind can be used. Including, but not limited to:

* ~[Vite](https://vitejs.dev/)~ - via https://github.com/Brain-Bones/skeleton/issues/83 - works with v0.32.19+
* ~[Astro](https://astro.build/blog/astro-1/)~ - via https://github.com/Brain-Bones/skeleton/issues/93 - works with v0.32.19+
* [Routify](https://www.routify.dev/) - (coming soon)
* Client-side only or Micro-Framewoks:
    * Symphony for PHP - via https://github.com/Brain-Bones/skeleton/issues/85 (need more testing on this)
    * (Django/Flask for Python, ..., others?)
* (welcoming other suggestions here)

niktek:
New svelte project with ts, added Skeleton and just tried doing an import of all the components, these are the ones that error out:

`
  import Alert from './lib/Alert/Alert.svelte';  //referes to $lib/Card
  import Avatar from './lib/Avatar/Avatar.svelte';  //refers to $lib/Filters
  import CodeBlock from './lib/CodeBlock/CodeBlock.svelte' // fails to resolve highlight.js
  import Filter from './lib/Filters/filter' //refers to $lib/Filters/svg/* 
  import LightSwitch from './lib/LightSwitch/LightSwitch.svelte'; // refers to $app/env
  import Menu from './lib/Menu/Menu.svelte'; //refers to $app/stores
  import Dialog from './lib/Notifications/Dialog.svelte'; //refers to $lib/Button
  import Toast from './lib/Notifications/Toast.svelte'; //refers tp $lib/Notifications
  import Paginator from './lib/Paginator/Paginator.svelte'; //refers to $lib/Button
  import Step from './lib/Stepper/Step.svelte'; //refers to $lib/Button
  import ThemeGenCustom from './lib/ThemeGenerator/ThemeGenCustom.svelte'; //refers to palettey
  import ThemeGenTailwind from './lib/ThemeGenerator/ThemeGenTailwind.svelte'; //refers to CodeBlock
`

Main issues are $lib reference doesn't exist in Svelte and then we have the trickier ones where functionality is received from $app/stores and $app/env

endigo9740:
[...]

PR #113, opened 2022-08-17
Docs/review guides

niktek:
This updates #68 and makes the guides correct for SK.405+ as well as general wording, typo checks.

I did not do an actual run through of setting up an astro project yet


endigo9740:
Still a bit on the fence about this one, but I suppose having the option to go relative or alias is valid. We'll give it a go!

PR #118, opened 2022-08-18
Chore/extract svg's into components

niktek:
Couple of notes - importing direct from the .svg file is not quite what we wanted.  We would need to either utilise a Vite plugin as mentioned here https://natclark.com/tutorials/svelte-working-with-svgs/ or import them as Svelte components.  I've chosen the latter for now and it all works fine.


endigo9740:
Hmm, Nik I apologize in advanced, this is on me, but I don't think this is the right path. I'm very leery of turning all SVG into components. Let me try to explain...

First off I think we may be overlooking the elephant in the room - that being using .svg and standard img tags. There's really no reason we have to import via Vite at all. In fact we would probably be better served to utilize .svg in the `/static` directory. It's what it's there for! Additionally, in hindsight, perhaps trying to group all SVG in one place is not the right idea. We have two distinct groups:

1. Decorative SVGs use in the docs, these are ideal candidates for `/static`
2. Functional SVGs which illustrate an action within a component (ex: Accordion caret)

Frankly, I just don't think we've settled on a convention that hits that sweet spot yet. Reviewing our options we have:

* Component SVGs - I feel this obscures what each SVG is and may affect how we style them.
* Embedded SVGs - unlimited design, customizable, but affect the readability of the component code.
* External SVGs - better code readability but where do they live? `/lib` directory or more localized?
* Unicode - simplest option, but design is limited, and display is not consistent between operating systems

I'm not going to close the PR yet, but rather investigate a few potential conventions. If I settle one that works really well we can discuss. If we do use this PR, I still may have to ask for your to resubmit for a few reasons:

* The AccordionItem has the import randomly in the middle of the document rather than at the top. Maybe others too?
* The names are descriptive, but should probably be semantically prefaced like: "SvgAlignLeft"
* We would probably want to create a non-listed docs page that visually demos each component. Cearly show what they are. For example, I have no idea the difference between GitHub 1 versus 2 off hand.

That last point is an excellent example of my concern with components, we should not need multiple instances of the same SVG. One of the main benefits of SVG is they can be infinitely resized, recolored, and otherwise customized without distortion. I'm not sure components allow this flexibility without introducing additional wrapping elements. If we use img tags then this comes turn key.

Again I'll investigate and see what I can come up with, but all the same I appreciate the time and effort! I'll feel awful if we do not merge this because I can tell you spent a good amount of time on this. Sorry dude!

endigo9740:
I'll post my findings here:
https://github.com/Brain-Bones/skeleton/issues/121

endigo9740:
Closing in favor of: https://github.com/Brain-Bones/skeleton/pull/122

See https://github.com/Brain-Bones/skeleton/issues/121 for an explanation of how this works.

PR #130, opened 2022-08-20
Formatted and changed tabIndex to tabindex

niktek:
Just ran a format on files in dev so that they don't show up on formatting with other PR's.

It was complaining about `tabIndex` not being a valid attribute for sag, so renamed it to `tabindex` in SvgIcon.svelte


endigo9740:
In my queue to test tomorrow. Good catch! So much a11y stuff lately my brain must be melting!

endigo9740:
FYI I caught one more instance in the SVG Icon documentation tables. That's been added so I'll merge this now.

PR #131, opened 2022-08-20
JS only version of test strip

### LINKED ISSUE #127: Fix regression issue with Vitest tests
See the screenshots below. Will need to research the issue, but most tests are failing.

Screen Shot 2022-08-19 at 3 12 54 PM

Screen Shot 2022-08-19 at 3 13 01 PM

niktek:
Looking at https://kit.svelte.dev/docs/configuration#package

Their docs are wrong - it needs to be nested under kit.package: and the '!**/build.*' exclude is nonsense as they never get sent to that function on either build or package.  Have changed this to a JS only implementation.

This closes #127

### DIFF
--- a/svelte.config.js
+++ b/svelte.config.js
@@ -19,8 +19,13 @@ const config = {
 			assets: 'build',
 			fallback: 'index.html', // index.html (SPA) | null (SSR)
 			precompress: false
-		})
-		// vite: { ...moved to vite.config.js... }
+		}),
+		package: {
+			// strip test files from packaging
+			files: (filepath) => {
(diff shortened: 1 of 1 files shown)


endigo9740:
This sounds like an excellent improvement! Two birds one stone!

That said, I'm going to sounds like a crazy person, but I go out of my way to avoid new dependencies when possible. Id think this would be something the vite/vitest config would handle. Let's take one more good look to see if we can do without the dep, otherwise I'll test and merge tomorrow!

endigo9740:
This is PERFECT, thanks man! Merged!

PR #133, opened 2022-08-20
Removed highlight.js and refactored with prismjs for Codeblock

### LINKED ISSUE #88: Reimplement CodeBlock with adjusted support for Highlight.js
> NOTE: moving this here to track progress

For reference - https://github.com/highlightjs/highlight.js/issues/3527 there is a screen grab showing the exact error we get.  

I've also gone ahead and asked them for any advice https://github.com/highlightjs/highlight.js/issues/3597

_Originally posted by @niktek in https://github.com/Brain-Bones/skeleton/issues/83#issuecomment-1214314346_

luke-hagar-sp:
Fixes #88

I ran tests on what I changed and I validated everything using the Codeblock page on the site and updated the documentation.

I will note that this is my first real commit to any project that isn't mine so please be brutal if I butchered anything

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,7 +8,7 @@
 			"name": "@brainandbones/skeleton",
 			"version": "0.30.6",
 			"dependencies": {
-				"highlight.js": "^11.6.0"
+				"prismjs": "^1.28.0"
 			},
 			"devDependencies": {
 				"@sveltejs/adapter-auto": "^1.0.0-next.64",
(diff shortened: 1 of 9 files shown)


endigo9740:
>  please be brutal if I butchered anything

_gets my axe_ ;)

Seriously that's awesome this is your first contribution, I'm humbled! Knee jerk reaction you're most of the way there but two things:

1. If you add a prop, make sure you add it to the component docs. Otherwise no one will know about it! The next point kind of makes that moot though...
2. We can probably drop the custom prop in favor of `contenteditable={$$props.contenteditable||false}`

> NOTE: this should work in theory but please test. I haven't done it with this prop specifically!

This is great for native passthrough props, but just be cautious of the default value being `undefined`, hence the fallback.

Also major props for updating the tests. That's the part everyone always forgets (myself included!)

That said, I'm going to ping @niktek to help test this. The reason we're removing CodeBlock was not due to the highlighter switch, but because it was causing breaking changes for end users when bundled in our package. This was due to the (ESM?) bundling method used by Highlight.js. Nik can probably explain better. But we need to confirm this does not break things for folks, because that's obviously bad! :)

FYI it's late might time, so I'll aim to test personally tomorrow. Expect more feedback then. That said, great job @luke-hagar-sp Keep up the great work!

luke-hagar-sp:
Thank you @endigo9740, I implemented your suggestions, additionally, I had already corrected the docs and all uses of the Codeblock on the site locally but didn't want to push too much, those are included in an additional Commit.

endigo9740:
Cool, looks good. I believe Nik is giving it a go now. Expect more feedback as either of us complete our reviews! Thanks!

endigo9740:
Seeing issues the break most functionality on the page, so I'm going to have to reject for now. I'll update the ticket with my thoughts.

Screen Shot 2022-08-20 at 12 32 23 PM

Issue #134, opened 2022-08-20
New Library project type from svelte-create


niktek:
There is a new option now for `npm create svelte@next` that adds a library project type.

I've done a diff on the two project structures and the main things we should look at are:

package.json:
`"build": "svelte-kit sync && svelte-package"` which presumably ensures that the types are properly generated before the packaging is done.

tsconfig.json
`"moduleResolution": "NodeNext"` under compilerOptions.

I've done a test of these settings in Skeleton and there doesn't seem to be any issues at first glance.

PR #135, opened 2022-08-20
Settings to match new library from svelte-create

niktek:
Test of new settings as mentioned in #134


endigo9740:
Good catch, merged!

PR #138, opened 2022-08-21
Decoupled Highlight.js from CodeBlock component

endigo9740:
See the attached ticket for details. In short, this drops the Highlightjs dep from CodeBlock, reinstated CodeBlock, update docs. Highlight.js is now an optional progressive enchantment if you choose to add it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,8 +49,5 @@
 		"vitest": "^0.19.1"
 	},
 	"type": "module",
-	"types": "index.d.ts",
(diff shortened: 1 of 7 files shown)


thomasbjespersen:
Let's leave this, but keep this open for QA.

endigo9740:
Yeah this and the other PR aren't going anywhere. When we deploy tomorrow these will still be sitting here for review.

PR #139, opened 2022-08-22
Improvements to the guide and wordsmithing

niktek:
Did a pure npm run through and added a few missing imports and some wording around the examples.

### DIFF
--- a/src/routes/guides/astro/+page.svelte
+++ b/src/routes/guides/astro/+page.svelte
@@ -72,7 +72,7 @@ import BasicLayout from '../layouts/BasicLayout.astro';
 <BasicLayout>
     <h1>Hello Skeleton</h1>
 </BasicLayout>
-            `.trim()}
+`.trim()}
 		/>
 	</section>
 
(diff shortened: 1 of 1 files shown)


endigo9740:
Looks like there's a small mistake here, it's showing the logo imported and used in the "dynamic component" example. The only thing that should be imported and used here is the layout and wrapper components. The other changes look good, if you want to patch that up. I'll keep this PR in place until then, just let me know when the change is ready.

Screen Shot 2022-08-22 at 12 05 21 PM

niktek:
That was added so that you have complete, copy pastable examples to work with.  The static component is never displayed in previous versions.  I've changed a few surrounding words to make that a bit clearer.

endigo9740:
The logo is fine for the static example, no harm there, but I'm referring to the use of the Logo component in the dynamic example. Sorry I missed the heading showing that in the above screenshot.

Plus the Logo component is a child component of the LogoCloud and has no place being using alone, much less on a code sample meant to showcase dynamic components.

endigo9740:
I just noticed the `.astro` extension. I see what you're going for, but I believe this is adding too much to a simple example. We don't need to teach users to how to import Astro components into an Astro project. That's what their docs are for. We need to show them how to use our functional components in Astro.

So dropping those lines mentioned in my previous comment are still valid. The text on the page may need to be altered as well.

niktek:
I think it accomplishes two things:
1) Don't make a user do something if they can't see the results
2) Show the subtle differences in the imports - Astro doesn't even get to multiple frameworks until the end of it's guide https://docs.astro.build/en/core-concepts/framework-components/ - so this helps users get to the action quicker.

I've changed the astro Logo to LogoDisplay to disambiguate from the internal Skeleton Logo.

PR #169, opened 2022-08-30
Adding tab styles

ak4zh:
Similar to: https://flowbite-svelte.com/tabs

Planned styles:

- [ ] full
- [ ] icon
- [ ] pill
- [ ] underline

### DIFF
--- a/src/lib/Tab/Tab.svelte
+++ b/src/lib/Tab/Tab.svelte
@@ -7,6 +7,7 @@
 	export let border: string = getContext('border');
 	export let fill: string = getContext('fill');
 	export let color: string = getContext('color');
+	export let style: string = getContext('style')
 
 	// Props
 	export let value: any = $selected.value;
@@ -26,10 +27,17 @@
 		}
 	}
 
+	function getStyleClass(style: string) {
+		if (style === 'pill') {
+			return `py-2 px-5 rounded-full duration-300`
+		}
+		return ''
+	}
+
 	// Reactive Classes
 	$: isSelected = value == $selected;
 	$: cHighlight = isSelected ? `${border} ${color} opacity-100` : 'border-transparent';
-	$: classesBase = `${cBase} ${cHighlight}`;
+	$: classesBase = `${cBase} ${cHighlight} ${getStyleClass(style)}`;
 	$: classesIcon = isSelected ? fill : 'fill-surface-500';
 </script>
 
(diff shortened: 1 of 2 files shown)


endigo9740:
Hey @ak4zh, this seems like an interesting ideal. I'll pull this down and do a deeper review later today, but a couple quick suggestions.

Consider using `variant` instead of `style` as the prop name. We have precedence for this on the Button component:
https://skeleton.brainandbonesllc.com/components/buttons

Likewise, the way you're adding the style to the reactive class definition works, but to follow convention I suggest something like this:

> NOTE: only including the relevant code here:

```ts
export let variant: string|undefined;

// NOTE: adjust the base to only include universal styles
const cBase: string = 'list-none flex items-center space-x-2 cursor-pointer border-b-2 opacity-60 hover:opacity-100';

// NOTE: Use this to append the extra/unique styles per variant
let cVariant: string|undefined;

function setVariant(): void {
    switch(variant) {
        case ('full'): cVariant = `...`; break;
        case ('icon'): cVariant = `...`; break;
        case ('pill'): cVariant = `py-2 px-5 rounded-full duration-300`; break;
        case ('underline'): cVariant = `...`; break;
        // If no variant is set, handle default styles:
        default: cVariant = 'py-2 px-4 ...'; 
    }
}

$: classesBase = `${cBase} ${cHighlight}`;
```

I'd also advise you look at how we're passing values from the parent TabGroup down to all child tabs using the Svelte Context API (setContext/GetContext). This would allow you to define the variant style once on the parent, and carry through down to al the children!

If you need help with any of this let me know! Glad to do so!

endigo9740:
And if this all seems too complicated please feel free to file a feature request ticket. I'd be glad to take a look at implementing something like this.

endigo9740:
Hey @ak4zh sorry for the delay, it's been a busy week! I finally got a chance to do a more in depth review and have a couple additional notes to add here:

- Make sure you document any changes in the Docs > Tabs section, people need to know how to use the new features
- Obviously verify the test cases continue to work with the additions, feel free to write more test cases to cover additions
- Given the number of large sweeping updates this week your branch is fairly far removed from the `dev` branch changes. Please make considerations for this going forward. 

If you need to drop and recreate the PR and restart against the latest branch updates I'd understand! The merge might be quite difficult otherwise! Thanks

ak4zh:
Yes I will close this PR for now.
I will be able to work on this only after few days so I will base off the latest changes instead.

PR #182, opened 2022-09-04
Code block syntaxhighlight bug

LukeHagar:
This PR is to address a bug that removes syntax highlighting when the content of a CodeBlock is altered, specifically in this case the Skeleton sites Theme Generator

This PR also corrects an issue with HTML when highlight.js is not installed, the HTML needs to be passed directly as text instead of HTML in the component


endigo9740:
@LukeHagar good catch here! I've submitted one small additional commit that makes the conditional for the template agnostic. That way we can use it for any highlighter, without having to create a whole chained if-if/else-else conditional set for every highlighter added in the future. We'll just flip on the `formatted` flag regardless of which highlighter is used.

Thanks for the PR and congrats on now being an official contributor :)

PR #186, opened 2022-09-06
Format pass

niktek:
Format pass, package update and one little Aria label correction in SkeletonAppBar.svelte for the Utilities link


endigo9740:
I promise I'll look into a way to auto-trigger formatting on save soon! This will prevent a lot of this. Looks like a large number of changes, including config files though, so I'll aim to do a quick regression test pass tomorrow before merge!

PR #187, opened 2022-09-06
Removed a11y warning on Gradient Header

### LINKED ISSUE #183: Gradient Heading throws A11y warning when package imported
When creating a SvelteKit project using a tarball (.tgz) version of the package, built against dev, I receive the following A11y warnings when running `npm run start` to start the local dev server:

```
3:22:08 PM [vite-plugin-svelte] /Users/chris/Development/testbed/sveltekit-skeleton-app/node_modules/@brainandbones/skeleton/GradientHeading/GradientHeading.svelte:11:99 A11y: Elements with the ARIA role "heading" must have the following attributes defined: "aria-level"

3:22:09 PM [vite-plugin-svelte] /Users/chris/Development/testbed/sveltekit-skeleton-app/node_modules/@brainandbones/skeleton/GradientHeading/GradientHeading.svelte:11:99 A11y: Elements with the ARIA role "heading" must have the following attributes defined: "aria-level"
```

The Gradient Heading should be updated to adhere to this suggestion and silence this warning.

niktek:
As per https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-level the proper level will be inferred from the `tag` being h1-6 or explicitly defined by the end dev.  Just added an ignore for the warning so as to not be a false positive.

This fixes #183

### DIFF
--- a/src/lib/GradientHeading/GradientHeading.svelte
+++ b/src/lib/GradientHeading/GradientHeading.svelte
@@ -11,6 +11,7 @@
 	$: classesWrapper = `${cBaseHeading} ${direction} ${from} ${to}`;
 </script>
 
+<!-- svelte-ignore a11y-role-has-required-aria-props-->
 <svelte:element this={tag} class="gradient-heading {$$props.class}" data-testid="gradient-heading" role="heading">
 	<span class={classesWrapper}>
 		<slot />


endigo9740:
@niktek Do you think we go one step further and still provide the an a11y prop for `level`? In the off chance someone sets a non-heading tag they would still have the ability to set this value manually. I'd say we add that to the component/docs -or- just give the recommendation that to keep a11y you should use headings. The former seems more flexible though.

endigo9740:
Look like I already thought of this, so I'll go ahead and merge. Thanks!

Screen Shot 2022-09-06 at 10 53 28 AM

PR #192, opened 2022-09-07
Open on hover for Tab and Accordion

LukeHagar:
this allows for tabs and accordions to be opened when hovering on the item that would normally be clicked, all that is required is adding the openOnHover prop to the group component

### DIFF
--- a/src/lib/Accordion/AccordionGroup.svelte
+++ b/src/lib/Accordion/AccordionGroup.svelte
@@ -1,9 +1,12 @@
 <script lang="ts">
-	import { onMount } from 'svelte';
+	import { onMount, setContext } from 'svelte';
 
 	// Props
 	export let collapse: boolean = true;
 	export let spacing: string = 'space-y-4';
+	export let openOnHover: boolean = false;
+
+	setContext('openOnHover', openOnHover);
 
 	let elemAccordian: HTMLElement;
 
--- a/src/lib/Accordion/AccordionItem.svelte
+++ b/src/lib/Accordion/AccordionItem.svelte
@@ -1,8 +1,16 @@
 <script lang="ts">
 	import SvgIcon from '$lib/SvgIcon/SvgIcon.svelte';
 
+	function openAccordion() {
+		if (openOnHover) open = true;
+	}
+	function closeAccordion() {
+		if (openOnHover) open = false;
+	}
+
 	// Props
 	export let open: boolean = false;
+	export let openOnHover: boolean = false;
 	export let hover: string = 'hover:bg-primary-500/10';
 	export let spacing: string = 'space-y-2';
 	export let padding: string = 'px-4 py-2';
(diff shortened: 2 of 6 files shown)


luke-hagar-sp:
Kapture 2022-09-07 at 08 34 40
Kapture 2022-09-07 at 08 38 45

endigo9740:
@luke-hagar-sp @LukeHagar 

Hey Luke, thanks for sending the PR to review. A few things here...

## Per the Contribution Guide

- Make sure you've read the contribution guide, linked above. There's a few guidelines you've not account for here.
- You've submitted a few issues without prior discussion. This is a valuable step for discussing and agreeing upon the best way to implement updates or features. It also ensures we don't have two or more people affecting the same code in different ways. I could have probably made a few suggestions beforehand had I known you were going to add this.
- We have a recommend branch naming convention which you're not following. This is helpful to testers like me when testing contributions from a number of folks. The bigger the numbers get the harder that'll be, so best get that ingrained now so others follow our lead
- The newly introduce prop does not follow the recommended naming convention (see my recommended change below)
- The newly introduced prop is not documented, so no one would know about it!

> Side note: there is a way to send "draft PRs" for this purpose. We had someone do it the other day but I haven't a clue how it's done. Let's work together to figure this out please! We may want to include the steps to do so in the contribution guide. I'd welcome a PR from you adding those steps :)

## Within the Contributed Code

- I'd recommend the new prop name be `hoverable: boolean` or something simple and descriptive. I'm not opposed to `hover: boolean` but it may conflict with hover CSS style props conventions used in other comonents
- I'd recommend you keep the Props/Context in roughly the same order top-to-bottom. This helps readability in the code.
- On the Tab component, the `selectTab()` function is kind of thrown in the middle of the prop declarations at the top. Please move it down near the other function (onKeyDown) and consider a generic name that matches the event it's paired to (ex: `onMouseEnter()`). There may be other functionality that needs to use this event in the future, similar to how `onKeyDown()` can listen and fork to handle multiple keyboard shortcuts.
- I know we won't have our meeting until tomorrow regarding testing, but make a mental note that we'll need to update the tests to utilize any newly introduced props and contexts. Per this, you're welcome to wait for your follow up commit when this is done. I won't accept the PR before that.

The final thing - if we had we had prior discuss of this update beforehand, I would have asked if you made considerations for how it will affect mobile users. Off hand I believe this will still operate for those users, but it may fire the select event twice (one per "click", and one per "focus"; I'm not sure!). We're not adverse to desktop-only features (Tooltips are as well!) but we need to make sure we state plainly this is a desktop/mouse only feature in the documentation for this prop. There's no such thing as "hover" on mobile!

endigo9740:
@luke-hagar-sp @LukeHagar FYI Luke it looks like these changes would not longer merge cleanly. This is due to the recent structural change in the /lib directory. If you're still keen to make the recommend updates let me know. If we need to postpone that's fine too, just note we might need to catch up to the change on dev. If you don't feel like you'll be able to revise and update these soon we might be best to close our the PR and come back to this later. Either you or I can close a PR here.

endigo9740:
With the tab redesign on the other tab this is way out of date. Closing for now. We can revisit in the future though!

PR #235, opened 2022-09-16
Updated Roadmap Link to point to the finished Roadmap instead of the proposal.

Sarenor:


endigo9740:
I can't believe GitHub doesn't persist the old links. It was the same page! But thanks for this. Very useful!

PR #245, opened 2022-09-20
Changed exported icons to a const Map type.

twobitunicorn:
I was working on something using SvgIcon and was having an issue with the explicit  `any` type.  Replaced it with a `Map<{path:string, viewBox?:string}>` type.


endigo9740:
Hey @twobitunicorn thanks for the submission. I might advise you give our contribution guidelines a look before additional submissions:

We ask that as a courtesy that folks give contributors a heads up when making non-trivial changes. That can be a GitHub issue, reach out on Discord, or even send a draft PR early to stake claim on a task.

A few things to note per this component in particular:

1. This component is currently unlisted in our docs, which means it's not intended for public consumption yet. We should make that more clear, so that's on us!
2. This one is very much a work in progress. We like to test internally, typically on our docs site. When it reaches a stability point we'll document and announce it in a proper release.
3. We're anticipating some changes to be made for this component, potentially a major refactor, which means major breaking issue that would affect anyone using it.

Mainly I just want you aware that either the component may update drastically or perhaps even be discontinued and removed. In either case if this is something you rely on in a production environment I highly advise rolling your own version. Perhaps cloned from ours with your required changes for now?

I'm going to deny this for now. However, I do encourage you to send more details on why you made the change. It's not immediately clear where the failure was prior. We can definitely take this into consideration going forward!

Hope that makes sense. Please reach out if I can clarify anything!

Thanks,
Chris

twobitunicorn:
@endigo9740 Fair enough!  I wanted to checkout how the icons looked on my site and copied over the icon code into my project.  The `any` was getting picked up by the linter and I changed it to the `Map` type to give it some solid typing that typescript can work with.  As it was late, and I was on a roll, I just carried over those changes and created the PR.  You are totally right though, unsolicited so my apologies.

I have been using the approach of `@steeze-ui/svelte-icon` in other projects and have liked that the icons are full fledged objects instead of strings that may or may not exist at runtime.

Thanks again, looking forward to the next release!

endigo9740:
Ok I'll take note of that! I'll also take a look at the icon library you mentioned. Our goal with Skeleton is to keep it icon agnostic, that way people can use whatever solution they need or want. So that useful information.

Thanks for sharing!

PR #246, opened 2022-09-20
Spelling (compoonent --> component) DocsSidebar.svelte

twobitunicorn:


endigo9740:
@twobitunicorn good catch, thanks. What a silly typo too hehe :)

FYI we're right on the tail end of a fairly large set of changes with a new release in a couple days. We've been moving quickly, so I'm positive we'll see similar issues in the public facing docs too. I'd gladly accept more PRs for this if you see them.

Thanks for your contribution though!

PR #253, opened 2022-09-21
Update packages

niktek:
This is going to become more important for GH actions to auto build the package. My current testing shows that it is really pedantic about matching package number versions in package.json and package lock files.


endigo9740:
Good call. I had planned this but it just got lose in the shuffle! Given that even minor updates can sometimes introduce breaking changes I'll leave this one for a deeper review in the morning. If all is well I'll merge proper before moving forward with the other testing mentioned on Discord.

endigo9740:
Ok, so I see this updates Svelte/Package to the v4 that we want to avoid. I'll hold off on merging this until you give the all clear. We're probably fine to launch with the current deps for the time being.

endigo9740:
We'll test this instead:
https://github.com/Brain-Bones/skeleton/pull/263

PR #256, opened 2022-09-21
Re-add deprecated components

niktek:
List and ListItem are baaaacccck!  At least for one more version to assist users with transition.  This was taken from commit `3c70daff99` which was the format pass.

The package.json is just pinning the @sveltejs/package version till a fix is found.


endigo9740:
Per:
https://github.com/Brain-Bones/skeleton/discussions/250

Screen Shot 2022-09-21 at 12 40 43 PM

PR #262, opened 2022-09-21
ListBox was Listbox at few places

ak4zh:
Just fixed a few typos.
And normalized the use of text `Listbox` to `ListBox` in docs to avoid any confusion.


endigo9740:
That definitely needed to be fixed in the import example, but the versions in the `<h3>` tags will need to be reverted. These follow two conventions:

- When spelled out, for humans to read, it's `Listbox`. Capital case ([example](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/))
- For components, we utilize camel case for consistency, so `ListBox`

If you don't mind doing that, otherwise I don't mind grabbing that on my either. Either way!

endigo9740:
Went ahead and handled the title reverts. Merging now! Thanks for the contribution!

PR #263, opened 2022-09-21
Bumped to package.next.5

niktek:
I think there are still issues, but would love a sanity check of someone else making a package.  Basically, the generated Skeleton package should fail to install in a new project due to tarball integrity.


endigo9740:
Sounds good. When I'm back I'll pull this down on it's own branch. Update my packages. Then build a tarball to test in a fresh new SvelteKit project. Assuming that goes well, I'll merge and build the proper release with this. If it doesn't we'll keep what we have and postpone these updates to the next release.

Honestly I'm fine to postpone. We know what we have is working, as we've been testing it for a week or two at least. So let me know if you want to hold off for now!

PR #265, opened 2022-09-22
A11y-homepage-changes

ryceg:
Mostly just some straightforward housekeeping work of ensuring that heading elements increment sequentially. 

Also of note: 

* Made more use of the <nav> element for components that are always going to be navigation related (such as Breadcrumb)
* Changed a couple to use native Svelte event listeners
* Added a couple sr-only elements in the docs to help with the navigation


endigo9740:
Just FYI I'll aim to do a proper review of this in the morning (US/CST time). I'll probably any notable feedback here when I do, but at a glance things looked good. We may just need to review the heading hierarchy stuff we reviewed in our chat! Ping me if you see me online on Discord tomorrow and we can do that!

endigo9740:
I complete forgot this today! I'm going to make a note to get this tomorrow though. Sorry for the delay! It's been a busy week heh

endigo9740:
@ryceg Ok, finally had a chance to complete a proper review of this. The following is going to be a mix of questions, compliments, and just general notes. I want to ensure I understand a few of these changes so I can ensure we institute these as the recommended practice going forward.

- General:
    1. I noticed you're using a `.sr-only` class. Is that from Tailwind? I didn't see it defined anywhere.
    2. Thanks for adding additional Typescript types, especially for keyboard events!
    3. I wasn't aware we could define roles on `<slot>` elements (per Menu). So thanks for the heads up!
    4. I see a few files with code indention issues, make sure you'r running Prettier please! It can be set to run automatically on save if that's easier
- Breadcrumbs:
    1. I noticed you change to a `nav > ol > li` format. Is this because breadcrumbs have an inherit order left-to-right (in LTR languages at least)
    2. I noticed this was the only instance where you set a default for the label prop value. Should we have concerns with this not being unique? I figured no label was better than users accidentally getting two of the same label. Though multiple breadcrumbs on one page is not really recommended obviously!
- Menu:
    1. There appears to be a breaking issue on the documentation page. Only a portion of the page loads and there's a console error (screenshot attached below).
    2.  I like the use of the `<svelte:window>` for keyboard events. Great change there.
- Range Slider:
    1. I noticed you dropped a slew of ARIA roles for this one. Does the native range input handle this for us?
    2. Also the `on:keydown` directive was removed, but yet the function still remains in the script tag. I see the input events I set already seem to work native and were redundant. Seems we're safe to drop the function here?
- Tab Group:
    - Thanks for switching to the `on:keydown` from an event listener. I kept meaning to go back and fix that one!
- Homepage
    - Thanks for fixing the `<card> -> <div>` issue. We had a lot of references of these, so figures I would mess one up!

Finally, my big question comes in regards to heading hierarchy. I'm aware it was an issue, due to the Axe extension, but I have mixed feeling about this change. I understand the structural importance this would have for impaired users, however, it does make it difficult to design around. Especially since, visually, the heading also tend to refer to a particular size heading size. For example, look at the homepage compared to the production site. We now have text wrapping where it didn't before which breaks the flow of the design. Any tips for balancing the two? (visual versus a11y).

Overall this looks great though. If we can get this caught up to to the `dev` branch, and fix the couple issues mentioned above, I'd be glad to merge asap. Thanks!

Screenshot of the Menu issue:

Screen Shot 2022-09-23 at 11 32 47 AM

ryceg:
Hi @endigo9740 thanks for taking the time to look at this! 

General

1. `.sr-only` is indeed a Tailwind class; see https://tailwindcss.com/docs/screen-readers#screen-reader-only-elements for info
2. I'll add more in as I see them
3. Svelte is truly dope.
4. Will do- I believe that you can set up a Husky hook to handle this sort of thing, although I'm not very familiar with that.

Breadcrumbs:
1. `<nav>` is needed because it is a navigational element, and `ol` is used over `ul` because the breadcrumb is inherently a linear pathway. Ordered lists do not lose any semantic meaning in rtl languages (although I admit I didn't test this).
2. That was a late night addition, and in hindsight, I can see instances where you might have two breadcrumbs on the same page. Depends on how much hand-holding you want to do. It's important to note that `aria-labelledby` is preferred for a11y, and we don't currently support that.


Menu:
1. Sorry! I'll have a look, not sure how I missed that.
2. We could also use an inline `on:resize`.

Range slider:
1. Yes, native elements handle all that stuff. It's important to avoid nesting interactive elements.
2. Yes, we can remove that, whoops.

With regard to the headings, I think that you would be better served by passing in custom styling to force the elements to display as differently styled elements. One of the things that I really love about Tailwind is the way that it eliminates the default stylings so that you have the autonomy to make conscious decisions about how to style elements. 

Overall, I think that what we need to remember is that `aria` attributes are fallbacks, and it is _always_ better to use native HTML elements over adding aria attributes to generic elements to force them to behave in the same way. No ARIA is better than bad ARIA!

endigo9740:
Hey @ryceg Some great progress so far. Thanks for replying to all my questions, some good info here that I'll try and find a way to share with a couple other regular contributors.

A few more things of note:

- I wish I had mentioned this sooner, but I recommend checking the open GitHub issues. Several components you're updating are on the roster for a bit of a refactor, if not a couple rebuild. Not saying your effort is wasted, just may be better spent in other places until those are ready.
- Likewise I see a few updates to the components are now handled by "Tailwind Elements". The component versions are scheduled to be dropped by next release, so best to leave those be. Totally understand some of that may be bulk find/replace stuff.
- I see several instances of setting undefined values to empty strings. You might check how those render. I was under the impression undefined would not appear in most cases, where as an empty string will. But I could be wrong! If the empty string is working then we may just want to opt for `foo: string` and skip the possible undefined type. (aka "K.I.S.S")
- What you said about headings makes total sense. I'll setup heading classes instead of styling the native element for these. Folks can always "map" those to their own headings if they prefer that format.
- Agreed on the range slider, there is some useful logic in there for determining the datalist ticks. But I almost wonder if this is something better served as an action paired with a native datalist and input:range. Something to think about. But I'm allllll about native where we can make it work.

Ok, I'll leave it there. Don't want to backseat too hard, but also want to ensure you're not putting in effort where you don't need to. Especially stuff that's slated to be removed! Just ping me when you're ready for another review.

endigo9740:
New issues created based on the discussion above:

- https://github.com/Brain-Bones/skeleton/issues/279
- https://github.com/Brain-Bones/skeleton/issues/280
- https://github.com/Brain-Bones/skeleton/issues/261#issuecomment-1256857047 (note added)

ryceg:
Ah, I see that stuff now- I'll have to put some time into reading the reasoning behind the change, though it does look good!

> I see several instances of setting undefined values to empty strings. You might check how those render. I was under the impression undefined would not appear in most cases, where as an empty string will. But I could be wrong! If the empty string is working then we may just want to opt for foo: string and skip the possible undefined type. Keep it simple!

So, this is a fun quirk of JavaScript. String concatenation is screwy, but coercing `undefined` to a string (e.g. console.logging) returns `'undefined'`. For a quick example, try this:

```ts 
const fizz = undefined
const foo = 'foo'
const bar = ''
const biff = undefined
const boff = 'py-10'
console.log(fizz + bar);
console.log(fizz + bar + boff);
console.log(fizz + foo + bar + 'baz1');
console.log(`${fizz}${foo}${bar}baz2`);
console.log(bar + foo + bar + 'baz3');
```

`export let bar = ''` has the _type_ of `string` but is a falsey value, which means that we can perform all of the same truthy tests that we would normally. However, it's a design choice above my paygrade to say whether it's appropriate to remove the `foo: string | undefined` typing. The difference is subtle, but it's essentially marking it as optional, with a falsey initialisation when a value isn't passed through, as opposed to `export let bar: string = ''` which has an implicit type safety that relies on user knowledge that `string` is not always truthy (just like how `0` is a falsey value which can really mess up people using the double pipe operator). It's always good practice to test for falsey values, but the difference is that keeping the `undefined` in makes it more explicit.

endigo9740:
@ryceg what do you think about putting a pin in what we have so far and getting this reviewed and merged?

My concern is trying to do too much on a single feature branch. Especially since this touches a lot of files. We might do better to break remaining changes up and do dedicated passes in a smaller piecemeal fashion.

I'd really like to get the deprecated components dropped from the library, but I need your changes in first since you modified some of those files.  I had to give someone else a heads up not to touch those today, so they're already causing some trouble!

Main thing is I'm trying to avoid the merge becoming a major pain in the rear! Bite size updates will always be best in a collaborative setting like this.

ryceg:
Of course, apologies- didn't mean to tread on toes, happy to have you ignore the bits that touch components that are getting deprecated.

endigo9740:
Oh no worries at all! Unfortunately I believe it's all or nothing. I wish Git/GitHub let us selectively pick what to merge. Those files have been modified so they'll be part of the PR and merged. General workflow will be this:

- Let's wrap on any remaining fixes/adjustments you want to make here
- Tap "ready for review" on here to signal I can pull and review
- I'm happy to make any final modifications needed to prep for merge
- We'll merge and then I'll immediately drop the deprecated components and get dev up to date
- We're then good to begin work on whatever you want to tackle next, a11y or otherwise!

Thank you sir!

endigo9740:
I've gone ahead and finalized what I could of these changes. At least enough to get it to a point we can merge and unblock a few things.

A few of things of note here:

- Some of these changes were meandered a bit. Best rule of thumb is only change what you have to change in a single feature branch. Keep it laser focused on one type of tasks - ex: modify headings, modify empty states, fixing a particular component, etc. Trying to do too much at once makes it very difficult to review and will make merges a bit messier since lots of people are touching the codebase
- I've reverted the heading changes for now, let's handle this in a dedicate pass per this ticket. The idea is sound, but again, focus: https://github.com/Brain-Bones/skeleton/issues/280
- I've removed the AccordionGroup `allowMultipleOpen`. This is redundant with `collapse: false`. Perhaps we need to document that better?
- Great job with the range slider changes though. The focus on the native element features are perfect and I wasn't aware of the `crypto` feature in JS. That will get a lot of usage out of me!
- I've reverted the `string | undefined = ''` change. I'll attach two screenshots below and explain in detail.

This is how an element renders when and empty string is provided at the value for a prop. Notice the random `id` without a values and extra spaces added to the class list?

string

Here's the same element using `undefined`. The unset ID is not present at all, and those extra spaces are not present:

undefined

This is something I stumbled upon a while back. I was under the same impression as you, that empty string would suffice. But that's not how Svelte handles these unfortunately. Perhaps we revisit and take another approach with this? I understand wanting to keep the type fixed.

As mentioned, we're not done with the changes you had in mind or were implementing, we'll just want to handle each in a smaller and more localized fashion. Let's sync before we tackle the next wave please!

Either way there's lots of great improvements here, so great job!

endigo9740:
One more thing, it appears the Data Table and Range Slider test cases are now failing. I believe these are due to changes you've implemented. The ranger slider test doesn't seem to be able to find `crypto`. Let me know if you have any thoughts for this. A bugfix PR is welcome here.

Screen Shot 2022-09-26 at 3 23 54 PM

src/lib/components/Menu/Menu.svelte
```diff
@@ -45,33 +47,25 @@
 		setOrigin();
 	}
 
-
-	/**
-	 * Searches for the first parent node that can scroll
-	 * @see {@link https://thewebdev.info/2021/06/27/how-to-find-the-first-scrollable-parent-element-with-javascript/|This article by The Web Dev}
-	 */
+	// Searches for the first parent node that can scroll
+	// Source: https://thewebdev.info/2021/06/27/how-to-find-the-first-scrollable-parent-element-with-javascript/
```
ryceg, line 51: I'd suggest preferring JSDoc syntax, as it provides the info on hover, which greatly improves the DX. Though it's not uniform with the rest of the codebase, it would save a big touch if we decide to autogenerate documentation further down the line. Double slashes obfuscates code documentation.

endigo9740: Hmm, I've heard of JSDoc but haven't used it personally. Can you provide an example of what you mean by "oh hover?". Does that require an editor plugin, I don't recall seeing that.

At least in VS Code you can control + click to open links in the current format.

I'd be interested in reviewing this. If we do it, I'd rather go all in rather than mixing here and there.

PR #281, opened 2022-09-24
DataTable-typing

ryceg:
This refactors some of the code behind the data tables, to allow for better typings underneath a unified function that handles both strings and numbers. I also added in some tests, but I haven't been able to get the test runner to work on my machine, so I'm unable to verify that they actually work.


endigo9740:
See my comments on Discord. Some of these changes may be relevant to the upcoming refactor I have in mind. So I'll go ahead and accept. These have been merged!

Issue #291, opened 2022-09-26
LightSwitch component updates


endigo9740:
Per this ticket, we've received a request to expose the current light/dark mode state into a globally visible scope:

- https://github.com/Brain-Bones/skeleton/issues/218

Additionally, we should probably go ahead and update the component utilize the new LocalStorageStore utility added here:

- https://github.com/Brain-Bones/skeleton/issues/283

Likewise, we should also give this one more pass to see if we can improve load time performance. Currently there's a "flash" as the page loads and the correct mode is applied.

endigo9740:
Resolved by @luke-hagar-sp . We'll revisit performance again in the future, but we may be hitting a bottleneck with pure Svelte at this point.

PR #293, opened 2022-09-27
Exposed LightSwitch store, Used LocalStorageWritable

luke-hagar-sp:
Created a separate stores file to be exposed and altered the lightswitch functionality to use the LocalStorageStore to its fullest

Partially Addresses #291

I feel we have this as fast as reasonable at the moment.


endigo9740:
Again, nailed it. I just made some minor adjustments.

I agree with the performance, our bottleneck is the `onMount` method, but to my knowledge we don't have a faster means to access DOM related data with pure Svelte. SvelteKit offers `browser` but it's not available for Vite/Astro so we cannot proceed with it.

This update is going to make James very happy! Great job Luke!

PR #299, opened 2022-09-28
Table-Typings

ryceg:
Creates a new folder called `types` and adds a property that enforces consistent titles for headers.

### DIFF
--- a/src/lib/types/tableProps.ts
+++ b/src/lib/types/tableProps.ts
@@ -9,6 +9,7 @@ type TablePropKeys =
 | 'Description'
 | 'Slot'
 | 'Event'
+| 'Class'
 
 export type TableProps<T = void> = T extends Record<string,unknown> ? TableKeyedSource<T> : TableArray
 type TableArray = {
--- a/src/lib/types/tableProps.ts
+++ b/src/lib/types/tableProps.ts
@@ -1,0 +1,22 @@
+type TablePropKeys = 
+| 'Name'
+| 'Prop' 
+| 'Style Prop'
+| 'Type' 
+| 'Default' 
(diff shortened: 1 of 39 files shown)


endigo9740:
Unfortunately @ryceg I don't think I can move forward with this one. As I mentioned on Discord, it's well intended, but poses a number of issues.

- Doesn't follow convention for where the Type file should be stored (let's keep these localized, in the component directory)
- It tries to specify all the various Table prop keys we use in the docs, but this would only be useful for us. Not for end users, which kind of diminishes its usefulness. You would be surprised how many end users try to use stuff that's just for us, such as the SVGIcon component, even though it's completely unlisted. Then of course run into trouble!
- Just kind of drops the import into each route page wherever. Whereas I'm making a consorted effort to group and organize imports to keep this clear for everyone what and where things are coming from, and what they represent. Introducing an opinionated structure in other words. This is something we absolute need to add to the "style guide" section mentioned for the contributor doc updates.
- It might have some troublesome overlap with this on-going PR, which started first:  https://github.com/Brain-Bones/skeleton/pull/292
- It'll definitely have issue with my upcoming Menu changes, since that's being moved from a Component -> Utility

That said, I think what you have here is solid. We're not going to implement this YET, but we can alongside these changes:

- https://github.com/Brain-Bones/skeleton/issues/217
- https://github.com/Brain-Bones/skeleton/issues/302

But the important part of not what we're changing, but WHEN we're changing this. So it has minimal impact on others.

FYI the PR will remain here on GitHub so I'll link it in the Table refactor ticket as a reminder to reference this and reimplement a version of the types you did. Again the idea is solid! I want to add it. Just not quite yet!

PR #347, opened 2022-10-10
Removed unnecessary 'header' from AppShell Property descriptions

Sarenor:
## What does your PR address?

Fixed a small issue in the AppShell documentation with unnecessary mentions of header


endigo9740:
Ahh great find. Thanks!

PR #364, opened 2022-10-13
Sveld auto doc of Props and Accessibility

### LINKED ISSUE #351: Update to utilize JSDocs

@niktek confirmed we can use JSDocs comments to add additional intellense information for component props. We should aim to implement these asap

image

Enhancement

If possible we should look into a way to share the same data between the JSDocs comment AND the docs page UI description, so we're not repeating ourselves. Assuming this is possible.

### Reference PRs

- https://github.com/Brain-Bones/skeleton/pull/412

## Sveld Documentation Progress

> NOTE: use the Accordion and Paginator components as reference for each data type

Chris:

- [x] Accordions
- [x] Code Blocks
- [x] Dialogs
- [x] Drawers
- [x] Lightswitch
- [x] Toasts

Nik:

- [x] Alerts
- [x] App Bar
- [x] App Shell
- [x] Avatars
- [x] Breadcrumbs
- [x] Conic Gradients
- [x] Dividers
- [x] File Buttons
- [x] File Dropzone
- [x] Gradient Headings
- [x] SVGIcon

Rhys:

- [x] Data Tables
- [x] Listboxes
- [x] Paginators
- [x] Progress Bars
- [x] Progress Radials
- [x] Radio Groups
- [x] Range Sliders
- [x] Slide Toggles
- [x] Steppers
- [x] Tabs

niktek:
DO NOT MERGE - this is for previewing
## What does your PR address?

Fixes #351 

Notes: This actually uses https://github.com/mattjennings/vite-plugin-sveld which in turn uses sveld.

To differentiate between props that are Props and props that are for Accessibility, it is using magic strings in the description, if `a11y` appears in the description, it will be moved into the Accessibility tab.

To enable this for a components doc page, add `components={["Accordion/AccordionGroup","Accordion/AccordionItem"]}` to the DocsShell's props. It handles a plain string for a single component and an array of strings components.

Although it handles migrating the Props tabs over incrementally (it replaces the `props DocsShellTable`, it can't be done for Accessibility due to extraneous tables being added like KeyBindings that sveld can't detect.

I've done the Accordion components as an example to show what's needed.  They are not for publishing yet, but rather to show what's needed and how different scenarios work.

Where a type has a union of possible values, it must be declared using the `@type` directive, otherwise only the last type of the union is declared from the auto-inference.

Vite also complains mightily about doing dynamic imports like this, so some more testing needs to be done of a full published docs site.  There is a fallback option if issues arise, but it's uglier to wire up.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
 		"tslib": "^2.4.0",
 		"typescript": "^4.8.3",
 		"vite": "^3.1.3",
+		"vite-plugin-sveld": "^1.0.3",
 		"vitest": "^0.23.4"
 	},
 	"type": "module",
--- a/src/docs/DocsShell/DocsShell.svelte
+++ b/src/docs/DocsShell/DocsShell.svelte
@@ -14,6 +14,7 @@
 	import { toastStore, type ToastMessage } from '$lib/utilities/Toast/stores';
 
 	// Props
+	export let components: string | string[] = '';
 	export let settings: DocsShellSettings;
 	export let properties: DocsShellTable[] | undefined = undefined;
 	export let events: DocsShellTable[] | undefined = undefined;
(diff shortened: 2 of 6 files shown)


ryceg:
So, Sveld is mostly fit for purpose, but there's a couple deficits that get in the way of us using it fully for documentation, which I think is a very good long-term goal to aim for. 

The Sveld interpreter is notably missing many of the JSDoc tags, which is working against us, but isn't actually a dealbreaker- it just interprets everything as an extension of the description, and includes new lines. If we use JSDoc tags, then we can use  `@` as the opening tag, and `/n` as a closing tag. For a robust documentation system, I think that we would want to make use of some custom tags, easy enough to set up. For writing the logic for dealing with tags, it might be good to keep in mind the different tag kinds (https://tsdoc.org/pages/spec/tag_kinds/)

Sveld has a rudimentary version of this to handle slots; `@slots` is kinda janky as it's closer to an inline tag, rather than a block tag, and only lets you define a couple different parameters. Notably, we do not expose any information about what props are passed through to slots at the moment in our current documentation, which is definitely sub-optimal. We also break from Svelte architecture a little bit by defining whether a slot is "required" to be filled; typically Svelte slots are entirely optional, whereas we have elements that (rightly) have a need to be filled, like the details component. Adding in some extra logic into the Sveld interpreter to allow us to tag it with a JSDoc custom  modifier tag `@required` would be needed to make the slots stuff work for our needs, but we could always fake it with a passed prop (but I'm really loathe to do that).

endigo9740:
@ryceg see my response on Discord.

endigo9740:
This will still be available for reference in the closed PRs, but closing for now.

PR #407, opened 2022-10-20
Base functionality needed for JSDocs

niktek:
This functionality is only in the docs site and only gets activated as doc pages start adding the `components` prop on DocShell.  Just greatly simplifies progressively adding JSDocs and doc pages without having to do 3 way branch merges etc.

### DIFF
--- a/src/docs/DocsShell/DocsShell.svelte
+++ b/src/docs/DocsShell/DocsShell.svelte
@@ -15,6 +15,7 @@
 
 	// Props
 	export let settings: DocsShellSettings;
+	export let components: string | string[];
 	export let properties: DocsShellTable[] | undefined = undefined;
 	export let events: DocsShellTable[] | undefined = undefined;
 	export let classes: DocsShellTable[] | undefined = undefined;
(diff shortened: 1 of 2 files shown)


endigo9740:
Here's my notes and feedback from a preliminary review:

- The fact that we're parsing info from strings does make me uneasy. There's so many ways this can go wrong.
- Per this bit, we have to import the component dynamically? There's no way we can pass this info through from the pages that already have the components imported?
`const compData = await import(/* @vite-ignore */ '../../lib/components/' + component + '.svelte?raw&sveld');`
- Speaking of things that can go wrong with string parsing, relying on this is probably going to bite us. All is takes is one `http:` URL and we've got problems:
`const urlIndex = prop.description.indexOf('https');`
- I mentioned the plan to move this outside of the doc shell component, but just reiterating that here publicly

I'm going to need to go hands on with this tomorrow before I feel comfortable merging.

endigo9740:
@niktek I believe this older PR is redundant with the new one sent last night, so closing this to tidy up.

PR #412, opened 2022-10-21
Foundation for Sveld Auto-Docs

niktek:
Need some help on this one:

In src/docs/DocsShell/sveldParser.ts the components param regresses to be just an `object` as per the console.log, even when I'm passing in what should be an array of DocsShellComponentRef's.


niktek:
Foundation elements in place.

niktek:
re-open

niktek:
Close you heathen

PR #418, opened 2022-10-22
Fix svelte guide component import error

freedumb2001:
Move GradientHeading import statement from +layout to +page


endigo9740:
@freedumb2001 FYI I had to make a couple quick adjustments, including copying the same info over to the Vite guide which mirrored these instructions. But should be good now. Expect this to come as part of the next release. And thanks for your contribution!

PR #429, opened 2022-10-26
First pass on nik's list

niktek:
A few AUDIT items in there.


niktek:
All descriptions are brought over, fixed `<a type> | undefined = undefined` throughout (except for where it matters like indeterminate spinner).  Have `//AUDIT` sprinkled through to draw your attention to types that need tightening and other queries.

Pretty sure I got most slots (problem remaining that some default slots really do need to be described in some way).

Deleted commented out sections of doc pages.

PR #452, opened 2022-10-27
Feat/enhance-with-regex

ryceg:
Allows us to have our cake and eat it by taking markdown links from tsdoc and formatting it into the correct <a> in the docs. Code is still not done, nor are tests, which are CRITICAL for this.


endigo9740:
@ryceg Note that with the upcoming Table refactors I had to reworked a bit of the Sveld Mapper utilities. The data structure for the new tables is different. You can see the changes on my draft PR for tables.

endigo9740:
@ryceg Given this one is becoming a bit long in the tooth, and we've not really had any major issues to the existing setup, what do you think about closing this issue for now? Perhaps returning to this in the future if we see a major Svelte update occur. Make it a more consorted effort to push this forward in a major way?

ryceg:
Agreed- YAGNI is probably going to hold true here, and we can always revisit if we _do_ end up needing it.

PR #479, opened 2022-10-31
Chore/migrate to using @bnb/skeleton alias

### LINKED ISSUE #478: DataTableService exports
### Current Behavior

There is currently two exports from DataTableService that are used internally of DataTable.svelte (sortAsc and sortDesc), then there is a mapTableSource that is used in the doc page for data-tables.

On the docs page there is also a CodeBlock that shows an import of `import { mapTableSource } from '@brainandbones/skeleton/components/Table/DataTableService';` which I'm pretty sure won't work as the alias `@brainandbones/skeleton` points to the `src/index.ts` rather than a directory alias.

What probably needs to happen is that these DataTableService exports need to migrate to the top level src/index.ts file and probably have a bit of a namespace scoping like dtsSortAsc or something like that ?

### Steps To Reproduce

_No response_

### Anything else?

_No response_

niktek:
DO NOT MERGE:
Can't put it into draft mode for some reason.
In preparation for migrating to a monorepo.

- all doc site pages and components now use @brainandbones/skeleton as the alias as $lib/hardcoded/paths will not exist there post migration.
- added exports in index.ts for AppRail and AppRailTile - done in separate PR
- added exports in index.ts for DataTableService.ts and updated data-tables doc page - fixes #478 (though they are not scoped name wise - not sure if we need to).  May no longer be needed.


niktek:
I seem to have lost the ability to put this PR back into draft mode :(

endigo9740:
I've set it as a draft, but will be curious to hear more about what this entails.

PR #480, opened 2022-10-31
gitignore fixes for monorepo

niktek:
The leading slashes made it only work at the root level, rather than flow down into multiple projects


endigo9740:
I've confirmed there's no notable changes with this update applied so I'll go ahead and merge.

PR #484, opened 2022-11-01
Docs/fix-fileDropzone-typo

### LINKED ISSUE #468: FileDropzone default text information is plural
### Current Behavior

`FileDropzone` is in single file mode by default.
But the information text is
> Drop files or click to select

This would be better
> Drop a file or click to select

And after enabling multiple file mode,
> Drop files or click to select

### Steps To Reproduce

```
<FileDropzone />
```

### Anything else?

_No response_

ryceg:
## Before submitting the PR:
- [x] Does your PR reference an issue? If not, please [chat to the team on Discord](https://discord.gg/EXqV7W8MtY) or [GitHub](https://github.com/Brain-Bones/skeleton/discussions) before submission.
- [x] Did you you update and run tests before submission using `npm run test`?
- [x] Does your branch follow our [naming convention](https://skeleton.brainandbonesllc.com/docs/contributions)? If not, please amend the branch name using `branch -m new-branch-name`

## What does your PR address?

Implements @leranjun's suggested fix, and closes #468

### DIFF
--- a/src/docs/DocsShell/sveldMapper.ts
+++ b/src/docs/DocsShell/sveldMapper.ts
@@ -1,33 +1,89 @@
+
 import type { Component } from './types';
 
 interface MapperOutput {
-	headings: string[];
-	source: string[][];
+	headings: string[]
+	source: string[][]
 }
 
 // Mapper: Props
 export function sveldMapperProps(component: Component): MapperOutput {
 	const { props } = component.sveld;
-	const propsHeadings = ['Name', 'Type', 'Value', 'Description'];
+	const propsHeadings = ['Name', 'Type', 'Value', 'Description']
 	// Filter props with undefined types (exclude getContext)
 	const propsFiltered = props.filter((p) => p.type !== undefined);
 	// Return table headings/source
 	function cleanValue(value: string | undefined): string {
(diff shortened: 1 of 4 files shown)


endigo9740:
@ryceg This seems to bundle in a bunch of changes to the Sveld Mapper. Not sure if you intended to include those in this.

ryceg:
Aww balls I thought I removed those 😖

leranjun:
I think you might need to drop the unrelated commits so that it matches `dev` :p

Btw, copying my original comment over here:

> `$$restProps` seems to be [generally not recommended](https://svelte.dev/docs#template-syntax-attributes-and-props) due to optimisation issues. Could we consider just exporting `multiple` plainly instead?

endigo9740:
@leranjun I'm aware of Svelte's stance on restProps. We basically only use it for input-related inputs due to the large number possible attributes required. It's not just about multiple, but also id, name, required, etc. It's like a cat and mouse game. I don't like using it, but I also don't want to research every possible input proper out there per each use case. This is one reason we generally try to avoid input components.

endigo9740:
@ryceg Ping me when this is fixed and I'll merge right away!

PR #499, opened 2022-11-02
Dialog Refactor

### LINKED ISSUE #197: Refactor and improve the Dialog utility
Updating these instructions to detail the current list of requirements and desired updates:

- [x] Better transitions when moving between queued dialogs
- [x] Should allow for custom/interchangeable svelte components as the modal contents
- [x] Consider implementing a universal "focus trap" action for code reuse, include as part of the library
- [x] Should remove hardcoded English words, for basic i18n support (ex: "Confirm" for button)
- [x] Should be updated to the standards of all current components, per the consistency and Sveld updates

*EDIT*

We've opted to pass on using the native dialog element. It has too many quirks to work around right now:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog

### LINKED ISSUE #501: New Action: focus Trap

This is already completed, just filing a ticket so it can be noted for the next release.

image

New Feature

Coming as part of the Dialog overhaul:
https://github.com/Brain-Bones/skeleton/issues/197

endigo9740:
Fixes #197 
Fixes #501


src/lib/utilities/Dialog/examples/DialogExampleForm.svelte
```diff
@@ -0,0 +1,51 @@
+<script lang="ts">
+	// Props
+	/** Exposes parent props to this component. */
+	export let parent: any;
```
ryceg, line 4: `unknown` might be a better type to use for this.

src/lib/utilities/Dialog/stores.ts
```diff
@@ -2,30 +2,38 @@
...
+	slot?: string;
 }
-export interface DialogPrompt extends DialogAlert {
-	type: string;
-	value: any;
-	result: any;
+
+export interface Dialog {
+	/** Designate what type of component will display. */
+	type: 'alert' | 'confirm' | 'prompt' | 'component';
```
ryceg, line 16: Extract this out to a named type

src/lib/utilities/Dialog/stores.ts
```diff
@@ -2,30 +2,38 @@
...
+	/** Provide the dialog header content. Accepts HTML. */
+	title?: string;
+	/** Provide the dialog body content. Accepts HTML. */
+	body?: string;
+	/** Provide a URl to display an image within the dialog. */
+	image?: any;
+	/** Prompt Only - provide a prompt value. */
+	value?: any;
```
ryceg, line 24: unknown

src/lib/utilities/Dialog/stores.ts
```diff
@@ -2,30 +2,38 @@
...
+	/** Provide {component, props} to create a generate a custom component. */
+	component?: DialogComponent;
+	/** Provide a function. Returns the repsonse value. */
+	response?: any;
```
ryceg, line 28: unknown

src/lib/utilities/Dialog/stores.ts
```diff
@@ -2,30 +2,38 @@
...
+	type: 'alert' | 'confirm' | 'prompt' | 'component';
+	/** Provide the dialog header content. Accepts HTML. */
+	title?: string;
+	/** Provide the dialog body content. Accepts HTML. */
+	body?: string;
+	/** Provide a URl to display an image within the dialog. */
+	image?: any;
+	/** Prompt Only - provide a prompt value. */
+	value?: any;
+	/** Provide {component, props} to create a generate a custom component. */
```
ryceg, line 25: Reword for clarity

src/lib/actions/FocusTrap/focusTrap.ts
```diff
@@ -0,0 +1,45 @@
...
+	};
+	onInit();
+
+	function onDestory(): void {
+		if (elemLast) elemLast.removeEventListener('keydown', lastLastKeydown);
+	}
+
+	// Lifecycle
+	return {
+		update(newArgs: any) {
```
ryceg, line 37: `enabled = newArgs` makes me think that this should be a boolean?

src/lib/actions/FocusTrap/focusTrap.ts
```diff
@@ -0,0 +1,45 @@
...
+		if (e.code === 'Tab') {
+			e.preventDefault();
+			elemFirst.focus();
+		}
+	}
+
+	const onInit = () => {
+		if (enabled === false) return;
+		// Gather all focusable elements
+		const focusableElems: any = node.querySelectorAll(elemWhitelist);
```
ryceg, line 18: Since you're using a querySelectorAll, this would be a `NodeList`

src/lib/utilities/Dialog/Dialog.svelte
```diff
@@ -1,156 +1,177 @@
...
+	// Dialog Store Subscription
+	dialogStore.subscribe((dArr: any[]) => {
+		if (!dArr.length) return;
+		// Set the local dialog value (for prompt)
+		promptValue = dArr[0].value;
+	});
+
+	// Event Handlers ---
+
+	function onBackdropClick(e: any): void {
```
ryceg, line 68: `MouseEvent`

src/lib/utilities/Dialog/examples/DialogExampleForm.svelte
```diff
@@ -0,0 +1,51 @@
...
+	import { dialogStore } from '$lib/utilities/Dialog/stores';
+
+	// Form Data
+	const formData: any = {
+		name: 'Jane Doe',
+		tel: '214-555-1234',
+		email: 'jdoe@email.com'
+	};
+
+	// We've created a custom submit function pass the response and close the dialog.
```
Sarenor, line 16: Irrelevant typo:
```
// We've created a custom submit function to pass the response and close the dialog.
```

src/lib/utilities/Dialog/stores.ts
endigo9740: @ryceg This should always be a function. is there a better type for that? I realize most functions are Objects, like everything in JS, but that doesn't seem right.

src/lib/utilities/Dialog/stores.ts
```diff
@@ -2,30 +2,38 @@
...
 }
 
 function dialogService(): any {
 	const { subscribe, set, update } = writable([]);
 	return {
 		subscribe,
 		// Trigger - append to end of queue
-		trigger: (dialog: DialogAlert | DialogAlert | DialogAlert) =>
+		trigger: (dialog: Dialog) =>
 			update((dStore: any) => {
```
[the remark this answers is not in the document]
endigo9740: The preview doesn't highlight which line you're referring to, but I'm assume it's the last. The `dStore: any`, correct?

None of the following worked without error:

- `any[]`
- `DialogSettings[]`
- removing the type

Thoughts?

src/lib/utilities/Dialog/stores.ts
```diff
@@ -2,30 +2,38 @@
...
+
+export interface Dialog {
+	/** Designate what type of component will display. */
+	type: 'alert' | 'confirm' | 'prompt' | 'component';
+	/** Provide the dialog header content. Accepts HTML. */
+	title?: string;
+	/** Provide the dialog body content. Accepts HTML. */
+	body?: string;
+	/** Provide a URl to display an image within the dialog. */
+	image?: any;
```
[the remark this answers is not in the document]
endigo9740: I had no clue you could specify a specific string type. That's useful!

src/lib/actions/FocusTrap/focusTrap.ts
endigo9740: Immediately causes line 21/22 to error. Doesn't like those are set as `HTMLElement`, it suggests `Node`. But that change causes a bunch of stuff to error. So a bit of a rabbit hole. I might need you to review directly.

src/lib/utilities/Dialog/Dialog.svelte
endigo9740: Refuses to allow for `e.target.classList` when I do this, even though the value is there when logged. @ryceg I might suggest testing these changes directly before suggesting theme.

The goal here is we need to ensure the click is on the dialog backdrop, and not the modal window. If so, then allow the close to trigger. I tried rewriting to use `stopPropigation()` but no dice. So I'm open to suggestions.

src/lib/actions/FocusTrap/focusTrap.ts
```diff
@@ -0,0 +1,45 @@
+// Action: Focus Trap
+
+export function focusTrap(node: HTMLElement, enabled: boolean) {
+	const elemWhitelist: string = 'a[href], button, input, textarea, select, details, [tabindex]:not([tabindex="-1"])';
+	let elemFirst: HTMLElement;
+	let elemLast: HTMLElement;
+
+	function lastLastKeydown(e: KeyboardEvent): void {
```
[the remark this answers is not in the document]
endigo9740: Good call, was a bit oddly worded. I've opted for `onElemLastKeydown`. This follows my standard naming convention.

src/lib/utilities/Dialog/examples/DialogExampleForm.svelte
endigo9740: The code is being reference for documentation purposes, so very relevant! Good catch!

endigo9740:
@ryceg @Sarenor I've made all updates I could. Left comments on the outstanding issues. I welcome additional testing now that it's marked "ready for review". Otherwise we'll aim to merge soon.

src/lib/actions/FocusTrap/focusTrap.ts
ryceg: Yep, NodeLists are different to arrays- `const focusableElems: HTMLElement[] = Array.from(node.querySelectorAll(elemWhitelist);)` would get the job done.

src/lib/utilities/Dialog/stores.ts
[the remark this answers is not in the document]
ryceg: `URL` is actually an object with several utility methods, but it's handy to be explicit up front- we want to be able to handle passing URL objects, in addition to strings that are then interpreted as URLs.

src/lib/utilities/Dialog/stores.ts
ryceg: What type does the `response` value return? You can mark it as a function like `response?(): unknown;`, where the `unknown` is what is returned.

endigo9740: It any be anything, we know what our canned options will provide, but custom dialogs use this as well and can return anything.

ryceg: Does the callback return anything? It looks like it never is, so in that case, you'd be looking at something like: `response?: (arg: unknown) => void;`

src/lib/utilities/Dialog/stores.ts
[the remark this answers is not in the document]
ryceg: This function has a couple issues that would be solved with some more rigorous ESLinting- you're still popping in `any` typings. There is virtually no situation where we should be letting `any`s slip into prod. When you really have to, prefer `unknown`, but it's turning off typescript for lots of code. Additionally, if you use JSDoc syntax, intellisense will provide the comments for the `close`, `trigger`, and `clear` functions which are very useful.

I don't wanna get my tree messy, but here's how it might look: 

```ts
// Dialog Store Queue

import { writable } from 'svelte/store';
import type { DialogSettings } from './types';

function dialogService() {
	const { subscribe, set, update } = writable<DialogSettings[]>([]);
	return {
		subscribe,
		/** Trigger - append to end of queue */
		trigger: (dialog: DialogSettings) =>
			update((dStore: DialogSettings[]) => {
				dStore.push(dialog);
				return dStore;
			}),
		/** Close - remove first item in queue */
		close: () =>
			update((dStore) => {
				if (dStore.length > 0) {
					dStore.shift();
				}
				return dStore;
			}),
		/** Clear - remove all items from queue */
		clear: () => set([])
	};
}

export const dialogStore = dialogService();
```

```
// Dialog Store Queue

import { writable } from 'svelte/store';
import type { DialogSettings } from './types';

function dialogService() {
	const { subscribe, set, update } = writable<DialogSettings[]>([]);
	return {
		subscribe,
		/** Trigger - append to end of queue */
		trigger: (dialog: DialogSettings) =>
			update((dStore: DialogSettings[]) => {
				dStore.push(dialog);
				return dStore;
			}),
		/** Close - remove first item in queue */
		close: () =>
			update((dStore) => {
				if (dStore.length > 0) {
					dStore.shift();
				}
				return dStore;
			}),
		/** Clear - remove all items from queue */
		clear: () => set([])
	};
}

export const dialogStore = dialogService();
```

src/lib/utilities/Dialog/Dialog.svelte
ryceg: Sorry- was shooting from the hip. The issue is essentially that `e.target` is of the `EventTarget` type, which is not limited to `Element`s; other objects can be event targets, too. For example [XMLHttpRequest](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest), [AudioNode](https://developer.mozilla.org/en-US/docs/Web/API/AudioNode), and [AudioContext](https://developer.mozilla.org/en-US/docs/Web/API/AudioContext) are also event targets. *

Given that those types do _not_ have a `classList`, in order to keep dialog from breaking on those (admittedly weird) outliers, we'll need to test to make sure that the target is of an appropriate type. Sling a `if (!(e.target instanceof Element)) return;` in there, and we should be golden. 

This also is the same pattern that is present in `AccordionGroup` :

`if (!(e.currentTarget instanceof HTMLDetailsElement)) return;`


and also in `DocsAppBar.svelte` (line 35):

`if (!(e.target instanceof HTMLInputElement)) return;`


*: [copied from mdn](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget)

PR #515, opened 2022-11-08
Fix console error of route.to.id

niktek:
Fixes the error message for `The navigation shape changed - from.routeId should now be from.route.id`


endigo9740:
Noting that this was due to a breaking SvelteKit version update.

PR #521, opened 2022-11-10
Fixes #519, Tooltip should stay open now as long as the cursor is inside the tooltip element.

### LINKED ISSUE #519: Tooltips disappearing when cursor touches child component (e.g. icon on a button)
### Current Behavior

A tooltip on a button appears when the mouse enters it, but disappears as soon as the cursor moves over the icon contained within that button. As far as I can tell this doesn't affect text-only buttons unless the text is part of another child component such as a `<span>` tag. 

### Steps To Reproduce

The following code should replicate the problem:

```
<button class='btn btn-icon ttaMenu' 
  use:tooltip={{ content: 'Text', position: 'left' }}
>
  <span><UserCircle size='30'/></span>
</button>
```

### Anything else?

Thank you! :-)

Sarenor:
Fixes #519

## What does your PR address?

Fixes tooltips closing when the cursor hovers over a child element of the element with the tooltip action

### DIFF
--- a/src/lib/utilities/Tooltip/tooltip.ts
+++ b/src/lib/utilities/Tooltip/tooltip.ts
@@ -89,7 +89,7 @@ export function tooltip(node: HTMLElement, args: ArgsTooltip) {
 	// -- State
 
 	// On mouse over - show the tooltip
-	const onMouseOver = (): void => {
+	const onMouseEnter = (): void => {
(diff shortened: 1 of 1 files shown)


endigo9740:
@Sarenor Really smart fix here, great job! Thanks!

PR #542, opened 2022-11-14
chore: upgraded deps

niktek:
Dep upgrade and quick run through docsite - no new warnings from SK.


endigo9740:
@niktek Seems to be working well enough, so I'll go ahead and merge. We'll test drive this as dev efforts continue the next couple weeks.

Issue #528, opened 2022-11-13
Update Drawers to include their own store


endigo9740:
### Describe what feature you'd like. Pseudo-code, mockups, or screenshots of similar solutions are encouraged!

Should allow for the following:

- Controlling visibility
- Passing style parameters (position, margins, etc)
- Allow you to configure what is shown in the drawer (swap out forms, etc)

### What type of pull request would this be?

Enhancement

### Any links to similar examples or other references we should review?

The store should work similar to toasts/dialogs.

endigo9740:
Closed in favor of https://github.com/skeletonlabs/skeleton/issues/609

PR #580, opened 2022-11-19
Toast store implementation change

### LINKED ISSUE #522: Unsafe store behaviour of utility components in SvelteKit
### Current Behavior

Currently components like Toast use [exported const stores ](https://github.com/Brain-Bones/skeleton/blob/dev/src/lib/utilities/Toast/stores.ts#L47) to function. This is completely fine for all uses cases outside of SvelteKit, as any const variable imported from a library will become global/shared state between requests. This could and will cause unintended behaviour when using SvelteKit. 
I haven't spent time to properly test what the impact would be in skeleton's case, so I don't know what would break. 

Potential solutions were discussed in discord, also the way Drawers are written should mean that they are safe to use, it looks like only Dialog and Toast are affected.

### Steps To Reproduce

Import any const store in sveltekit with SSR enabled.

### Anything else?

[Dicussion about shared state/stores and SvelteKIt](https://github.com/sveltejs/kit/discussions/4339#discussioncomment-2384978) 
[Issue about shared state behaviour](https://github.com/sveltejs/kit/issues/1542)
[Another related issue](https://github.com/sveltejs/kit/issues/2213)

buni:
## Before submitting the PR:
- [ ] Does your PR reference an issue? If not, please [chat to the team on Discord](https://discord.gg/EXqV7W8MtY) or [GitHub](https://github.com/Brain-Bones/skeleton/discussions) before submission.

## What does your PR address?

Potential "fix" for the problem described in [Fixes #522 ](https://github.com/Brain-Bones/skeleton/issues/522) this shouldn't change the default component behaviour at all. It also allows SvelteKit user to set and manage their own stores, essentially mitigating any shared state concerns. To manage global stores in SvelteKit the best solution is to put them inside component [context](https://svelte.dev/tutorial/context-api) . This PR has a halfway solution, the best one would be to not have const global stores, but that would change how the components work (you'll always have to bind a store), I assume this is not desired behaviour. One caveat with this solution is that there is still a global variable, but by overriding it all should be fine (I'm not exactly sure how to test variable lifecycles in SvelteKit).

### Tips
- Link to an issue using the verbiage [Fixes #522 ](https://github.com/Brain-Bones/skeleton/issues/522)

### DIFF
--- a/src/lib/utilities/Toast/Toast.svelte
+++ b/src/lib/utilities/Toast/Toast.svelte
@@ -13,7 +13,8 @@
 	export let max: number = 3;
 	/** The duration of the fly in/out animation. */
 	export let duration: number = 150;
-
+	/** Toast store used for the component, by default the const toastStore is used. */
+	export let store: any = toastStore; // the store is still a global variable, but there is the option to override it. If its overriden it *should* no longer pose a problem, although I'm not sure what the best way to test global server-side variable lifecycle is in sveltekit 
 	// Props (styles)
 	/** Provide classes to set the background color. */
 	export let background: string = 'bg-accent-500';
@@ -62,8 +63,8 @@
 	}
 
 	function onAction(): void {
-		$toastStore[0].action.response();
-		toastStore.close();
+		$store[0].action.response();
+		store.close();
 	}
 
 	// Reactive
@@ -72,12 +73,12 @@
 	$: classesBase = `${cToast} ${background} ${width} ${color} ${padding} ${spacing} ${rounded} ${shadow}`;
 </script>
 
-{#if $toastStore.length}
+{#if $store.length}
 	<!-- Wrapper -->
 	<div class="snackbar-wrapper {classesWrapper}" data-testid="snackbar-wrapper">
 		<!-- List -->
 		<div class="snackbar {classesSnackbar}" transition:fly={{ x: animAxis.x, y: animAxis.y, duration }}>
-			{#each $toastStore as t, i}
+			{#each $store as t, i}
 				{#if i < max + 1}
 					<!-- Toast -->
 					<div class="toast {classesBase} {t.classes}" role="alert" aria-live="polite" data-testid="toast">
@@ -85,7 +86,7 @@
 						<!-- prettier-ignore -->
 						<div class="flex items-center space-x-2">
 							{#if t.action}<button class="btn {buttonAction}" on:click={onAction}>{@html t.action.label}</button>{/if}
-							<button class="btn-icon {buttonDismiss}" on:click={() => { toastStore.close(t.id) }}>✕</button>
+							<button class="btn-icon {buttonDismiss}" on:click={() => { store.close(t.id) }}>✕</button>
 						</div>
 					</div>
 				{/if}
--- a/src/lib/utilities/Toast/stores.ts
+++ b/src/lib/utilities/Toast/stores.ts
@@ -11,20 +11,33 @@ function randomUUID(): string {
 	return Number(random).toString(32);
 }
 
-// If toast should auto-hide, wait X time, then close by ID
-function handleAutoHide(toast: ToastSettings): void {
-	console.log(toast);
-	if (toast.autohide === true) {
-		setTimeout(() => {
-			toastStore.close(toast.id);
-		}, toast.timeout);
-	}
-}
-
-function toastService(): any {
+export function toastService(): any {
 	const { subscribe, set, update } = writable([]);
+	/** Remove first toast in queue */
+	function close(id: string) {
+		update((tStore) => {
+			if (tStore.length > 0) {
+				var index = tStore.findIndex((t: ToastSettings) => t.id === id);
+				tStore.splice(index, 1);
+			}
+			return tStore;
+		});
+	}
+
+	// If toast should auto-hide, wait X time, then close by ID
+	function handleAutoHide(toast: ToastSettings): void {
+		console.log(toast);
+		if (toast.autohide === true) {
+			setTimeout(() => {
+				close(toast.id!);
+			}, toast.timeout);
+		}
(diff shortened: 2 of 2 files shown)


endigo9740:
@buni We're juggling a lot right now, so I may have to leave this PR idle for the time being. I will return and look asap though. Thanks for your patience in the meantime.

buni:
No worries

endigo9740:
@buni I've finally had a chance to review this. So what you're proposing is replacing the built-in store with a user-provided store, or at least providing the opportunity for such.

I'm really conflicted on this. Sure it may solve the issue raised, but my concern is how this impacts the DX for Skeleton. We already get a lot of flak for using stores through the library, as misguided as it may be, so I'm not this extra hoop to jump through will sit well with most end users.

I think we're approaching this the wrong way though. Per the original discuss thread on this issue, several Svelte/Kit members summarize this well - it's a usage issue. If users are using stores in this very particular context (SSR/load) then sensitive information can be exposed.

But thinking about it in the context of our library, these stores are used for modals, toasts, and (eventually) drawers. I'm struggling to find a real world use case where folks might submit data in this fashion. Additionally, I think rather than trying to solve this problem FOR users, and compromising our DX, we instead take the approach of informing users and let them decide how they want to handle this.

Essentially this may mean that folks can't use these Skeleton features for this particular use case - triggered within a SSR load method. But I kind of feel that's an anti-pattern to begin with.

I'm going to open this up to review by a few of the other core contributors. We'll follow the consensus there.

buni:
In terms of DX, it doesn't really change how the components are used at the moment, you have the option to not pass your own store. It doesn't break the API in that sense. Why I'm concerned about using stores in the out the box way is its wrong, while it might not seriously impact skeleton's components in particular, it can lead to bugs like showing an alert to the wrong user or similar ones. 

As a user of the library I would like to be informed about the potential issues with using global variables with SvelteKit, and have an option to not use them. Then I can decide with which approach to go, use the default store or manage my own. Again I really don't see any downside with the way this PR deals with the store issue.

endigo9740:
I see, so you leave the default store in place and keep it as the default for the prop. Then the prop handles the job from then on. This means the current behavior persist. I follow now.

In that case this may work. But I think we're both in agreement here that we need more than just the code change, we need this to be written out explicitly in the docs, so users are made aware of the risk. You'll note this is a requirement for any PR we accept per the checklist in the original message above:

Screen Shot 2022-11-25 at 2 37 42 PM

Here's the set of requirements I'm going to lay out for this:

- Mirror this change for Modals, since the store works in an identical fashion
- We need to update the Drawer component to include it's own store per #528, then follow this pattern as well
- If we're going to use this branch, we need it current with the recent bugfix changes in #587
- Then finally, we need to document the availability and purpose of this change directly in the Usage section of the docs. Probably near the bottom since it's only for select use cases.

I was just about to start work on #528, so I'm going to propose we just go ahead and handle all these changes in a new branch/PR for which I will handle. I'll keep your PR open as reference, but implement all these changes. Once ready I'll close this PR and merge the new changes.

I'll ping you when this new PR is ready for review so you can provide feedback on the changes, doc wording, etc. Work for you?

buni:
Sounds good to me, I'm not too fussed about it, and this PR was always intended as a proposal.

endigo9740:
Sounds good. Thanks for the proposal then.

Btw, unrelated to above - are you a member on our Discord? If so what's your usename? I chat to so many folks here and there it can sometimes be hard to keep everyone straight. But I was asking specifically as you're a sponsor and I wanted to fill you in on some upcoming changes. We hav a dedicated channel for contributors, I can get you added and explain.

endigo9740:
@buni I haven't forgot about this stuff. We're just prepping for some big updates dropping Tuesday. As soon as that's out the door this will be one of the first items on my todo list!

endigo9740:
@buni I'm currently in-progress on the new Drawer rework here:
https://github.com/skeletonlabs/skeleton/pull/614

I was just about to implement your suggestion for the `store` prop to allow users to override this. However, I feel this includes a major flaw. We aren't using vanilla stores for Drawers, Modals, and Toasts. We're using modified store that includes custom methods:

- Toast Store: https://github.com/skeletonlabs/skeleton/blob/dev/src/lib/utilities/Toast/stores.ts
- Modal Store: https://github.com/skeletonlabs/skeleton/blob/dev/src/lib/utilities/Modal/stores.ts

It will not be trivial for users to override these, so this gives me some pause. Technically they could clone and replace our stores. But I'm not sure if everyone will understand how these operate compared to a basic writable store.

I'm going to hold off on this change until we've had more time to review and discussion.

endigo9740:
@buni FYI I've been discussing this issue with a couple contributors on our Discord. @AdrianGonz97 has pointed out that there is an odd data persistence issue for toasts if you trigger them immediately within a route init. Not even just strictly related to a load function, just a plain old route script tag init.

Refresh this page and you'll briefly see a queue of multiple messages, even though the toasts and store have no persistence built in:
https://global-store-test.vercel.app/

His response explains this well:
> so i think it's related to what Buni's issue was talking about. because it's a global store and the toast is being triggered in SSR, the store on the server is mutating which is why we get that persistence across different requests

So for the time being we're leaning towards no modifying the codebase, awaiting an official fix from SK's end, but still adding a message or warning on the docs for Toasts, Dialogs, and Drawers that says something to the effect of:

1. Don't use a global store in a SvelteKit app within a Load function
2. If you need a overlay feature triggered on event, put it inside a Browser context `if (browser) ...`

Then we can link to the original discussion to provide more context. It's a duct tape fix, but probably the best we can do for the time being. I just really don't think replicating our stores are practical for the majority of our users.

endigo9740:
I'm going to go ahead and close this PR, but will make reference to it in the upcoming ticket for the proposed messaging change.

PR #589, opened 2022-11-24
Links, GH, package refs change except Trey

niktek:
I've changed the sponsorship fetch url to be correct, but it fails for now until sponsorship is fixed on the org.
NPM Package link won't work till we actually publish a new package.

Looks like format whacked a few of the modified files

Only references to Brain should be Trey's stuff on the home page.

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,6 +1,6 @@
 # These are supported funding model platforms
 
-github: [skeletonlabs]
+github: [Brain-Bones]
 patreon: # Replace with a single Patreon username
 open_collective: # Replace with a single Open Collective username
 ko_fi: # Replace with a single Ko-fi username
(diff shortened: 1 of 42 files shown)


endigo9740:
@niktek Looks good, but since it modifies the package to `@skeletonlabs/skeleton` I'm going to hold off on a merge until that's a package that actually exists!

endigo9740:
@niktek I've gone ahead and revised a couple sponsorship related items here:

1. Reverted the `FUNDING.yml` to point to Brain & Bones until we can get the new org squared away
2. Disabled the HTTP call for the list of sponsors on the homepage. Replacing with an email link temporarily.

We'll revise these as soon as we have the new sponsorship stuff sorted. I'm going to go ahead and merge, so this commits us to getting the new package setup on NPM asap.

PR #595, opened 2022-11-26
Fixes required for monorepo to work

niktek:
## What does your PR address?

Currently no one can work on both Skeleton and their own project at the same time, unless their current project is the docs site.  This PR finally fixes that.

You can test it by checking out this repo https://github.com/niktek/minimono

For the following steps pnpm is **NOT** optional - pnpm also does **NOT** become a requirement for Skeleton itself, this is a completely separate repo for seeing how it works. Having all 3 package managers installed at the same time causes no problems - the create-skeleton-app testing has confirmed this.  The version of Skeleton in `minimono` is a copy with the changes in this PR already applied.

```
pnpm i
pnpm dev
```

Both Skeleton and Newsite projects are launched on separate ports.

Newsite's package.json shows @skeletonlabs/skeleton as a workspace reference and can therefore import anything as an end dev would from the published package.  NOTE: VSCode will show red squiggles for component imports - it is lying.

Editing anything in Skeleton will now update Newsite's usage of that thing.  e.g. change the default `width` in packages/skeleton/src/lib/components/Avatar/Avatar.svelte and the Avatar on Newsite updates.
Likewise editing the theme css in Skeleton (vintage in this case), flows through to Newsite - this is the main thing that was killing all prior mono repo attempts.

# How does this work?
The missing ingredient is violence. (j/k)

As usual it's a number of things:
## package.json 
During development, SK and Vite shield you from setting up a correct package.json with the appropriate "exports" field.  They "fix" this on build and insert those entries into the package/package.json file.  This is of course useless for other projects that are referring to the dev package.json file.  Unfortunately `svelte-kit package` is respectful of any entries you already have in the dev package.json and copies them over into the package version without adjusting the paths, this results in a broken published package.  To fix this required the addition of two scripts - one to strip the "exports" before `svelte-kit package` and one to re-insert the "exports" after build.

## Aliases
This was the spanner in the works all along.  While it's a great helper for a single project with an internal library, it just totally stuffs things up for multiple projects working together.  The key thing to note here is that each project's Vite server has their own set of aliases, so Newsite's Vite would find a component in Skeleton that used $lib and try to resolve it within it's own filesystem space, where of course it would not find Skeleton's actual components and just fail.  The proper place for package resolution is at the package manager layer, not trying to fake it at Vite's level.

**This requires that everything in Skeleton's src/lib/ folder uses relative path addressing**

Fortunately there were only a few cases where this was in use and this PR fixes the remaining ones.

## Shut up Vite
Vite would often complain about `The request url "blah blah" is outside of Vite serving allow list.`.  Even though it seemed to be working fine, it was annoying to see this spammage on the console.  The additional config in Newsite's `vite.config.ts` turns off those warnings and the onus is on the consuming project to add that.

## What about doc site ?
It remains tightly coupled to Skeleton itself.  On one hand, it's helpful that it remains with Skeleton because any additions to Skeleton need to have documentation created.  On the other hand, it's just going to become harder and harder to separate - mainly due to autocomplete sticking in references to components using the $lib alias, which as outlined above will break docsite when it moves into it's own project.  It also increases the likelihood of affecting the Skeleton package through accidental inclusion of deps/code or a docsite package having a breaking change that prevents publishing.

Once we are happy that this mono repo approach is working how we want, we should break out doc site sooner rather than later.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
 	"scripts": {
 		"dev": "vite dev",
 		"build": "vite build",
-		"package": "svelte-kit sync && svelte-package",
+		"package": "node ./scripts/pre-build.js && svelte-kit sync && svelte-package && node ./scripts/post-build.js",
 		"patch-package": "npm version patch && svelte-kit package",
 		"publish": "npm publish ./package",
 		"preview": "vite preview",
@@ -18,9 +18,9 @@
 		"coverage": "vitest run --coverage"
 	},
 	"devDependencies": {
-		"@bobthered/tailwindcss-palette-generator": "^3.1.1",
+		"@bobthered/tailwindcss-palette-generator": "^3.1.0",
 		"@sveltejs/adapter-auto": "^1.0.0-next.89",
-		"@sveltejs/kit": "^1.0.0-next.561",
+		"@sveltejs/kit": "^1.0.0-next.560",
 		"@sveltejs/package": "1.0.0-next.5",
 		"@tailwindcss/forms": "^0.5.3",
 		"@testing-library/dom": "^8.19.0",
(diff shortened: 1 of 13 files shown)


endigo9740:
@niktek With the big updates out the door I figure we give this a proper review either later tonight or tomorrow. Let me know and I'll set aside some time for us to discuss.

Issue #538, opened 2022-11-14
Data Table feature


endigo9740:
> **Warning**
> This feature is currently delisted per: https://github.com/skeletonlabs/skeleton/issues/930. It will return in the future, but no ETA has been provided yet.
> We are accepting recommendations for alternatives here: https://github.com/skeletonlabs/skeleton/issues/821

---

### Describe what feature you'd like. Pseudo-code, mockups, or screenshots of similar solutions are encouraged!

Here's a running list of features and improvements we would like to added to data tables going forward.

- [ ] Review options to sanitize data table content  (missing object keys should be set to `undefined`)
- [ ] Review options to add row selection in addition to cell selection for tableA11y action
- [ ] Evaluate the use of `role="gridcell"` for a11y keyboard interaction. Goes against ARIA guidelines.
- [ ] Consider adding search whitelist, to limit search certain fields
- [ ] Allowing specifying the default sort on load (runs + UI arrow added)
- [ ] Quick mapping of available columns (to add/remove columns on demand)
- [ ] Consider adding [3-stage checkbox](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#indeterminate_state_checkboxes) for select all, including: current page, all results, no results ([Tailwind selector](https://tailwindcss.com/docs/hover-focus-and-other-states#indeterminate))
- [ ] Sticky header: https://twitter.com/CodyWebHouse/status/1592907092226060288
- [ ] Consider support for sorting complex cell data (ex: objects, arrays, etc)

Known Issues:

- Search + Pagination are not working as expected

### What libraries have been reviewed

* [Tanstack](https://github.com/olifolkerd/tabulator) - not recommended, documentation is poor
* [Tabulator](https://tabulator.info/) - data-driven, does not allow for template manipulation
* [Svelte-Headless-Tables](https://github.com/bryanmylee/svelte-headless-table) - documentation has gaps, poor accessibility
* [Svelte Simple Datatables](https://github.com/vincjo/datatables) - recommended, but have not extensively tested extensively

endigo9740:
NOTE: additional suggestions are welcome here!

ak4zh:
Add row grouping
https://datatables.net/examples/advanced_init/row_grouping.html

younesk31:
A more convenient current page selection, since if the table has multiple pages its less suitable to use the arrows alone
[...]

PR #599, opened 2022-11-29
Feat: Reworking Data Tables

### LINKED ISSUE #613: Data Table search only searches the current pagination
### Current Behavior

When searching, only the rows that are on the current pagination are filtered by the search.

### Steps To Reproduce

1. Go here https://www.skeleton.dev/utilities/data-tables 
2. Search `qui est esse` - row ID 2 should be the only one showing
3. Click on the paginator for the next page, row ID 2 disappears.

### Anything else?

_No response_

AdrianGonz97:
## Before submitting the PR:

## What does your PR address?

- Added better types for Data Tables
- Added a function `createDataTableStore` that constructs and returns the svelte writable store
- Moved `sortState` to be part of the data table store
- Added types to `/utilities/data-tables/+page.ts` page load for improved type safety
- Updated docs for `createDataTableStore`

### Why create `createDataTableStore` ? 
This function provides a super convenient, typescript and intellisense friendly way to create the data table store for users. Users won't have to import any types, such as `DataTableModel` or `Writable` or even the type of their `Source` data as typescript will infer everything from the types passed into the function - something that's *especially* helpful for users that don't use TS. 

Additionally, I noticed that properties, such as `filtered`, should only be *derived* from the `source` data, so having it be part of the `DataTableModel` properties during the svelte store declaration didn't really make much sense to me.  `createDataTableStore`  also makes it easier to add additional properties to rows, such as `dataTableChecked`. One of the main reasons to do it this way was because, previously, we were *mutating* the `source` data in order to add these properties. With the current implementation, we are no longer mutating the source.  

Lastly, `sortState` was a global instead of being part of a store.

### Additional note: 
This would be a breaking change for data tables as `createDataTableStore` will now be needed to create the store.

## Overview
- [x] Better type safety
- [x] Refactored writable store into a custom store
- [x] Custom store is created using `createDataTableStore`
- [x] Update the docs
- [x] Fixes #613


endigo9740:
@AdrianGonz97 This seems interesting, just please note we do have a strict policy for non-trivial updates per:

> 3. Review with core contributors to define the requirements before you write your first line of code.

While asinine, this helps prevent redundant and duplicate effort. Thankfully we're not touching data tables at the moment, so this is fine, but this very well could have overlapped with issues such as this one:

https://github.com/skeletonlabs/skeleton/issues/538

Common courtesy here is to discussion either in a GitHub Issue or the contributor channel on Discord in the future please.

That said, tomorrow is going to be swamped, so I'd expect the first review for this will be Wednesday (US central time) at the earliest. Expect more feedback when I complete a full review then. Thanks!

AdrianGonz97:
> While it may seem asinine, this helps prevent redundant and duplicate effort. Thankfully we're not touching data tables at the moment, so this is fine, but this very well could have overlapped with issues such as this one:
> 
> * [Data Table feature expansions #538](https://github.com/skeletonlabs/skeleton/issues/538)
> 
> Common courtesy here is to discuss either in a GitHub Issue or the contributor channel on Discord in the future please.

@endigo9740 Very sorry about that! I completely missed that issue 🤦‍♂️. I'll make sure it won't happen again!

endigo9740:
@AdrianGonz97 With the new announcements out the door I've finally had a chance to review this. Unfortunately I'm seeing a couple errors right out of the gate:

Screen Shot 2022-11-29 at 1 01 37 PM

Screen Shot 2022-11-29 at 1 03 23 PM

Additionally, I make no claims to be an expert in Typescript. I learned the basics needed for Angular years and years ago but I haven't really kept up with the changes over the years. Plain and simple, these updates are over my head.

I think in order to move forward with this I'll either need to get a second opinion from someone I trust, such as @ryceg or you and I will need to devote some time to discussing these changes. It's my responsibility to ensure we're introducing changes that positively benefit end users, and while I think these changes will do that, I am not in the position to adequately judge the quality of the changes here.

A few things I don't understand specifically that perhaps you can clarify:

- I believe `<T>` is a generic type reference, but I have zero working knowledge of this. I'm going to need the use case of this one explained to me.
- I don't understand the use of `extends`, ex: `<T extends Record<PropertyKey, unknown>>`
- I'm having trouble locating the source of `PropertyKey`...is that a custom interface?
- I feel this is something rudimentary that I've completely missed, but I don't understand the type declaration on the functions themselves. Here's a contrived example:

```js
function doSomething<whatDoesThisDo?>(key: type): returnType { ... }
```

We already have the IO covered (params going in are typed, response value coming out is typed)....what does this typing give us?

Per the overall logic:

- I appreciate the goal of `createDataTableStore` is assist users that don't use Typescript, but I don't have a clear understanding of what was preventing users from doing this with the current implementation. Could users not just [leave off the type here](https://github.com/skeletonlabs/skeleton/blob/dev/src/routes/(inner)/utilities/data-tables/%2Bpage.svelte#L48)?
- I do like that it simplifies the initialization of a table, that source is only provided once, and options are clearly separated. But I think I need to look how this functionally works. Imagine a scenario where the source updates periodically - will this allow folks to easily update the source/filtered at will? This question has come up a lot since the refactor went live.

Given that it seems there's a lot to discuss here let me know how you would like to proceed. I'll try to make myself available on Discord later today if we want to review there.

AdrianGonz97:
> I believe <T> is a generic type reference, but I have zero working knowledge of this. I'm going to need the use case of this one explained to me.

We use generics here because we don't know the data type of the `source`. By using generics, we can retain our type safety by simply saying that we expect an `Array` of `T` as the `source` and `filtered`. 

By doing so, we get awesome intellisense when accessing the store like so: 
awesome intellisense gif

Where previously, we don't get any hints:
sad no intellisense gif

---
> I don't understand the use of extends, ex: <T extends Record<PropertyKey, unknown>>

In the places where we use `<T extends Record<PropertyKey, unknown>>`, we are telling Typescript that the generic type `T` is a type of object. If we don't, we woudn't be able to use _object specific_ things like:
```ts
Object.values(obj)
// or 
obj[key]
```
If we were to leave it as `T`, primitive types like `number` and  `string` would be allowed, which would cause errors as they can't be used like those shown above.

---
> I'm having trouble locating the source of PropertyKey...is that a custom interface?

`PropertyKey` is a handy type alias that Typescript provides that represents all the possible types that a `key` of an object can be. If we examine it, it's just a union type:
```ts
type PropertyKey = 'string' | 'number' | 'symbol';
```

---
> I feel this is something rudimentary that I've completely missed, but I don't understand the type declaration on the functions themselves. Here's a contrived example:

> ```js
> function doSomething<whatDoesThisDo?>(key: type): returnType { ... }
> ```

Practically the sane explanation in the second reply of this message. We use the angled brackets `<>` in function signatures for type narrowing. In `<T extends Record<PropertyKey, unknown>>`, we are _narrowing the type_ of `T` to only accept types that are **objects**. 

To provide, perhaps a clearer example, consider the following: 
```ts
function sayMyName<T extends { name: string }>(obj: T) {
    console.log(obj.name);
}
```
`<T extends { name: string }>` is requiring the type that is passed into the function `sayMyName` **must** have a property, `name`, that is of type `string`. Otherwise, Typescript will have a type error.

```ts
// this is not valid
const foo = {
    someOtherProp: 40,
};
sayMyName(foo); // typescript will error

// this is valid
const bar = {
    name: 'Adrian',
    someOtherProp: 45,
};
sayMyName(bar); // typescript will be satisfied
```

---
> * I appreciate the goal of `createDataTableStore` is assist users that don't use Typescript, but I don't have a clear understanding of what was preventing users from doing this with the current implementation. Could users not just [leave off the type here](https://github.com/skeletonlabs/skeleton/blob/dev/src/routes/(inner)/utilities/data-tables/%2Bpage.svelte#L48)?

They could leave off the type there and just do 
```ts 
const dataTableModel = writable<DataTableModel>({...});
``` 
but they would still need to import the `DataTableModel` type in order to get any kind of type checking. Because of this, users that don't use Typescript won't have any type checking at all since they can't import types. It's also a whole lot cleaner 😝

---
> * I do like that it simplifies the initialization of a table, that source is only provided once, and options are clearly separated. But I think I need to look how this functionally works. Imagine a scenario where the source updates periodically - will this allow folks to easily update the source/filtered at will? This question has come up a lot since the refactor went live.

`source` and `filtered` are still mutable by way of the store. If someone want's to change those values, they just have to access it by doing this: `$dataTableStore.source = ...` or `$dataTableStore.filtered = ...`

I can modify this implementation to allow the `source` to be mutated when the original `source` that's passed in is changed. I've got a couple of ideas to make that happen.

---

Here's a little example of `createDataTableStore` in use. Everything is inferred and no types have to be passed in to achieve type safety. I especially really like having autocomplete for `sort` (please excuse my slow typing):
fun intellisense autocomplete


**Side note:** Writing this out made me realize that I should add an `extends` on the data model type.

endigo9740:
Hey @AdrianGonz97 I noticed the new updates over the weekend. Note the PR is currently in the "draft" state. You'll want to tap the "ready for review" button on this page to notify me when this is ready for another review. No rush, just wanted to confirm status here.

AdrianGonz97:
@endigo9740 Yea I'm still working on it. Currently trying to figure out a "clean" way to have the source data updated. Maybe you can lead me down the right path 😄. Right now I have 2 methods that _can_ work:

Consider that we have following data table:
```ts
const posts = [ { id: 1, title: 'some post title', body: 'some post body' }, ...etc ];

// Store
const dataTableStore = createDataTableStore(posts, {
    sort: '',
    search: '',
    pagination: { offset: 0, limit: 5, size: 0, amounts: [1, 2, 5, 10] }
});

dataTableStore.subscribe((store) => dataTableHandler(store));
```

The first way would be to update the store's `source` with the new data:
```ts
// posts gets updated
posts.push({ ...someNewPost }); 

// we reassign the store's source with the newly updated data
$dataTableStore.source = [ ...posts ]; // NOTE: we must use assignment operator to have the table rerender
```

Alternatively, we can provide a function to update the source for them:
```ts
// posts gets updated
posts.push({ ...someNewPost }); 

// pass the store and the new source array
dataTableUpdateSource(dataTableStore, posts); // name of the function can be changed since names are hard
```

The latter is more handholdy and may look unnecessary, but it allows us to not have to remap the entire source array **every** time _something_ in the store changes. 

The first method would require `dataTableHandler` to look like this:
```ts
export function dataTableHandler<T extends Record<PropertyKey, any>>(store: DataTableModel<T>): void {
	store.base = store.source.map((row, i) => {
		return { ...row, dataTableChecked: store.base[i]?.dataTableChecked ?? false };
	});
	// Then
	searchHandler(store);
	selectionHandler(store);
	sortHandler(store);
	paginationHandler(store);
}
```

endigo9740:
@AdrianGonz97 Taking a step back before I address your question, I like this concept of us owning the store instance. I wonder, rather than creating a vanilla store and living with the constraints of that, we don't create a custom store with it's own built in methods. Something like what I do for Toasts/Modals/Drawers:

https://github.com/skeletonlabs/skeleton/blob/dev/src/lib/utilities/Toast/stores.ts
(has trigger, close, clear methods)

https://github.com/skeletonlabs/skeleton/blob/dev/src/lib/utilities/Modal/stores.ts
(has trigger, close, clear methods)

https://github.com/skeletonlabs/skeleton/blob/dev/src/lib/utilities/Drawer/stores.ts
(has open/close methods)

So with this approach in mind, if we can build custom stores with a unique method like like `store.update(newSourceData)` that would allow the store to be updated in a very succinct and intuitive fashion.

So technically a third option to your other two proposals, but I think users would understand this. Plus this could be setup with type handling to auto-suggest the method for discoverability and auto-complete. Thoughts?

AdrianGonz97:
@endigo9740 haha, a couple hours after writing that post I considered writing a custom store as well to try it out, but haven't yet. It's certainly something that could work so I'll give it a shot!

endigo9740:
Great minds think alike, right?! I think honestly it's the way to go. They're much easier to setup that you might think, just make sure to pass through the default methods for set/update/subscribe. In hindsight that's 100% how I would have built this.

endigo9740:
@AdrianGonz97 thinking about this a bit more - to ensure scope doesn't grow too large. Let's focus this PR on doing three things:

1. Introducing your type improvements - most of which are ready I believe
2. Introduce the concept of generating a store with the `create` method - ready too
3. Introduce the custom store idea with the new `update()` method

I think with that we can merge and then decide how we want to expand further. Don't be afraid to make more notes here as ideas come to mind:
https://github.com/skeletonlabs/skeleton/issues/538

I think with these limitations we can at least get a first draft of these updates up, documented, and tested before the next release in a week or so from now. Then I can rest easier knowing these changes are in before the Svelte Sirens stream, given we might see a big influx of folks from that!

AdrianGonz97:
Since we've made a custom store now, we are able to add all of the functions that manipulate the store directly into it. No need for additional imports. Super clean and concise!

endigo9740:
@AdrianGonz97 Awesome, for some reason GitHub didn't notify me of your comment so glad I just saw this.

I'll review now! Totally cool to wait on the docs until we feel the changes are finalized though.

endigo9740:
I've done my review and added couple minor comments and code changes. Nothing worth detailing here.

Additionally, these are items I've noticed that will need to be addressed. Probably bugs in my original logic:

- BUG: To to the second page, search "qui est". This should probably set the pagination back to the first page by default.
- BUG: Search is local to only the current filtered results, such as current page, so it doesn't span multiple pages

I'd say let me know when the doc changes are ready and I'll do one final pass, otherwise I'd say this is good to go. I like how much less there is for the end user to juggle here!

AdrianGonz97:
@endigo9740  Thanks for the review! 

> * BUG: To to the second page, search "qui est". This should probably set the pagination back to the first page by default.

I just pushed a fix for this, a very simple 1 liner 😄. Let me know if this is now working as expected.

> * BUG: Search is local to only the current filtered results, such as current page, so it doesn't span multiple pages

Are you referring to the same bug as #613? If so, then that _should_ be working now since it looks like it's working in the [preview](https://skeleton-docs-od6igzw6e-skeleton-labs.vercel.app/utilities/data-tables), unless I'm misunderstanding and you're talking about something else.

endigo9740:
@AdrianGonz97 awesome. I'm going to try to finish up the design token ticket and then this will be next on my list.

endigo9740:
Noting this here before the PR is merged. I've implemented a final round of documentation copy edits. This includes a new "experimental" badge added to the left hand navigation, as well as an alert at the top of the page. This links to this project as an alternative:

https://svelte-headless-table.bryanmylee.com/#headless

The goal is to set expectations that you can use this, but it's subject to on-going updates and changes. It is not what I would call production-ready yet. The hope this will funnel feature requests and contributions to the "feature expansion" ticket here:

https://github.com/skeletonlabs/skeleton/issues/538

Screen Shot 2022-12-07 at 4 38 18 PM

That aside, this is approved and ready to merge! Great job @AdrianGonz97

PR #611, opened 2022-11-30
Chore/eslint mega fix

niktek:
## Before submitting the PR:
Few that need to stay - part escapes, part empty interfaces in ambient app.d.ts file generated by SK

All the 109 warnings need better types to replace `any`


niktek:
Any notes from the deploy would be helpful - it's seemingly running all fine on local which is making it harder to track down

PR #614, opened 2022-11-30
Drawer Updates

### LINKED ISSUE #609: Updates to the Drawer utility feature

I'm merging a handful of small tasks together here:

- [x] Drawers should have their own store (like Modals/Toasts). This will control open/close state as well as content
- [x] Allow prop setting overrides an customization per drawer instance
- [x] Bugfix for `window undefined` error if Drawer is on at page load
- [ ] Allow the "built-in" store to be overwritten per #580

Documentation and tests will also need to be updated with these changes.

Enhancement

- https://github.com/skeletonlabs/skeleton/issues/528
- https://github.com/skeletonlabs/skeleton/issues/585
- https://github.com/skeletonlabs/skeleton/pull/580

endigo9740:
Fixes #609

### DIFF
--- a/src/lib/utilities/Drawer/Drawer.svelte
+++ b/src/lib/utilities/Drawer/Drawer.svelte
@@ -1,18 +1,19 @@
 <script lang="ts">
 	import { fade, fly } from 'svelte/transition';
-	import { writable, type Writable } from 'svelte/store';
+	import { onMount } from 'svelte';
 
+	// Actions
 	import { focusTrap } from '$lib/actions/FocusTrap/focusTrap';
 
+	// Drawer Utils
+	import type { DrawerSettings } from '$lib/utilities/Drawer/types';
+	import { drawerStore } from '$lib/utilities/Drawer/stores';
+
 	// Props
-	/** Provide a store to manage visible state.
-	 * @type {Writable(boolean)}
-	 */
-	export let open: Writable<boolean> = writable(false);
 	/** Set the anchor position.
 	 * @type {'left' | 'top' | 'right' | 'bottom'}
 	 */
-	export let position = 'left';
+	export let position: string = 'left';
 	/** Define the Svelte transition animation duration.*/
 	export let duration = 150;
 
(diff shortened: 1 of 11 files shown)


src/lib/utilities/Drawer/types.ts
```diff
@@ -0,0 +1,48 @@
+// Drawer Types
+
+export interface DrawerSettings {
+	open?: boolean;
+	/** A unique identifier, useful for setting contents. */
+	id?: string;
+	/** Set the anchor position.
+	 * @type {'left' | 'top' | 'right' | 'bottom'}
+	 */
```
AdrianGonz97, line 9: Looks like this was meant to be deleted since position was moved below

src/lib/utilities/Drawer/types.ts
```diff
@@ -0,0 +1,48 @@
...
+	/** Pass abitrary information for your own persona use. */
+	meta?: any;
+
+	// NOTE: the following must style in sync with the drawer props.
+
+	// --- Core ---
+	/** Set the anchor position.
+	 * @type {'left' | 'top' | 'right' | 'bottom'}
+	 */
+	position?: string;
```
AdrianGonz97, line 19: Since `position` is meant to only be one of these `'left' | 'top' | 'right' | 'bottom'` we should make the type a string union.
```suggestion
	/** Set the anchor position.
	 * @type {'left' | 'top' | 'right' | 'bottom'}
	 */
	position?: 'left' | 'top' | 'right' | 'bottom';
```

src/lib/utilities/Drawer/Drawer.svelte
```diff
@@ -1,45 +1,46 @@
...
-	/** Provide a store to manage visible state.
-	 * @type {Writable(boolean)}
-	 */
-	export let open: Writable<boolean> = writable(false);
 	/** Set the anchor position.
 	 * @type {'left' | 'top' | 'right' | 'bottom'}
 	 */
-	export let position = 'left';
-	/** Define the Svelte transition animation duration.*/
+	export let position: string = 'left';
```
AdrianGonz97, line 16: This should also be a string union

src/lib/utilities/Drawer/Drawer.svelte
```diff
@@ -48,73 +49,97 @@
...
+	// prettier-ignore
+	const propDefaults = {
+		position, duration,
+		bgBackdrop, blur,
+		bgDrawer, border, rounded, width, height, margin,
+		labelledby, describedby
+	};
+
+	// Listen to drawerStore updates
+	drawerStore.subscribe((settings: DrawerSettings) => {
```
AdrianGonz97, line 67: You can remove `DrawerSettings` type. The `settings` param should be inferable from there

AdrianGonz97 (commented):
Just a couple of notes related to types

Issue #668, opened 2022-12-10
Open Search in Docs with CTRL+K 


huntabyte:
### Describe what feature you'd like. Pseudo-code, mockups, or screenshots of similar solutions are encouraged!

This feature is more and more common across documentation sites.

Pressing `CTRL` + `K` while on any page within the documentation site would open the search console.

### What type of pull request would this be?

None

### Any links to similar examples or other references we should review?

https://tailwindcss.com/

endigo9740:
@huntabyte Funny enough it's technically already implemented:
https://github.com/skeletonlabs/skeleton/blob/dev/src/docs/DocsAppBar/DocsAppBar.svelte#L45

However, different browsers, and even different OS versions of browsers handle this differently. In Chrome on Mac it works for me, but in Chrome on my Windows PC it doesn't.

Currently the app is listening to window event keys for this shortcut combo. However, given this is a browser shortcut for browsers like Chrome (search in the URL bar) our method is not (able?) to take precedence. I'm not sure how Angolia does it with their search. Unfortunately I don't believe their frontend component is open source to be able to look at the source either, but I'd welcome help from anyone to figure this out.

I'm kind of leery of using `preventDefault` or `stopPropigation` at the window-level like this, which could potentially cause really detrimental behavior, but my guess is it's something related to that.

PR #669, opened 2022-12-10
Add 'ctrl+k' to focus docs search

huntabyte:
## Before submitting the PR:
- [X] Does your PR reference an issue?
- [X] Does your branch follow our [naming convention](https://www.skeleton.dev/docs/contributions)? 

## What does your PR address?
Allows users to use CTRL+K to search the docs.

Prevents default browser behavior of focusing the Address bar in certain browsers. CTRL/CMD+K can also be used to close the modal (something else common across other doc sites).

Fixes: #668

### DIFF
--- a/src/docs/DocsAppBar/DocsAppBar.svelte
+++ b/src/docs/DocsAppBar/DocsAppBar.svelte
@@ -41,12 +41,20 @@
 	// Keyboard Shortcut (⌘+K) to Focus Search
 	let pressedKeys: string[] = [];
 	function onWindowKeydown(e: any): void {
-		if ($modalStore.length) return;
-		if (e.code === 'MetaLeft' || e.code === 'KeyK') {
+		if (e.code === 'MetaLeft' || e.code === 'ControlLeft' || e.code === 'KeyK') {
+			// Prevent default browser behavior of focusing URL bar
+			e.preventDefault();
 			// Set pressed keys
 			pressedKeys = [...pressedKeys, e.code];
 			// If both keys pressed, focus input
-			if (pressedKeys.includes('MetaLeft') && pressedKeys.includes('KeyK')) search();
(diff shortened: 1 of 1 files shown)


endigo9740:
@huntabyte I'm going through my backlog of items in a linear fashion so I responded to your message before I noticed there was a PR. I see you've implement this using the `preventDefault` method. I'm going to hold this for testing for now until we can safely confirm this doesn't introduce any detrimental behavior. But it looks very promising at first glance! Thanks!

endigo9740:
@huntabyte I've had a chance to do a quick bit of testing on this. I see the issue was that control/command are different codes on Mac/Windows. I had no idea, good to know! FYI I've shared the preview URL on Discord to anyone on Linux confirm as well just in case.

That said, I wanted to confirm this bit in your code comments:

> // If modal currently open, close modal (allows to open/close search with CTRL/⌘+K)

Were you intending for the shortcut combo to both open AND close the modal? Because it opens but doesn't close. Totally fine as is, the ESC key works fine for now. But just wanted to confirm.

The only other note I could add would be purely a nitpick, which would be that we should expose the keyboard shortcut so users know this works. Totally cool if you want leave that to me, but any other edits are needed let's consider that as well please.

Assuming the Discord testing is good I'll aim to merge before our release Tuesday. Thanks!

huntabyte:
@endigo9740 On my end, if I CTRL+K -> then release -> CTRL+K again, it closes the search modal. I need to test across different browsers to be certain. This is the behavior that algolia provides, which is why I added it. Maybe someone else knows why the different behavior and how we can address it.

I will certainly add the keyboard shortcut to the search bar!

endigo9740:
Interestingly enough the open/close shortcut works on Chrome for Mac. So the issue is appears to be specific to Windows browsers.

Issue #650, opened 2022-12-07
Update Plugin to support Tailwind Intellisense


endigo9740:
### Current Behavior

I've heard reports it works for Daisy UI, and am sure it works locally in the Skeleton project, so we'll need to see how to carry this through to end users.

endigo9740:
Note that @AdrianGonz97 (`@CokaKoala` on Discord) has made some headway on this. I've asked him to share hit notes when het gets a chance.

AdrianGonz97:
I haven't had too much time to further look into it in the past couple of days besides getting some intellisense to work, but after playing around with tailwind's plugin system now, I _think_ I have a decent idea for what we'd have to do to get this working.

The gist of it is that we have to transpile all of our CSS into CSS-in-JS to then be able to add our classes as a tailwind plugin. Without adding them as a plugin, we won't receive any intellisense for them. This is the reason why we get intellisense for classes such as `bg-surface-900` but not for `btn-ghost`.

Since we split our stylesheets into different purposes, we have a few approaches to go with:

## Importing ALL

- We first have to combine all of our different CSS layers into their respective groups (ex. `@layer components`, `@layer base`, `@layer utilities`, etc...)
- Transpile each group of stylesheets with `postcss` into vanilla CSS
- Convert the vanilla CSS files into CSS-in-JS syntax as a commonjs exported object (using prejss-cli or postcss-js)
- Add those CSS-in-JS objects into the tailwind plugin via [tailwind's helper functions](https://tailwindcss.com/docs/plugins)(https://tailwindcss.com/docs/plugins) `addComponents` `addBase` `addUtilities`

After this, all we'd need the users to do is to import the `all.cjs` plugin in their `tailwind.config.cjs` file as so:

```ts
// ...
    plugins: [
        require('@skeletonlabs/skeleton/tailwind/theme.cjs'),
        require('@skeletonlabs/skeleton/tailwind/all.cjs')
    ]
```

No need for importing these stylesheets into the root layout file anymore.

## Selective imports

Currently, we have the option for users to selectively choose their stylesheets. To get this working, we could make each available stylesheet into their own plugin to maintain this behavior. For example, rather than doing this in the root layout:

```ts
import '@skeletonlabs/skeleton/styles/{stylehsheets}.css';	
```

We could instead have them imported as plugins in their `tailwind.config.cjs` file as so:

```ts
// ...
    plugins: [
        require('@skeletonlabs/skeleton/tailwind/theme.cjs'),
[...]

PR #673, opened 2022-12-11
Feat: Adding our CSS classes to Tailwind's intellisense

AdrianGonz97:
## Before submitting the PR:

## What does your PR address?
Progress tracking and experimenting for #650.

## Running
 - Be sure to `npm install` to add the new dependencies
 - Run `npm run build:jss` to transpile the `all.css` stylesheet
 - You'll need to reload the editor so that TW's extension can restart and detect the plugin


endigo9740:
@niktek Given how much I'm juggling this week I'm going to set you as a reviewer on this one. I'd still like to make the final call and review before merge, but you may be able to jump in sooner than myself.

endigo9740:
@AdrianGonz97 this PR has been sitting idle for a couple weeks now. I'm not sure if you have any interest in returning to research CSS-in-JS transpiring. If not we might want to opt to close this out for now and return later.

However, please see my proposal here. It's not a silver bullet, and only get us part of the way there, but may be something we should look into sooner than later:
https://github.com/skeletonlabs/skeleton/issues/650#issuecomment-1366127292

AdrianGonz97:
@endigo9740 I'm willing to give another deep dive into this again. I think I have fully recovered since my last CSS-in-JS venture 😂. Your suggestion in https://github.com/skeletonlabs/skeleton/issues/650#issuecomment-1366127292 seems like it should work! I'll give it a shot here.

endigo9740:
@AdrianGonz97 I know you're still working on this and testing things. But a quick review - from my perspective it looks like you're building a whole tool inside the Skeleton project to handle the transpiling process. I see quite a few additional commands, dependencies, scripts, etc. Something to consider here is is perhaps this would be better to split off to it's own package that Skeleton can then implement?

If built in the right way this might serve a dual purpose. First, it can be something you can own completely if you so choose, or we can fold it under the umbrella of Skeleton. I have no strong feelings either way.

But consider for a moment that if we're struggling with this issue, how many others might as well. It might be great if this is something that could be opened up and provided to others in a reusable fashion.

Let me know your thoughts!

AdrianGonz97:
@endigo9740 Most of the scripts I've added were part of my experimental phase, so I'll be deleting them once I get more dialed in on what's actually necessary. Same goes for the added packages. I'm not too sure that this would work as a separate package, but I can look into that later.

AdrianGonz97:
@endigo9740 This PR is _almost_ complete. 

One problem that I've been dealing with is the additional bundle size that this plugin adds on top of our already considerably large bundle. It was adding about ~30kB to the bundle by itself, which was unacceptable. 

I figured out nice way to negate that completely by only allowing the plugin to function during development. Now we can have that sweet intellisense without the penalty of increasing our bundle size!

One last thing to do is to add your suggestion https://github.com/skeletonlabs/skeleton/issues/650#issuecomment-1366127292. Might be a bit difficult to do because I'll have to convert to CSS-in-JS. We'll see how it goes!

AdrianGonz97:
@endigo9740 
Should be good to go now. Had to make a small change to prevent duplicate rules showing up in the intellisense because of #766, but that should be good now. We now have intellisense for all classes!

PR #690, opened 2022-12-14
Fix bug where menu wouldn't open on the first click after a page reload

### LINKED ISSUE #685: Menu won't open on first click
### Current Behavior

Menus don't open on the first click of a newly loaded page. If you click anywhere on the page and then click on the Menu, it'll work as expected. But, reloading the page and immediately clicking on a Menu will not open it.

### Steps To Reproduce

1. Go here: https://www.skeleton.dev/
2. Immediately click on "Features" in the top right, it won't open on first click.
3. To repeat, refresh the page and try to click on the menu.

### Anything else?

Introduced in the latest release `0.80.8`

bennymi:
…ad and remove node click event listener so that window click event listener takes care of click events

## Before submitting the PR:

## What does your PR address?

This would Close #685, where the menu wouldn't open on the first click after a page reload.

The problem was because `onToggleClick` was checking if the `elemMenu.style.display === 'none'` but this only gets set when the menu is closed, so when a click is made anywhere on the page but the menu button. I changed the if statement to fix this.

I also made another update and removed the node click event listener, since we already have a window click event listener, and having both of them was a bit confusing.

### Tips

### DIFF
--- a/src/lib/utilities/Menu/menu.ts
+++ b/src/lib/utilities/Menu/menu.ts
@@ -57,9 +57,9 @@ export function menu(node: HTMLElement, args: ArgsMenu) {
 	
 	// Click Handlers ---
 
-	const onTriggerClick = (): void => {
+	const toggleMenu = (): void => {
 		// When the node is clicked, open the menu if it is closed, otherwise close it
-		if (elemMenu.style.display === 'none') {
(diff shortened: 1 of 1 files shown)


endigo9740:
This is great Benny, this feels much better to use in practice. Thank you for this!

PR #691, opened 2022-12-15
Update package.json and fix css import

niktek:
## What does your PR address?
Update to to all the new 1.0 packages and Vite 4.0.1 so that we can stress test the new libs during development.

With Vite 4 any named imports such as `import skeleton from '$lib/themes/theme-skeleton.css?inline';` now need to have ?inline appended.  **DO NOT** append ?inline for unnamed imports like `import '../app.postcss';` as this will break the imports.


endigo9740:
@niktek again?

https://github.com/skeletonlabs/skeleton/pull/689

niktek:
god damnit, I didn't sync my fork before creating this one

endigo9740:
Someone needs more coffee!

PR #696, opened 2022-12-15
Expand toasts with preset feature

### LINKED ISSUE #682: Toasts: Add canned preset options like success, error, etc

Suggested by @huntabyte. Might make it easier to use rather than having to generate these on the fly or generate all your preset each project.

Feature

https://react-hot-toast.com/

endigo9740:
Fixes #682

I'm submitting this for review and feedback. The ticket details the options available. Let me know if we should add/tweak/adjust.

### DIFF
--- a/src/lib/utilities/Toast/stores.ts
+++ b/src/lib/utilities/Toast/stores.ts
@@ -28,8 +28,22 @@ function toastService() {
 		trigger: (toast: ToastSettings) =>
 			update((tStore) => {
 				const id: string = randomUUID();
+				// Apply Preset Color Styles
+				let classes = toast.classes ?? '';
+				// prettier-ignore
+				switch (toast.preset) {
+					// Theme
+					case('primary'): classes += '!bg-primary-500 text-on-primary-token'; break;
+					case('accent'): classes += '!bg-accent-500 text-on-accent-token'; break;
+					case('tertiary'): classes += '!bg-tertiary-500 text-on-tertiary-token'; break;
+					case('warning'): classes += '!bg-warning-500 text-on-warning-token'; break;
+					// Success/Error
+					case('success'): classes += '!bg-green-500 !text-black'; break;
+					case('error'): classes += '!bg-red-500 !text-white'; break;
+				}
 				// Merge into store
-				const tMerged = { ...toastDefaults, ...toast, id };
+				const tMerged = { ...toastDefaults, ...toast, id, classes };
+				console.log(tMerged);
 				tStore.push(tMerged);
 				// Handle auto-hide, if needed
 				handleAutoHide(tMerged);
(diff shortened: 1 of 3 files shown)


src/routes/(inner)/utilities/toasts/+page.svelte
```diff
@@ -54,10 +54,17 @@
 		toastStore.trigger({ message: 'Message will auto-hide after 5 seconds.' });
 	}
 
+	function toastPreset(preset: 'primary' | 'accent' | 'tertiary' | 'warning' | 'success' | 'error'): void {
```
AdrianGonz97, line 57:
```suggestion
	function toastPreset(preset: ToastSettings['preset']): void {
```
This'll just nicely tie the type of the preset to the type of the parameter

src/lib/utilities/Toast/stores.ts
```diff
@@ -28,8 +28,22 @@ function toastService() {
...
+				switch (toast.preset) {
+					// Theme
+					case('primary'): classes += '!bg-primary-500 text-on-primary-token'; break;
+					case('accent'): classes += '!bg-accent-500 text-on-accent-token'; break;
+					case('tertiary'): classes += '!bg-tertiary-500 text-on-tertiary-token'; break;
+					case('warning'): classes += '!bg-warning-500 text-on-warning-token'; break;
+					// Success/Error
+					case('success'): classes += '!bg-green-500 !text-black'; break;
+					case('error'): classes += '!bg-red-500 !text-white'; break;
+				}
```
AdrianGonz97, line 43: This could be simplified down to this:
```suggestion
				switch (toast.preset) {
					// Success/Error
					case('success'): classes += '!bg-green-500 !text-black'; break;
					case('error'): classes += '!bg-red-500 !text-white'; break;
					// Theme
					default: classes += `!bg-${toast.preset}-500 text-on-${toast.preset}-token`; break;
				}
```
Though it may not be as explicit as you'd like

AdrianGonz97 (commented):
Just 2 super tiny things. The rest looks good!

src/lib/utilities/Toast/stores.ts
AdrianGonz97: Huh, the tabs look wild in this suggested changes preview

src/routes/(inner)/utilities/toasts/+page.svelte
endigo9740: I had no idea this was possible, very useful. Thanks!

src/lib/utilities/Toast/stores.ts
endigo9740: One of those "oh duh" moments. Implementing this too. Thanks!

endigo9740:
@AdrianGonz97 I've implemented your suggestions, thanks!

PR #703, opened 2022-12-16
Fix `DataTable` sorting with BigInts

tqwewe:
## Before submitting the PR:

## What does your PR address?

The `DataTable` sorting is broken when the data contains `BigInt` values. This is because `Array.sort` doesn't support returning `BigInt`'s, and errors with:
> Cannot convert a BigInt value to a number

A simple fix for this is to return -1, 0, or 1 from the sort function, which is what this PR implements.

<https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt#comparisons>

Note I didn't run any tests for this and simply made the change in Github.


endigo9740:
@AdrianGonz97 Do you mind giving this a review?

AdrianGonz97:
LGTM and works as expected. 

While reviewing this though, I did spot a bug where search doesn't work while a sort is active (when you see the arrow pointing up or down in the table headers). I'll add that to the list of things to fix, but that's unrelated to what's been changed here.

PR #733, opened 2022-12-19
Theme System Overhaul

### LINKED ISSUE #598: Theme System Overhaul
This update will cover a number of updates and expansions to the theming system, including:

- [x] Finalize the available number of color slots
- [x] Improve/rewrite/replace the hex color generation script to account for better contrast
- [x] Redesign and refactor the theme generator UI
- [x] Update theme picker and preview features on the documentation site

## Per Color Slots

Let's look to extend the color palette with success and error colors. There's numerous use case for these.

```
- Primary
- Accent
- Tertiary
- Success <-- add this
- Warning
- Error <-- add this
- Surface
```

Convention: success will typically be green, warning will be yellow, while error will be red. These can be used in everything from toasts, to inputs, alerts, and more.

## Per the Hex Color Library

We currently depend on the following package for this feature:
https://github.com/bobthered/tailwindcss-palette-generator

However, the dark colors generated tend to have too much contrast (be too dark) compared to the Tailwind color palettes. The library has recently undergone an update and the author seems responsive to our communication. So we should either consider a direct contribution to this project or consider rewriting this script to work as we need it.

General idea is to add luminosity and other configurable options long term. Not unlike this:
https://tailwind.simeongriggs.dev/blue/3B82F6

## Redesign the Theme Generator UI

Go through a full design process to wireframe, mock, and implement the new UI. Then refactor the UI simplify and make this easier to manage. This should represent as close to possible the v1.0 launch state.

## Update Theme Picker and Preview in Docs

I feel like this will come naturally given the other updates occurring. But primary focus should be reducing the "flash" on page load if possible.

endigo9740:
Fixes #598


src/routes/+layout.svelte
```diff
@@ -25,30 +25,23 @@
...
-	import seasonal from '$lib/themes/theme-seasonal.css?inline';
-
 	// Default Theme, injected immediately:
 	import skeleton from '$lib/themes/theme-skeleton.css?inline';
+
+	// Dynamically load selected theme
+	const themesGlob = import.meta.glob('$lib/themes/*.css', { as: 'raw' });
+	async function getTheme(theme: string): Promise<string> {
+		return (await themesGlob[`/src/lib/themes/theme-${theme}.css`]()) as string;
+	}
```
AdrianGonz97, line 35: slight reduction, heh
```suggestion
	function getTheme(theme: string): Promise<string> {
		return themesGlob[`/src/lib/themes/theme-${theme}.css`]();
	}
```

PR #751, opened 2022-12-29
Feature: Table of contents

### LINKED ISSUE #357: New Component: Table of Contents

We need a new table of contents. We'll immediately put it to use in the Skeleton docs site, but seems like a great addition to the library itself too. This will allow quick navigation on long pages with a lot of content.

Here's a mock of where it would live on the page:

Screen Shot 2022-10-12 at 12 19 04 PM

Here's a full version of the page to compare:

image

Suggested features:

- I love how Mantine's version updates automatically as you scroll up/down the page
- Support subheadings, which are indented
- Perhaps we implement `data-*` attributes on the headings to make the pairing process easier

Beyond that, this is open ended. Additional features and such as welcome!

New Feature

References (see the right side of the page)

- https://mantine.dev/core/button/
- https://mui.com/material-ui/react-button/

endigo9740:
Fixes #357

### DIFF
--- a/src/docs/DocsNavigation/links.ts
+++ b/src/docs/DocsNavigation/links.ts
@@ -97,7 +97,8 @@ export const menuNavLinks: any = [
 			{ href: '/components/steppers', label: 'Steppers' },
 			// { href: '/components/svg-icons', label: 'SVG Icons' }, // Keep Disabled
 			{ href: '/components/tabs', label: 'Tabs' },
-			{ href: '/components/tables', label: 'Tables' }
+			{ href: '/components/tables', label: 'Tables' },
+			{ href: '/components/table-of-contents', label: 'Table of Contents' }
 		]
 	},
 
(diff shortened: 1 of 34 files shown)


endigo9740:
I've bumped this to "ready for review" status

bennymi:
@endigo9740 looks great! 

I think one thing we could add for the documentation is to make it sticky so it's always visible even when scrolling down and perhaps adding some kind of styling to active/visible elements. Example: https://css-tricks.com/sticky-table-of-contents-with-scrolling-active-states/

endigo9740:
@bennymi That would be built into the page layout, not really the component. Currently that's not easily implemented in our current page layout.

But I have a design refresh in mind for the doc site prior to the v1.0. It can include improvements like that, plus sticky tabs as well for Usage, Props, etc.

bennymi:
@endigo9740 yeah I meant for the documentation as well. The sticky table of contents would be something the user would have to implement on their own, or we could add a section to the table of contents documentation describing how to make it sticky with the AppShell component.

PR #752, opened 2022-12-30
Chore/stepper clickable

### LINKED ISSUE #408: Option to make Steps in Stepper clickable

It would be nice if I could navigate through Steps in the Stepper component by clicking on their name/number. You could introduce a new setting to activate this behaviour with the options: enabled, disabled and unlocked.

unlocked (as the name suggests) would only allow to click on Steps that have already been completed or is currently active.

_No response_

_No response_

bennymi:
## Before submitting the PR:
- [X] Did you update documentation related to your new feature or changes?

## What does your PR address?

This PR would Close #408.

- Added an extra prop navigateOnClick with three different options. When active, it allows the user to click on the header or step numbers to open the step.
- Added a rounded prop, so the numeral shape can be changed.
- Added a step to the documentation to explain the new navigateOnClick prop and changed the headers for two steps to be more descriptive.

### Tips

### DIFF
--- a/src/lib/components/Stepper/Step.svelte
+++ b/src/lib/components/Stepper/Step.svelte
@@ -31,6 +31,7 @@
 	export let rounded: string = getContext('rounded');
 	export let duration: number = getContext('duration');
 	export let navigateOnClick: string = getContext('navigateOnClick');
+	export let highestStepReached: Writable<number> = getContext('highestStepReached');
 	// Context (overrides)
 	export let color: string = getContext('color');
 	export let background: string = getContext('background');
@@ -63,7 +64,8 @@
 	// Reactive
 	$: isLastItem = index === length - 1;
 	// clickable & cursor logic
-	$: isClickable = navigateOnClick === 'enabled' || (navigateOnClick === 'unlocked' && !locked);
+	// $: isClickable = navigateOnClick === 'enabled' || (navigateOnClick === 'unlocked' && !locked);
+	$: isClickable = navigateOnClick === 'unlocked' && index <= $highestStepReached;
 	$: classCursor = isClickable ? 'cursor-pointer' : 'cursor-default';
 	$: btnTabindex = isClickable ? 0 : -1;
 	// Base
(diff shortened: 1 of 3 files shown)


endigo9740:
@bennymi thanks for this! I do have two quick bits of feedback

1. I think we should default to the `rounded-token` class to auto-adapt the numeral values border radius to the theme settings. One of the major selling points is Skeleton auto-adapts to your theme by default
2. I was under the impression that we would only be navigating to numerals that have been viewed and thus unlocked.

Per number 2 - being able to jump from step 1 to step 6 with a locked step between seems wrong. The point of steppers is linear progression from start to finish, otherwise it's just an accordion, and we already have a component for that which can accept numerals in the `lead` slot.

Primarily click to navigate just being a quick way to go back to prior completed steps. What do you think?

bennymi:
@endigo9740 hey, yeah that makes sense. I'll update the the default value to `rounded-token`.  

About your 2nd point, there might be a scenario where someone might still want to allow people to look ahead to see what  needs to be done for each step, even if a step is unlocked. However, with the current implementation this might not really make sense, since steps don't have their own `complete` status, but rather a step is seen as completed if it's before the currently selected one. Changing this behaviour would require changing the Step API as well, so I can also just remove the `enabled` `navigateOnClick` option for now. What do you think?

endigo9740:
@bennymi With step progression, we always have the shared state between the parent for the current step. As long as we keep a record for what the "highest" step index reached was, then that is our barrier. If any step index is below that threshold then it's clickable.

Given `active` is a writable, that means we can listen to it in the parent component:

```ts
let highestStepReached: Writable<number> = writable(0)

active.subscribe(v => {
    if (v > $highestStepReached) highestStepReached.set(v);
});
```

Then pass the "highest step" value down to the children using the context API:

```ts
setContext('highestStepReached', highestStepReached);
```

Then we can compare the local step index value compared to this and know where or not it should be allowed to be clickable.

FYI all the above code is untested, just a concept. I'd also welcome a better name than `highestStepReached` as it's a bit of mouthful heh

bennymi:
@endigo9740 I personally like long and understandable variable names so I just went with it 😄 I updated the code as well, changed the default to the token, changed the `navigateOnClick` behaviour, and updated the documentation.

endigo9740:
@bennymi I've made some final nitpick-y adjustments to prop names, example settings, etc. I think with this we're good to merge. Doing so now!

Issue #377, opened 2022-10-15
New Component: Autocomplete


endigo9740:
![ref](https://uploads.sitepoint.com/wp-content/uploads/2021/01/1611572201datalist-country.gif)

### Describe what feature you'd like. Pseudo-code, mockups, or screenshots of similar solutions are encouraged!

I'm marking this as a "feature" because I don't know if it should be a component or action yet. Perhaps a combination of the two, then treated as a utility?

General idea is when typing in an input you should see a list of suggested values. The HTML5 DataList can handle this, but I'm not sure about browse support and visual styling. Please review and discuss the capabilities with other contributors before implementing. Functionality matters first and foremost though.

If at all possible we should avoid reinventing the wheel here and creating this custom. However, if that is our only practical solution we can consider it. But again, let's review our options before a final decision is made!

### What type of pull request would this be?

New Feature

### Any links to similar examples or other references we should review?

Datalist Guide:
https://www.sitepoint.com/html5-datalist-autocomplete/

MDN reference:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist

Mantine reference:
https://mantine.dev/core/autocomplete/

endigo9740:
After some research and testing, auto-complete is available in Chrome/Safari/Firefox, however they not really intended to be styled. Each browser has its own UI and styling. Unfortunately, if we wish to use this for our search modal we'll need to implement our own version of this.

bato3:
I would like to propose [attractions](https://github.com/illright/attractions/tree/main/attractions/autocomplete) as an reference.

Example: https://illright.github.io/attractions//docs/components/autocomplete

endigo9740:
Awesome, thank you @bato3!

xmlking:
https://mskocik.github.io/svelecte/
I wish this component can be customize with tailwind css, that is why I want this component offered in skeleton

robotcorner:
> https://mskocik.github.io/svelecte/

This would be an awesome place to start. Talk about doing one thing seemingly really well.

huntabyte:
@endigo9740 - I will take a stab at this one.

endigo9740:
@huntabyte For sure go for it, the only thing I might suggest is seeing if we use this in some fashion for our doc search. The next few components I have slated to work on were all selected because they have place in the docs.

I'm 100% ok with the design changing (function > form here), but just a heads up of the intention!

endigo9740:
[...]

PR #758, opened 2023-01-03
Add Autocomplete component

huntabyte:
## Before submitting the PR:

## What does your PR address?

This is only the initial groundwork for the Autocomplete component. This component is quite the beast when you think about the different ways a developer may want to use it, which we'd ideally like to support and prevent as many destructive errors from occurring on our side as we can.

At this moment, there are many things I will probably change once I have some more time to dive a bit deeper and I greatly welcome any feedback/refactoring tips, as most of this work was done around 2 am.

Some of the things that it supports at this moment (each of which will require further testing of course):
* Supports passing an asynchronous function to retrieve the data from an external source, such as an API
  * Covers both generator functions and regular functions
  * Debounce using user-defined props `delay` and `minCharactersToSearch` 
* Supports passing either an array of objects or an array of values as `items`. The component logic is able to determine which it is working with and function accordingly
  * If an array of objects is passed, the user can also set which properties will be used as the `label` and `value`
  * Users can also provide their own custom functions to construct the `label`, `value`, and `keywords`. `keywords` are used for filtering based on `x` different properties
  * Some baseline protection was added to deal with unpredictable data (especially if coming from an API), but more eyes/testing is required to say its complete.
* Hidden select input is populated with the selected values to enable this component to be submitted within a standard form.
  * `selectedItemValues` and `selectedItems` are provided as props to enable the user to `bind:localVar={selectedItemValues}`, etc.
  * Props to set select attributes are also provided `selectName`, `selectId`, `multiple`.
* Multiple selections using some logic from the `InputChips` component.
  * If `multiple`, users can pass a prop to `allowDuplicates` which informs the component to allow duplicate value selections, otherwise, they are ignored.

Some of the things I haven't had the time to handle just yet:
* Styles - at the moment I just slapped some random styles that looked decent enough for an initial concept.
* Docs - this of course will be pieced together once the component is complete. Right now there is just a test page to show the component to the maintainers
* Tests - this will also come once things are cleaned up a bit and we have an agreed way forward.

Things I know are currently flawed that I will begin to work on this week:
- [ ] Dropdown navigation can break if certain key combinations are pressed. Need to clean up the keymaps and isolate the issue there.
- [ ] Dropdown should probably show *something* (if not asynchronous of course) if the user doesn't match any results. Will probably add this as props with a `boolean` to show it and then some custom `noResultsText` or something like that.

Fixes: #377


endigo9740:
Thanks @huntabyte, I'll aim to do a deeper review tomorrow. Only quickly bit of feedback is make sure to link the component in the site nav so it's easier to find. You can add this quickly using the `links.ts` file.

huntabyte:
> Thanks @huntabyte, I'll aim to do a deeper review asap. Only quickly bit of feedback is make sure to link the component in the site nav so it's easier to find. You can add this quickly using the `links.ts` file.

Added to Nav ✅

@endigo9740 - I ran into a stacking context issue which will be a problem when users want to render out forms with an Autocomplete in them. Currently, a `filter` is applied on hover of the inputs across the application, which creates a new stacking context causing any inputs rendered after the autocomplete within the DOM order will overlap the dropdown on hover. At the moment, I'm unsure of a possible workaround for this at the moment, so I've temporarily set the autocomplete input to `filter: none`.

endigo9740:
@huntabyte sounds good. Thanks for keeping me updated. I'm going to try and squeeze in a review tomorrow, as time allows.

endigo9740:
Hey @huntabyte, I've finally had a chance to sit down and review this. With these reviews I try to emphasize what's working well, as be critical to the items that need improvement. So I hope you'll be receptive to my criticisms here.

First off, the component does appear to be doing what users expect - when tapped you see a popup with a list of suggestions. That's great. The styling is good and meshes with the themes. I like that you've created multiple examples working with and without and chip concept. I also want to say great job on the documenting inline in the component. A lot of folks miss the JSDocs-style comments for props, etc. Which of course we need for our documentation to auto-generate! So great job on all those fronts. I also loved that you've already built in keyboard interaction right from the start. A11y is very important to me, so thank you for this!

That said, there's a number of issues here. And I'm not going to live, some of which are foundational. And I want to make it clear I take responsibility for not outlining a better set of requirements upfront or reviewing this sooner. But that's why I always encourage folks to get a draft PR up ASAP. In hindsight I think a lot of these could have been avoided by doing that. I'll keep this in mind going forward, but for now I'll break down each issue below in detail.

---

## Intermingled Input Chip

I see you've opted to ingest and intermingle the input chip. Basically recreating those features within this component. While that works, it does pose some maintainability issues. Essentially you've recreated the wheel here. If we make a change to the Input Chips component, those won't necessarily carry over automatically. Which obviously poses some issues. It's the oppose of "D.R.Y." (don't repeat yourself). One of the first changes I suggest is ripping that out of the component.

## Keep it a Primitive

What we really need is this component to act as a primitive - meaning it's a building block, like Legos. You should be able to mix and match many different components to produce a different result. In this case, auto-complete should be able to be paired with any standard input (text, search, date, etc),  the existing Input Chips, or really anything that can take a value. To do this end, we need it be razor focused on doing only a few things:

- Take in a set of data - most like an array of objects to populate a suggestion list (aka selection list)
- Be able to map or define the label for each list item
- Be able to capture the selection on click - via either bind, and event, or ideally both

## Presentation

I'm thinking ideally the component might act as a "wrapper" around another component or input. Here's a quick visualization of how we might handle that:

Screen Shot 2023-01-04 at 12 16 52 PM

The resulting code might look something like this:

```html
<AutoComplete list={[arrayOfObjectsHere]}>
    <input type="test" bind:value />
</AutoComplete>
```
It would then work like this:

- The input is inserted into the default slot position
- The surrounding AutoComplete component would listen for clicks to the first child element (the input)
- When clicked, a suggestion list would be presented to the users - via dropdown menu or w/e

## Abstracting Inputs

Abstracting inputs into components is generally a bad idea. We do it in a few places in Skeleton, but more in a begrudging fashion. Where the input itself is not exposed , but instead controls state (ex: switch toggles). The biggest issue is that abstracts so much away that it makes it hard for devs to interface with the input directly. We have to provide this tunnel to/from the input, rather than just providing direct access. I've seen other libraries like Angular Material do this - and trust me it's a pain in the rear!

## Handling Selection

When a selection is made in the suggestion list, there's a few ways we could update the state of the input:

1. We could perhaps duplicate the binding to the AutoComplete component, so when an option is selected the value just updates immediately. Though I'd need to check this flow of component -> parent -> input.
2. We could use an event handler. So when an option is selected, dispatch and event, then listen for that on the component in the parent page and update the value that way.
3. If we're already listening to click events on the child input in the slot - then we have a direct way to interface it already. We just modify the value directly.

Number 1 and 2 seem very roundabout, so I'd most likely lean into number 3. The logic would be as simple as:

`elemInputReference.value = newValue`

## Style Issues

There's a number of style issues here. The biggest flaw is that you're stripping the styling to the native inputs and recreating them in an wrapping parent element. Again, we should lean into what is working already rather than recreate the wheel here. Keep in mind it is possible to extend `forms.css` if they do not cover all use cases currently.

Additionally I see the use of many styles in a `<style>` block at the bottom of the component. We almost never use these because there's almost always a better way to handle this. But you mentioned you needed help with styles so don't fret on this. We'll circle back on that.

---

I think for now I'll stop here. Note that I can't truly review the code itself in many places given that we might need to make some sweeping changes here. So anything input I provide might not be relevant until that's done.

But wrapping up I want to say I know you've put a ton of work into this and I hope I don't come across as discounting any of that. Definitely not my intention. I do think you've got a lot of interesting stuff going on here, so I might suggest copying what you've done and saving it locally. You never know when some of these techniques might come in handy later. But for Skeleton, we need this component to be much simpler, and much more generally applicable if that makes sense. 

I'll be on Discord through the afternoon so please do feel free to reach out if you would like me to expand or clarify anything here. I'd be happy to do so. I could even jump on a Zoom call or something if need be. Again, my role is to help anywhere I can, so please let me know if I can do so!

Thanks for the hard work Hunter!

endigo9740:
@huntabyte you mentioned restarting, so let me know if you would like me to close out this PR. I'm fine with segmenting the next pass as it's own PR. You're welcome to close this out directly as well. It'll remain in the closed PR section if we need to reference anything here.

huntabyte:
Closing to restart with a fresh slate ✅

PR #771, opened 2023-01-05
Additional Tailwind Elements Added

### LINKED ISSUE #756: Convert an additional set Svelte Components to Element styles

None of these components contain logic, which makes them prime candidates to convert to Elements. That being CSS styles like buttons, badges, cards, etc rather than Svelte components. This greatly simplified the easy of use (no import needed).

- [x] Alerts
- [x] Breadcrumbs
- [x] Dividers
- [x] Gradient Headings

The first three would be a similar conversion to buttons, badges, cards, etc. But Gradient Headings would be a bit different. In most cases we would recommend folks generate their own `@apply` styles that are dropped into a reusable class in their global stylesheet. Something like this:

image

For now I'm opening this to feedback and questions, but plan to jump on this in the next few days.

Enhancement

_No response_

endigo9740:
Fixes #756

### DIFF
--- a/src/lib/styles/elements/breadcrumbs.css
+++ b/src/lib/styles/elements/breadcrumbs.css
@@ -1,0 +1,13 @@
+/* Tailwind Elements: breadcrumbs.css */
+
+@layer components {
+	.breadcrumbs {
+		@apply flex items-center space-x-4 w-full hide-scrollbar overflow-x-auto;
+	}
+	.crumb {
+		@apply flex justify-center items-center space-x-2;
+	}
+	.crumb-separator {
+		@apply flex text-surface-700-200-token w-2 opacity-50;
+	}
+}
--- a/src/lib/styles/elements/breadcrumbs.css
+++ b/src/lib/styles/elements/breadcrumbs.css
@@ -1,13 +1,27 @@
 /* Tailwind Elements: breadcrumbs.css */
 
 @layer components {
-	.breadcrumbs {
+	.breadcrumb-nonresponsive {
 		@apply flex items-center space-x-4 w-full hide-scrollbar overflow-x-auto;
+	}
+	.breadcrumb {
+		@apply breadcrumb-nonresponsive;
 	}
 	.crumb {
 		@apply flex justify-center items-center space-x-2;
 	}
 	.crumb-separator {
-		@apply flex text-surface-700-200-token w-2 opacity-50;
+		@apply flex text-surface-700-200-token opacity-50;
+	}
+
+	/* === Auto-Responsive === */
+
+	.breadcrumb li {
+		@apply hidden md:block;
+	}
+	.breadcrumb li:nth-last-child(3),
+	.breadcrumb li:nth-last-child(2),
+	.breadcrumb li:nth-last-child(1) {
+		@apply block;
 	}
 }
--- a/src/routes/(inner)/elements/breadcrumbs/+page.svelte
+++ b/src/routes/(inner)/elements/breadcrumbs/+page.svelte
@@ -40,7 +40,7 @@
 				<ol class="breadcrumbs">
 					<li class="crumb"><a href="/">Home</a></li>
 					<li class="crumb-separator">/</li>
-					<li class="crumb"><a href="/">About</a></li>
+					<li class="crumb"><a href="/">Articles</a></li>
 					<li class="crumb-separator">/</li>
 					<li>Current</li>
 				</ol>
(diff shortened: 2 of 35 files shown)


endigo9740:
@ryceg

src/lib/styles/elements/breadcrumbs.css
```diff
@@ -0,0 +1,13 @@
+/* Tailwind Elements: breadcrumbs.css */
+
+@layer components {
+	.breadcrumbs {
+		@apply flex items-center space-x-4 w-full hide-scrollbar overflow-x-auto;
```
ryceg, line 5: Looks a bit unfinished on small screen sizes. A method I've used to great effect before is hiding all but the previous link on small screens? I might also suggest a shadow to show there's content being hidden- see https://css-tricks.com/books/greatest-css-tricks/scroll-shadows/ 

image

src/lib/styles/elements/breadcrumbs.css
```diff
@@ -0,0 +1,13 @@
...
+	}
+	.crumb {
+		@apply flex justify-center items-center space-x-2;
+	}
+	.crumb-separator {
+		@apply flex text-surface-700-200-token w-2 opacity-50;
```
ryceg, line 11: small sanity check, should it be max-w or min-w? Likely not since this is pretty static but I figure it's worth double checking

src/routes/(inner)/elements/breadcrumbs/+page.svelte
```diff
@@ -0,0 +1,123 @@
...
+		<section class="space-y-4">
+			<h2>Separators</h2>
+			<!-- prettier-ignore -->
+			<p>Apply the <code>.crumb-separator</code> class to each list item containing your separator icon or <a href="https://unicode-table.com/en/sets/arrow-symbols/#right-arrows" target="_blank" rel="noreferrer">unicode symbol</a>.</p>
+			<CodeBlock language="html" code={`<li class="crumb-separator">&rsaquo;</li>`} />
+			<CodeBlock language="html" code={`<li class="crumb-separator">(icon)</li>`} />
+		</section>
+		<section class="space-y-4">
+			<h2>Current Page</h2>
+			<p>We recommend the last item not include a anchor link, as this indicates the page the user is currently on.</p>
```
ryceg, line 119: an anchor

src/lib/styles/elements/breadcrumbs.css
```diff
@@ -0,0 +1,13 @@
+/* Tailwind Elements: breadcrumbs.css */
+
+@layer components {
+	.breadcrumbs {
+		@apply flex items-center space-x-4 w-full hide-scrollbar overflow-x-auto;
+	}
+	.crumb {
```
ryceg, line 7: Oh yeah, this should have a minimum width of `48px` to ensure that it is tappable. If you want to restrict this to just touch screen devices, you can use the `@media (pointer:coarse)` query.

src/lib/styles/elements/breadcrumbs.css
```diff
@@ -0,0 +1,13 @@
...
+		@apply flex justify-center items-center space-x-2;
+	}
+	.crumb-separator {
```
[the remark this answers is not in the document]
endigo9740: Likewise, but in order to support third party icon libraries folks need to be able to define the symbol using HTML markup. If we did it purely through CSS this would not be possible.

I've made the change to the docs and our own implementation of the crumb elements. Good call!

src/lib/styles/elements/breadcrumbs.css
endigo9740: I like your idea about shadows, but unfortunately this is not something that can be added with pure CSS. Even the CSS Tricks examples use JS to check scroll position.

Love the idea about limiting to the last two options on mobile, so I've implement exactly that:

`.breadcrumb-non-responsive` - shows all options, works as before
`.breadcrumb` - auto-responsive by default

It works really well in execution!

Oh and I dropped the "s" in `.breadcrumbs` in case you're curious. We use singular names for all other elements.

src/lib/styles/elements/breadcrumbs.css
endigo9740: Most of these styles are remnant from the component. It shouldn't have a width at all! Removed!

src/lib/styles/elements/breadcrumbs.css
endigo9740: GitHub doesn't specify which line you're referring to specifically, so I'm going to assume you meant `.crumb`. Unfortunately not all crumb elements may be clickable. It's up to the user to implement a button or link. Not sure that us setting a min-width on a `li` would be helpful here.

endigo9740:
QA changes have been applied, merging into dev branch, but still welcome feedback of course!

PR #834, opened 2023-01-16
Bugfix/Expose previously unexposed component props that are being used

### LINKED ISSUE #800: typescript errors when assigning `class` attribute to LightSwitch
### Current Behavior

[LightSwitch seems to support adding custom classes](https://github.com/skeletonlabs/skeleton/blob/dev/src/lib/utilities/LightSwitch/LightSwitch.svelte#L81)

```typescript
$: classesBase = `${cTrack} ${$$props.class ?? ''}`;
```

### Steps To Reproduce

```svelte
  <LightSwitch class="top-10 right-10 absolute hover:ring-secondary-500-400-token" />
```

The above works, however, since the props don't declare the `class` prop it leads to a type error

```
Error: Type 'string' is not assignable to type 'never'. (ts)
```

### Anything else?

_No response_

AgarwalPragy:
## Before submitting the PR:

## What does your PR address?

[Fixes #800](https://github.com/skeletonlabs/skeleton/issues/800)

Exposes previously unexposed component props that are being used


AgarwalPragy:
@endigo9740 

- Failed test `src/lib/components/SvgIcon/SvgIcon.test.ts > SggIcon.svelte > Renders with all props`. Test was failing in the dev branch that I forked from. This failure is not caused by the current PR.
- `npm run lint` and `npm run format` ended up changing several files that I did not intend to change. Basically existing code that did not follow the styleguide.
- I haven't updated the docs to include the newly exposed props. Let me know if that is required.

endigo9740:
@AgarwalPragy I'm sure you mean well here, but we have a pretty hard requirement for upfront discussion before changes are made. Much less sweeping changes like this. Please see our contribution guidelines here:

Items 1-3 state this very plainly.

So far we've not been able to reproduce the issue you've raised. You were asked provided a minimal reproduction, but instead provided steps to reproduce. So given this it won't be until tomorrow before I can this. Then, and only after then, will we begin fielding solutions or changes.

Please keep this in mind when submitting additional tickets or PRs.

endigo9740:
After reviewing a bit further - there's changes to `disabled` props and the `app.html` formatting which are way outside of scope of the original issue. I'm going to close this PR.

Going forward the process should be this:

1. Report an issue - include detailed steps to product, provide a reproduction repo if possible (or asked)
2. We'll test and try to reproduce, if we can't we'll ask for more info, if we can we'll discuss a fix
3. Then, and only then, should a PR be made that addresses that issue in the agreed upon manner

Thanks for your understanding.

AgarwalPragy:
@endigo9740 My bad. I've created a minimal reproduction here: https://github.com/AgarwalPragy/exposed-props

I agree, it would've been better had I discussed before creating the PR.

Nevertheless, I think that the changes to `disabled` props should be part of this PR, as this PR focusses on explicitly exposing any used props.

The changes in `app.html` was because of `npm run format` and `npm run lint` that are required according to your contribution guidelines :)

PR #842, opened 2023-01-16
Added CSS type to props

niktek:
Only merge this if it helps, totally fine if the typings come from the polishing process.


endigo9740:
The problem right now is I don't know what or how many props are going to change. It'll occur on a per-component basis. I recommend keeping this as a draft for reference, but 110% needs to happen during or after the refinement passes

endigo9740:
Closing this as the new PR has replaced this.

PR #847, opened 2023-01-18
Feat/contrast checker

ryceg:
## Before submitting the PR:

## What does your PR address?

Adds a contrast checker that allows the user to quickly make sure that their theme meets WCAG AA or AAA guidelines. Tooltip allows them to hover for more info (but is currently not updating when you change color)


endigo9740:
Quick feedback:

1. Using the current them colors for success/warning/error falls apart for randomization. I'd recommend using hardcoded green/yellow/red instead to indicate status. They may not not play well with the theme, but they're not part of the theme.
2. Not sure I understand the text crammed into the label text. It's breaking the layout. I'd suggest keeping this info in your tooltip

image

3. I should have been more specific - I meant native `tooltip` attributes not our utilities. This is a temporary home as I feel the tooltips may undergo some structural changes and should not be relied on atm. If you absolutely keep theme ensure they use a neutral color - even `bg-neutral-900` would be fine here.
4. There's little reason to use colored emoji when we have a proper icon library inserted now. We can use any that fall under "free". And with the proper on-color these should remain readable. Right now we're getting the "clown" effect where everything is so colorful it's grabbing all the attention. This is a tool, not the main attraction.

- https://fontawesome.com/icons/thumbs-up?s=solid&f=classic
- https://fontawesome.com/icons/thumbs-down?s=solid&f=classic

4. The use of buttons is odd odd. I might suggest badges, which can be much smaller by default.
5. Nik brought this up, but I also didn't understand the check mark. Again positive/negative should be the focus, with more info for power users in the tooltip. Anything more than that and we'll have to introduce a legend for normal users.

Ping me when these changes are in and I'll review again. Thanks!

endigo9740:
@ryceg I know this is still a work in progress, but a few small issues:

- Position is off. This needs to be bumped to the left a bit to overlap. Test with/without rounded inputs too.
- The badge is oblong (taller than it is wide). Not sure why.
- When hovering there needs to be a space after the RGB values

Screen Shot 2023-01-19 at 1 39 22 PM

endigo9740:
@ryceg First off great job with this, it seems like a very useful feature that can help a lot of folks.

That said there's a few things we might need to address here:

## Icons

The face icons just don't work on my monitor - there is too much detail crammed into too small of a space. My utr widescreen monitor cannot run native high pixel density, so the result is a fuzzy mess. Given the I'm going to suggest we lean into your other idea of using filled stars. It's not perfect, but it's better.

I've tried to capture this with my phone (sorry for the poor quality)
image

## Title Message

I'd suggest the following text changes to the text message.

`rgb(${backgroundColor})` (wrap the RGB values)

Plus:

```js
? 'has excellent contrast!' // <--- excellent
: smallAA
? 'is AAA for larger text' // <--- replace AAA with "satisfactory"?
: largeAA
? 'suitable only for large text' // <--- replace with "poor" (our audience won't limit like this)
: 'fails contrast guidelines')
```

## Badge Colors

It doesn't seem like the icons/message/badge colors sync. Note how I have a half start + full start with green backgrounds. It's not immediately obvious to me why though. This feels like a mistake?

Screenshot 2023-02-02 at 2 05 58 PM

PR #850, opened 2023-01-19
Updated to match latest svelte-create options

niktek:
While doing some updating for create-skeleton-app I noticed there have been a few changes to SK library projects.

I diffed the changes between a fresh SK lib project and ours - most are superficial (preprocess{ preprocess{}} is more clearly named with {vitePreprocess{preprocess{}} and vite.config is now TS with types.  The main one of concern was the tsconfig.json, but it looks like the extra options we had added ourselves are now being handled by the included `"extends": "./.svelte-kit/tsconfig.json"` base configuration.  There are still some differences in things like the "lib" section, but having gone through the entire docsite/testing/publishing process - everything seems to be working fine.

Had to make a change for an unused Alert on install page and fixed the SvgIcon vitest to make everything work.
svelte-kit sync is now recommended to be run more often to ensure that types are properly generated.


endigo9740:
@niktek Funny enough I had fixed a few of these in my pending PR as well. Looks good though, thanks for this!

PR #866, opened 2023-01-22
ListBox value during Form submission

Sarenor:
Added prop to optionally render a hidden input that contains the selected values to ListBox so that the values automatically get submitted during Form submission.

It has the same issues that the SubmitFunction -> FormData.append() solution that I proposed initially has: It's still FormData and can therefore only take strings or blobs.

That means that strings work perfectly well, numbers and arrays will have to be coerced because they are strings like this:
```json
{
    "selectedString": "stringvalue",
    "selectedNumber": "10",
    "multiSelection": "selectionOne,selectionTwo,selectionThree"
}
```

It's still a better solution than having to include a custom SubmitFunction.

On the other hand, with a custom SubmitFunction we could reset the ListBox to an unselected state like you'd expect it from a Form, that's currently not possible with the hidden input solution proposed here

### DIFF
--- a/src/lib/components/ListBox/ListBox.svelte
+++ b/src/lib/components/ListBox/ListBox.svelte
@@ -11,6 +11,8 @@
 	export let space = 'space-y-1';
 	/** Provide to set scrollable listbox region height. */
 	export let height = '';
+	/** Should the ListBox value be included in Form submissions*/
+	export let includeInFormSubmission = false;
 
 	// Props (Item)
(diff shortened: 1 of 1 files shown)


endigo9740:
@Sarenor Per our discussion today I'm going to close this, but we will have it available for review if need be. I might have you review the form components as I get through the polish pass review in https://github.com/skeletonlabs/skeleton/pull/862

PR #867, opened 2023-01-23
Chore/analytics

niktek:
Not sure why it's creating a new PR, maybe because the other one was closed


endigo9740:
Pinging @Sarenor on this so he sees this. He mentioned taking a stab at this too.

Issue #922, opened 2023-02-03
Polish Pass: Utility Features


endigo9740:
> **Warning** - this ticket is a work in progress. If you have suggestions please add a comment below!

Base on:
- https://github.com/skeletonlabs/skeleton/issues/782

## Goals

The goal of this pass will be to improve, refine, and finalize all Utility features in Skeleton.

- [x] Fix and address known issues and bugs
- [x] Refactor code, cleanup styles, and improve edge cases
- [x] Improve documentation for each feature
- [x] Expand with new features as relevant
- [x] Make overlays not suck, or at least suck less
- [x] Ensure Vitest test cases are passing
- [x] Resolve `FIXME` issues ([reference](https://user-images.githubusercontent.com/1509726/216708976-ea8c37a1-adbf-46e5-a09f-095402907c6e.png))
- [x] Update all style props to use `CssClasses` type

### Local Storage Stores

- [x] Update to the latest upstream version 0.40.0

### Code Block

- [x] Minor visual refresh and code refactor
- [x] Property `buttonCopy` renamed `button`, default value is now `variant-soft`
- [x] Added `buttonLabel` and `buttonCopied` for basic i18n support
- [x] Added style props for `blur` and `shadow`
- [x] Prepare documentation "previewer" component

### Drawers

- [x] Docs: make it clear these are a singleton
- [x] Greatly improved documentation and setup instructions
- [x] Provide advanced tip for animating background contents, per [#637](https://github.com/skeletonlabs/skeleton/issues/637)
- [x] Refactor of Drawer animation logic
- [x] Added `shadow` property for Drawer element
- [x] Added `regionBackdrop|Drawer` props. Available as overrides as well!
- [x] BREAKING: Removed drawer `margin` in favor of backdrop `padding` (provides better results)
- [x] Rounded corners by default, based on theme settings (set `rounded-none` to disable)

### Lightswitch

> **Warning** - this utility feature has been complete rebuilt from the ground up.

- [x] Lightswitch component rebuilt with new UI design and new props available
- [x] Now uses a service-like approach, per [#737](https://github.com/skeletonlabs/skeleton/issues/737)
- [x] Provided an alternative to to the component if you wish to use OS preference automatically
- [x] Users can now interface with our methods, or build their own Lightswitch UI
- [x] Provided and documented interface stores, getters, and setter methods.

> NOTE: we need to test [#905](https://github.com/skeletonlabs/skeleton/issues/905)

### Modals

[...]

PR #923, opened 2023-02-04
fix max toasts.

munxar:
## What does your PR address?

The Toast component renders max + 1 toasts, which contradicts the description of the max property.
Added test and fixed implementation.


endigo9740:
Thank you @munxar. I'll go ahead and merge this. FYI I'm about to kick off a major update to our Utility feature per:
https://github.com/skeletonlabs/skeleton/issues/922

If you come across any similar issues for other Utility features I'd advise you add them to that ticket, so we're not doubling up on work!

Thanks!

PR #946, opened 2023-02-09
fix typo in logo-cloud example doc

joffreybesos:
Small type fix in logo-clouds doc.
There's an extra </span> in the code sample causing a syntax error.


endigo9740:
Thanks, merged!

PR #947, opened 2023-02-09
Feat/GitHub workflow

christophsturm:
## What does your PR address?

This adds a github workflow that runs `vitest`, `pnmp lint` and `pnpm check` on every pull request and push to dev

I disabled all currently violated eslint rules, and autofixed those that could be autofixed.

`pnpm check` still reports errors so it should probably be disabled for now. (or kept as a reminder to fix it soon)

Also I hope that the workflow really works, because the only way to really test a github workflow is to submit a pull request that runs it.


endigo9740:
@niktek tagging you to review this as well.

@christophsturm We have a guidelines that non-trivial updates require prior discussion, per:

This saves everyone time if we decide the the work is not something we wish to move on yet. I see you asked in Discord, but did not follow Dominik's instruction to reach out in #contributors (again for prior discussion), and never got a proper thumbs up from Nik or myself (core maintainers).

First things first, I need to know more about how this works:

- When a user submits a PR that fails one of these tests, how are they informed of failure?
- Does this block the PR from being submitted altogether?
- What about a commit to a PR already present? Will it fail if there's a missing semicolon, etc?

General idea is we use draft PRs a lot, and those aren't always in a complete and production-ready shape. They're not always ready for linting or tests to be run against them. We might be testing a proposed change and failure messages or anything blocking that process would be more detrimental than useful.

christophsturm:
@endigo9740 I did read the guidelines, I just made this pr anyway because I had the workflow file in one of my private projects already, so I thought I make a PR to just show the general idea. 

now to your questions, they all are not really specific to this workflow, but are more generally about github ci workflows
>When a user submits a PR that fails one of these tests, how are they informed of failure?
Does this block the PR from being submitted altogether?
What about a commit to a PR already present? Will it fail if there's a missing semicolon, etc?

endigo9740:
@christophsturm First off we appreciate the attempt to help, but unfortunately I don't think this is something we're going to be able to accept. There's a number of flaws in execution here:

1. Acknowledging but failing to follow our contribution guidelines, especially prior discussion
2. Disabling, rather than resolving, linting issues
3. Plus this is an inherently flawed way to go about this, it's occuring post-commit rather than pre-commit

We've had some internal discussion about introducing Husky Hooks, which would allow for these sorts of checks in a pre-commit stage. I think that'll be MUCH more valuable and much easier for folks to act on, given they get the immediate feedback and we're not deploying code that doesn't met a certain threshold of expectations.

If you wish to help with this effort, I might suggest you sync with @ryceg on Discord (he's an Aussie, so mind the timezones) and let's discuss this upfront and decide on a plan to act. Thanks!

christophsturm:
no problem, i just wanted to show how I would do it. this was basically the start of the "discussion" step

>Disabling, rather than resolving, linting issues

I disabled them because I did not want to do everything in one PR. They are not meant to stay disabled

> Plus this is an inherently flawed way to go about this, it's occuring post-commit rather than pre-commit

This is meant to run in addition to checks that the users run pre commit. It is for checking pull requests before they are merged.

endigo9740:
All good - even closed we can still reference and learn from this PR. The core team and I are discussing some ideas for this but no conclusions yet. Basically we're still in flux about where and how these checks are going to be implemented. Once we can wrap our head around that we can start figuring out some actionable goals.

PR #953, opened 2023-02-10
Tabs docs: Fix href on link to local-storage-store

carlosg2:
Updated link to local-storage-stores docs


endigo9740:
Thanks @carlosg2

PR #936, opened 2023-02-07
Polish Pass: Utilities

endigo9740:
Fixes #922

PR #969, opened 2023-02-15
Update default `inline` value to match code

t-miller:
## What does your PR address?
The code defaults the `inline` argument of tooltip to `true` but the docs showed it as `false`, leading to confusion
https://github.com/skeletonlabs/skeleton/blob/master/src/lib/utilities/Tooltip/tooltip.ts#L45


endigo9740:
@t-miller I'm right in the middle of a major overhaul to basically ALL Utility features, including tooltips. See the ticket and PRs here:
- https://github.com/skeletonlabs/skeleton/issues/922 (pinned)
- https://github.com/skeletonlabs/skeleton/pull/936

The way tooltips operate is about to change in a dramatic fashion, which means this change is great, but won't be relevant for the next release or beyond. I appreciate it though, but for now I'm going to opt to pass on this one. I do welcome a review on the active PR though - feel free to provide feedback in the ticket or on Discord. Thanks!

PR #1000, opened 2023-02-21
Exported current step from Stepper for easier dynamic styling and behaviour.

Sarenor:
## Before submitting the PR:
- [x] Did you update documentation related to your new feature or changes? Updated automatically. Prop Discovery is awesome!

## What does your PR address?

Exports the current Step from the Stepper component so that conditional classes and other dynamic behaviour can be easily implemented without the user having to set up event listeners and make assumptions about the initial default state of the Stepper.

### DIFF
--- a/src/lib/components/Stepper/Step.svelte
+++ b/src/lib/components/Stepper/Step.svelte
@@ -49,11 +49,13 @@
 		$state.current++;
 		/** @event { $state } next - Fires when the NEXT button is pressed per step.  */
 		dispatchParent('next', { step: stepIndex, state: $state });
+		dispatchParent('step', { step: stepIndex, state: $state });
 	}
 	function onBack(): void {
 		$state.current--;
 		/** @event { $state } back - Fires when the BACK button is pressed per step.  */
 		dispatchParent('back', { step: stepIndex, state: $state });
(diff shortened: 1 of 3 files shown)


endigo9740:
@Sarenor Hey, so the binded prop doesn't really meet my expectations as a user - I feel like I should be able to set the step value this way, not just see it update. How about we do the following instead:

- Revert the current changes
- Implement a new prop `start: number` that passes the desired starting step into the Stepper. It would have a default value of zero, and we would run `$state.current = start` on the component initialization
- Then we add a new dispatched event called `on:step`, which fires whether you're navigating forward or back. It would just a a combined version of `on:next` and `on:back`

This does a few things for us:
1. Makes it explicit how and where you set the initial step value
2. Keeps the `on:step` "actionable", meaning you not only get notified of the change through the event, but can use it as a function to trigger code when this occurs.

Thoughts?

Sarenor:
Already agreed to the changes on Discord, let's see how well I code on just the laptop screen while in the train 😁

Sarenor:
And that should do it. Went a bit explicit in the documentation, just tell me if you think it's to much and I'll remove it again :)

endigo9740:
@Sarenor This is PERFECT, thank you sir! Merged!

PR #1045, opened 2023-02-26
Autocomplete Complete

JustBarnt:
## Before submitting the PR:
- [X] Does your PR reference an issue? [New Component Autocomplete](https://github.com/skeletonlabs/skeleton/issues/377)

## What does your PR address?
Still a work in progress.

Below are items that I have progress or completion on.

___

- [x] - Link in alphabetical order
- [x] - Renamed to `Autocomplete`
- [x] - Reverted page documentation in favor of new DocShell when ready
- [x] - Improve `filteredValues()`
> - [x] No search term present
> - [x] Search term present, do a fuzzy search against a list of options
> - [ ] Some way to provide an array of values, then filter the list to only options that DON'T include this (scenario: remove suggestions for input chip values that have already been selected)
^ I'm referring to this as "exclude" filtering? I'm 100% open to a better name.
___
- [x] Filtering with a whitelist included (Need to watch video on setting up tests, but I did quick test myself.) 
- [x] Use Skeletons Navigation Styles ( I believe I have achieved this) as well as pulling out styles that were originally inline.
- [x] updated `.item` to `.autocomplete-item`
- [x] Created `autoComplete` action that controls the show/hiding of the list items
- [ ] Started work on getting keyboard functionality to work, I will need some guidance on this as it is very rough still. (Grabbed the base functionality from `Popup.ts`)

### TODO
- [ ] Learn vitest
- [ ] Selecting an item and then click in the input again, shows the list pre-filtered because the input has value, to me this would not be the expected functionality. So I need to come up with a solution. 

@endigo9740 If you get a quick chance to look at this- its much appreciated. I just want to make sure `filterValues()` is working more inline with what you were describing above. 

Mainly the `excludes` filtering is what I am calling it: My interpretation is if the user has an array of values, uses the `excludes` filtering and they search foo, the popup below the input would update and show the user "bar and fizz" because foo is already typed in the input. 

Excludes Example
>Foo is not present in list, nor is FooBar (this could be considered a bug, but I think should be easily fixable if `regex` is used compared to `includes`?) 

image

___
IF I'm correct in that functionality I need some help figuring out how to keep removing from the list as they type. I was thinking would allow or allow users to specify `separators` to indicate when to search for a new string?


jameslounds:
>  Selecting an item and then click in the input again, shows the list pre-filtered because the input has value, to me this would not be the expected functionality. So I need to come up with a solution.

What worked for me (in my interim implementation until this is released) was clearing the input when it's clicked, and selecting the text in the input on keyboard focus. I've considered adding a listener for Command + Z to put the last selected value back.

[MUI](https://mui.com/material-ui/react-autocomplete/) doesn't consider a filter to be active when you've clicked an option and go back on the input
[Mantine](https://mantine.dev/core/autocomplete/) leaves the input alone on click, but selects the text in the input when focussed with the keyboard

endigo9740:
Hey Brent, so what you had was great. A lot of the core concepts are perfect. There's just some odds and ends that didn't quite align with how I was expecting this to work. Given the number of changes I hope you don't mind me committing these alterations direction. Sometimes it's just easier to show than tell!

I've noted everything in detail below. You should be able to pull the changes down and review locally.

## HTML/CSS

- Modified the HTML template to use a native `<select>` element for all the a11y benefits that provides
- Styles are now set via `.select` (which can be replaced in `element` to go "headless")

## Logic

- We don't want to control the visibility of this component, that's for the users. We're building primitives, remember, which means folks can drop this into a Skeleton Popup utility and do everything you built with the power of Floating UI!
- Given this I've disabled your Action - it's super cool, but redundant! I've not deleted it in case you want to grab it and save it for later use in your own projects!
- I've refactored the filter logic to make it a bit more "DRY" and modular. Still operates in the same fashion.
- I've adjusted prop `input` to semantically represent the input value (your target input)
- I've adjusted prop `options` to semantically represent the list of objects that populate the select > objects
- The `mode` type is simple enough we can set that inline, removing the need for `types.ts`
- I've modified the custom dispatched event to be `selection` so it doesn't overlap with native events.
- Finally `on:selection` returns the entire object row, so users can utilize any key/value pair from within this

## Remaining Tasks

- I think the whitelist you've instituted might be a bit too open ended. Reading the entire data set as a string is too loose. I might recommend you test each array row independently using [some()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some)
- The `select[size]` element is a bit more rigid in height. We'll need to find a clever way to adjust the height automatically based on result length, while still specifying a max height. I'm open to ideas.
- We do lose the ability to hit use keyboard navigation to drop from the input into the autocomplete, but tab should still work. Maybe this is where we keep the Action as an optional supliment?
- Fill free to nix the commented line items if you agree with their removal

Feel free to ping me on Discord if you want to discuss this in detail. I'm also down to jump on a quick Zoom call if that's easier! But let me know what you think about this direct. I think it offers a ton of flexibility without trying to reinvent the wheel where we don't need to.

jameslounds:
I’m interested to hear why the choice of a `select` element for the actual input, rather than a hidden one like in `InputChip`. On Safari (especially mobile), the styling doesn’t apply, and I’m not sure if this behaviour is intended.

The `select` also causes the filter to be reapplied, and the currently focussed element is selected, making the third item in the dropdown often unreachable.

I'm getting really excited watching this PR come together, great job!

https://user-images.githubusercontent.com/20153440/221989112-f4a25ead-fd33-4be5-ad44-96be3a27d3bc.mp4


https://user-images.githubusercontent.com/20153440/221989215-0440db89-6704-4c42-968a-08eadf9ad191.mp4

endigo9740:
> I’m interested to hear why the choice of a select element for the actual input, rather than a hidden one like in InputChip. On Safari (especially mobile), the styling doesn’t apply, and I’m not sure if this behaviour is intended.

The idea here is the autocomplete is a supplementary UI, not the form input itself that stores it's own state for the form. The choice in Select is primarily for presentation and a11y benefits that comes from native inputs.

> The select also causes the filter to be reapplied, and the currently focussed element is selected, making the third item in the dropdown often unreachable.

Hmm, not a fan of how that's handled on your mobile device. That's really frustrating. I can't tell from the UI, is that an iPad or Android tablet? If it's the latter, which browser?

jameslounds:
> Hmm, not a fan of how that's handled on your mobile device. That's really frustrating. I can't tell from the UI, is that an iPad or Android tablet? If it's the latter, which browser?

iPad. Also looks a little janky, but better on desktop safari.

endigo9740:
> iPad. Also looks a little janky, but better on desktop safari.
> I think this is just an annoying limitation webkit imposes (again). Other UI libraries seem not to use a `select`, but this may be for another reason I haven't considered

Yeah the UI presentation is the simplest part of all this, so honestly if this is an issue we can change that out. The more important bit is that the data mapping and filtering logic are working as intended.

JustBarnt:
@endigo9740 Another draft update! 

> I think the whitelist you've instituted might be a bit too open ended. Reading the entire data set as a string is too loose. I might recommend you test each array row independently using [some()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some)
---
- [x] Whitelist searching has been refactored based on what you talked about above, hopefully this is more inline with what you are envisioning. 

As we discussed, I changed the select back to a div as we were experience issues with the default way select elements acted. 

With that change, I have added what I think should be the correct `aria-labeling` for the items that make up the autocomplete component itself. I left the `select` options in, but commented out in-case you wanted to take a stab at seeing if you can figure out a way to get the select element to work opposite of how it should. 

## Remaining
- [ ] a11y items
- [ ] Keyboard navigaion
- [ ] Doc creation
- [ ] Test's
- [ ] Styling

After this review, assuming everything is okay- These items should be able to be started!

jameslounds:
Something really tiny, but I think you've misunderstood how tabindex works. 0 and -1 are the only values you'll typically need to use. Setting it to 0 adds the element to the tab index, making it possible to tab focus the element in the order it appears in the DOM. 
Setting them incrementally makes the first one focus as expected, then everything else that is tabbable on the page after the autocomplete, then the browser controls (address bar etc), then the second item in the list (with tabindex 1).

https://www.a11yproject.com/posts/how-to-use-the-tabindex-attribute/

JustBarnt:
> Something really tiny, but I think you've misunderstood how tabindex works. 0 and -1 are the only values you'll typically need to use. Setting it to 0 adds the element to the tab index, making it possible to tab focus the element in the order it appears in the DOM. Setting them incrementally makes the first option focus as expected, then everything else that is tabbable on the page after the autocomplete, then the browser controls (address bar etc), then the second option in the list (with tabindex 1). If there were 2 or more `Autocomplete` components on the page, the first option would tab for each as above, then the address bar etc, then the second option for each component (tabindex=1), then the third option for each (tabindex=2).
> 
> https://www.a11yproject.com/posts/how-to-use-the-tabindex-attribute/

Ah yes, thank you! I will get that fixed real quick!

jameslounds:
I was thinking more about handling multiple selected options. I don't think that a comma separated list in the input is a great solution. We could leverage the existing chips and input group to make something that looks similar to the `InputChip` component.

https://user-images.githubusercontent.com/20153440/223875291-10ff9b7c-de72-46ab-bc23-c62288cd28cf.mov

This is achievable with the component in its current state, just adding a different `on:select` listener which removes the selected item from the whitelist.

On that note, I think having a `mode` based algorithm choice running against  _the same `whitelist` array_ doesn't make semantic sense. In exclude mode, the whitelist acts as a blacklist. Perhaps passing (up to) two arrays, `includeList` and ` excludeList`, and check against both when filtering would be better?

```svelte
<label class="label space-y-2 w-full" for="autocomplete-search">
		<h3>With Dynamic Whitelist</h3>
		<div class="input-group grid grid-flow-row rounded-container-token">
			<input
				class="input autocomplete"
				type="search"
				name="autocomplete-search"
				bind:value={searchValue}
				placeholder="Begin typing to filter..."
			/>
			{#if selected > 0}
				<div class="input-chip w-full flex space-x-2 py-2 !justify-start">
					{#each selectedItems as item (item.label)}
						<button
							class="chip variant-filled-primary focus:variant-ringed-secondary"
							on:click={() => {
								autocomplete3Items = autocomplete3Items.filter((i) => i !== item);
								whitelist = [...whitelist, item];
							}}
							transition:slide
						>
							{item.label} <span class="pl-2">&times;</span>
						</button>
					{/each}
				</div>
			{/if}
		</div>
		<AutoComplete bind:input={searchValue} options={anythingList} bind:whitelist on:selection={onSelection}  />
	</label>
```

I've also had some thoughts on keyboard interaction, so I'm interested to hear yours as well. I've got  _a_ solution using this PR,  by adding prop in `Autocomplete` to accept a reference to the `inputEl: HTMLInputElement`. Once mounted (and the ref is not undefined), an event listener is added to the element for arrow keys and enter, updating a local `focusIndex`.

I've tried to think of a way to do this with a svelte action, but I can't wrap my head around how the data flow would work. I'm struggling to find a way that makes sense to pass the focusIndex to the component, while allowing it to still select the focussed element in the (filtered) array.

endigo9740:
> I was thinking more about handling multiple selected options. I don't think that a comma separated list in the input is a great solution. We could leverage the existing chips and input group to make something that looks similar to the InputChip component.

@jameslounds just so you're on the same page with us, some amount of discussion around this feature has happened on Discord. The intention of the autocomplete feature is to be agnostic to the element it's assigned to. Meaning, it could be native input, an input chip component, etc. As long as the data can be exchanged it should work.

It's too often the case that people/libraries try to tightly couple the input and autocomplete elements together. But we're very intentionally trying to avoid that, by design. So in this case our solution wouldn't just work *like* the input chip component, it would directly interface with our input chip, and potentially some component that hasn't even been introduced yet.

If you're not already part of our Discord I would encourage you to join and keep an eye on the `#contributors` channel. It is great when we don't have to repeat yourselves multiple times in multiple places :) Totally understand if Discord isn't your thing, but I wanted to provide that context here so we're not repeating ourselves.

JustBarnt:
@endigo9740 Quick update to the filter functionality. Give it a good look through and let me know your thoughts, I think we can remove the commented out functions. Also implemented our discussion here. 

Tonight I'm gonna work on setting up the test, doc page content, and what ever else you have left after reviewing.

endigo9740:
@JustBarnt I've done another pass through this. So a few things here:

1. Within `ismatch` we can just use `Object.values(obj)` to find the values. Much less code needed.
2. I agree with dropping the dedicated functions, I've nuked those comments
3. I've reimplemented the design using our `list-nav` element styles, at least for now.
4. I've removed the popup - currently the input is losing focus when the popup is triggered. We'll need a new popup event type to resolve this. Some work is occurring [here](https://github.com/skeletonlabs/skeleton/pull/1174)?

Additionally I've found another bug. Currently if you search for `bar` you should get both `Bar` and `FooBar` as suggested options. But this is not occurring:

Screenshot 2023-03-20 at 2 40 37 PM

I'll leave that to you to try and track down and resolve.

Also just general rule of thumb - no need commenting things for preserving the history, that's what version control is for. If you're not using something, just remove it! If something is temporarily disabled though, comments are fine (ala the popup)

JustBarnt:
> Additionally I've found another bug. Currently if you search for bar you should get both Bar and FooBar as suggested options. 

@endigo9740 I know exactly why that's happening- to me its a bug, as I would not expect to get `Foobar` when searching bar, because foo- comes before it. 
> `if (value.substring(0, input.length) === input)`

This is why, I'm checking if the sub-string of characters in value is equal to input, so `foobar` wouldn't match. 

But, I can remove that if in this case then. Haha

JustBarnt:
@endigo9740 Completed the first draft of doc writing. Looking forward to your critique as I'd like to improve my technical writing. 

- [x] Removed conditional that prevented `bar` from including `foobar` in its results.
- [x] Wrote first draft of docs.
---
### Questions
I noticed a couple of things I have questions on. 
1. I noticed you added an `{@html v.label}` for displaying the button text, I was curious as too why you did that instead of just `{v.label}` 
2. `prunedRestProps` is not longer being called in the parent container for removing `class` from any `restProps` being passed. Was that intended or just missed when converting the `nav` list? 
3. I now get a warning on `nav` or `ul` with role="listbox"` which is":
    - `A11y: Non-interactive element <ul | nav> cannot have interactive role 'listbox'`
    Not sure what needs to be done for that unfortunately so I'm looking to you to see if that's a   valid thing, or if VSCode is just on something haha.
4. Lastly I wanna propose passing the `ref` to the input element to the `Autocomplete` itself and create an `onSelection` event inside the autocomplete that handles setting the value of the input its passed. This would prevent having to use multiple event handlers outside the `Autocomplete` if the user has multiple `Autocomplete` components on the same page.

Hugos68:
Width of the demo input needs to be adjusted to be responsive for mobile devices

endigo9740:
With the latest changes @JustBarnt and I have agreed this is ready to merge. Great job Brent!

Just in case anyone else is wondering, we WILL be pairing this with the Popup feature, but we need some adjustments to the action's trigger events and focus before this will work properly. We'll then also aim to improve the keyboard interaction.

Issue #1049, opened 2023-02-27
Audit and Improve Doc Content


endigo9740:
Per: https://github.com/skeletonlabs/skeleton/issues/783

The goal of this ticket will be to divvy up ownership of several sections of the documentation. Each volunteer will be assigned an allotment of pages that they will own. You should audit, review, and either submit feedback - or better yet send a PR with fixes and adjustments in place.

## How to Participate

> **Warning** please DO NOT volunteer if you cannot have your changes completed by Sunday March 5th.

1. Reach out in the `#docs-overhaul` channel on [Discord](https://discord.gg/EXqV7W8MtY) to volunteer
    - Post your GitHub username (I'll assign you below)
    - Let us know which section you wish to be assigned
2. Immediately create a draft PR (pull request), including your checklist as shown below ([reference](https://www.skeleton.dev/docs/contributing))
    - Important: You will not be considered assigned until a draft PR is available
4. As you complete changes per page, commit it, then tick the box in the PR to alert everyone of your progress
5. Set your PR to "ready for review" when all sections are complete
6. Finalize all changes by Sunday March 5!

### Can't submit a PR?

View the dev branch docs here:
https://dev.skeleton.dev/docs/introduction

You can still help by locating issues for other volunteers, reviewing the blog posts, or helping with PR reviews. When submitting feedback, tag the assigned user, or consider keeping feedback feedback localized in each PR thread. Use a checklist so we can tick off items as completed!

## Responsibilities 

> Try to follow any obvious existing conventions. Contact a contributor on Discord for additional questions!

- Scan the page for typos, grammar, etc (including each tab)
- Check the page for broken links
- Keep headings short, consistent, and semantic to the section (no wrapping in Table of Contents)
- Keep paragraph text short, concise, and descriptive - ideally 1-2 lines max
- Use `h2` for main headers, `h3` for subheaders
- Use `<section class="space-y-4">...</section>` to divide page sections
- `prop`, `slots` and `event` tab data is auto-documented via inline JSDocs comments ([reference](https://www.skeleton.dev/docs/contributing/documentation))
- Ensure auto-documented tables aren't missing props (it happens!) and descriptions are semantic
- The `classes`, `params`, and `keyboard` tabs are manually maintained, which means more room for error
- Avoid inserting Font Awesome icons in code snippets - Skeleton is icon library agnostic!
- On CodeBlock snippets remove trailing indents so that they are not copied over to the clipboard

[...]

PR #1054, opened 2023-02-27
Svelte group 1 docs review

niktek:
Contributing to:
- https://github.com/skeletonlabs/skeleton/issues/1049

Checklist
- [x] Clipboard
- [x] Filters
- [x] Focus Trap
- [x] Accordions
- [x] App Bar
- [x] App Rail
- [x] App Shell
- [x] Avatars
- [x] Conic Gradients
- [x] File Buttons
- [x] File Dropzone
- [x] Input Chips

### DIFF
--- a/src/docs/layouts/DocsShell/partials/PanelProps.svelte
+++ b/src/docs/layouts/DocsShell/partials/PanelProps.svelte
@@ -22,7 +22,7 @@
 		<aside class="alert variant-ghost-warning">
 			<!-- prettier-ignore -->
 			<div class="alert-message">
-				<p>This component implements <a href="https://svelte.dev/docs#template-syntax-attributes-and-props" target="_blank" rel="noreferrer">restProps</a>. This passes all extra attributes to the component's <code>{pageData.restProps}</code> element.</p>
+				<p>This component implements <a href="https://svelte.dev/docs#template-syntax-attributes-and-props" target="_blank" rel="noreferrer">restProps</a>. This passes all extra attributes to the component's <code>{pageData.restProps}</code> elements.</p>
 			</div>
 		</aside>
 	{/if}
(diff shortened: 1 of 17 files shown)


Sarenor:
You might want to decide on either space or no in Clipboard ;)
image

Focus Trap:
> Apply use:focusTrap then the action value to either true or false to enable or disable focus.
then _set_ the action value?

AppRail:
> This component implements [restProps](https://svelte.dev/docs#template-syntax-attributes-and-props). This passes all extra attributes to the component's AppRailTile element.
Should probably be AppRailTile elementS?
> Apply abitrary classes (in props)
Probably the whole repo needs a Find & Replace for abitrary to arbitrary
> Page State
Isn't that also handled with AppRailTile values & the selected store? I feel like we should mention that that's the default, recommended way

AppShell
> Apply abitrary classes (in props)
Probably the whole repo needs a Find & Replace for abitrary to arbitrary

FileDropzone
> Apply abitrary classes (in props)
Probably the whole repo needs a Find & Replace for abitrary to arbitrary

Hugos68:
Under the clipboard input value section it states "To copy a form input value", shouldn't this apply to all input value's and therefor just be "To copy an input value"? No clue if I'm nitpicking here but something I noticed.

endigo9740:
> Under the clipboard input value section it states "To copy a form input value", shouldn't this apply to all input value's and therefor just be "To copy an input value"? No clue if I'm nitpicking here but something I noticed.

The intention is it copies the input that you target. So perhaps insert that term?

"To copy the target input value...."

endigo9740:
@niktek if all changes are current you've got my go ahead to merge!

riziles:
@niktek , seems to be a minor issue on https://skeleton-docs-oi8f3pcbb-skeleton-labs.vercel.app/components/app-shell -> slots -> header:

image

riziles:
... apologies if I'm looking at a stale page.

riziles:
I think this is the line? 	

https://github.com/niktek/skeleton/blob/f4a0fa09a55f9f69d93f2e56d85bb79c1288a8c0/src/lib/components/AppShell/AppShell.svelte#L3

Sorry. Wait out of my depth right now.

riziles:
moved here:

https://github.com/skeletonlabs/skeleton/issues/1049#issuecomment-1452915286

Issue #1060, opened 2023-02-28
allow `unstyle` class to inhibit `<del>`/`<ins>` typography


obayemi:
### Describe what feature you'd like. Pseudo-code, mockups, or screenshots of similar solutions are encouraged!

right now it is impossible to prevent skeleton from styling <del> and <ins>

### What type of pull request would this be?

Enhancement

### Any links to similar examples or other references we should review?

https://github.com/skeletonlabs/skeleton/blob/dev/src/lib/styles/typography.css

endigo9740:
Fixed as part of:
https://github.com/skeletonlabs/skeleton/pull/1061

@obayemi if you use `fixed #1061` it'll link the issue and PR so that the former is closed when the latter is merged.

Some GitHub magic! :)

obayemi:
oh yes thx, I thought just mentioning the issue was enough :sweat_smile:

PR #1061, opened 2023-02-28
allow unstyling del/ins html components

obayemi:
## What does your PR address?

Allow unstyling `<del>`/`<ins>` with `unstyle` or `prose` classes, see #1060


endigo9740:
@obayemi seems to be working perfectly. I'll merge this now.

Just note our next release is going to be a little further out than our standard 2 week cycle. We're hoping this will be our v1 , so thank you for your patience in the meantime!

But given this is merged it means it 100% will be part of this next release. Thanks!

PR #1085, opened 2023-03-04
Fix App Rail code example

ryceg:
Was missing an `'`.


endigo9740:
@ryceg thanks, just be sure to tag the original ticket so it's closed when we merge! I'll grab this one!

PR #1103, opened 2023-03-07
docs/examples/ all CodeBlock space trim and whole line with return #1089

dirk-bester:
- [X] Does your PR reference an issue? Kicked off from a discussion and code review by @niktek .
- [X] Did you update and run tests before submission using `npm run test`? Note some expected prop 'name' issues.
- [X] Did you update documentation related to your new feature or changes? This is pure documentation.

# What does your PR address?
This gives us consistent cut and paste behavior of all code samples.

# Changes
- All multiline CodeBlock items now end on a white space free newline.
- All single CodeBlock line items now end on a /n
- All /t /n in multilines expanded
- Removed some spurious trim() that did nothing, or specifically added whitestace to trim.

# Exceptions:
- Shell commands
- The quickstart Avatar include
- Some cases where a text variable was used. Probably a TODO for later.

# Note
- I would add this to our style manual?

### DIFF
--- a/src/routes/(inner)/actions/clipboard/+page.svelte
+++ b/src/routes/(inner)/actions/clipboard/+page.svelte
@@ -41,8 +41,8 @@
 				</button>
 			</svelte:fragment>
 			<svelte:fragment slot="source">
-				<CodeBlock language="ts" code={`const exampleData: string = 'This text was copied by the Skeleton clipboard action.';`} />
-				<CodeBlock language="html" code={`<button use:clipboard={exampleData}>Copy</button>`} />
+				<CodeBlock language="ts" code={`const exampleData: string = 'This text was copied by the Skeleton clipboard action.';\n`} />
+				<CodeBlock language="html" code={`<button use:clipboard={exampleData}>Copy</button>\n`} />
 			</svelte:fragment>
 		</DocsPreview>
 	</svelte:fragment>
@@ -54,8 +54,8 @@
 		<div class="space-y-4">
 			<h2>Copying HTML Contents</h2>
 			<p>
-				To copy the <em>innerHTML</em> for an element, set a <code>data-clipboard</code> on your target, then provide
-				an <code>element</code> reference to the action.
+				To copy the <em>innerHTML</em> for an element, set a <code>data-clipboard</code> on your target, then provide an
+				<code>element</code> reference to the action.
 			</p>
 			<DocsPreview background="neutral">
 				<svelte:fragment slot="preview">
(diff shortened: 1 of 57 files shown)


dirk-bester:
In order to be careful with this I edited each file then did another pass or two verifying.
Later I verified all the git diffs involved adding a /n or newline at end of a multiline or expanded /t or /n in a multiline.

Also a handful of cases where Prettier did its thing, but in a project approved way.

endigo9740:
@dirk-bester I hate to have to go down this path again, but I don't think this is a PR we will be able to accept.

Per a previous PR, I mentioned we have a pre-requisite of prior discussion before submitting non-trivial PRs. Modifying ~60 files and changing a core behavior of our Code Block component definitely falls outside the scope of "trivial".

Best I can tell, this is the only place you proposed this work. However, no one from the core team responded to or gave a thumbs up to move forward with this. In fact no one responded to it at all. So it's worrisome you moved forward with the work with full knowledge of our guidelines in this regard:
https://discord.com/channels/1003691521280856084/1003699523522142399/1082016923673444372

That said, I reviewed internally with our team, and I think we're of the mindset that this is not a change we would implement. Even if it was, we would not approach it in this fashion. The better way to have approached this would have been to replace the `.trim()` in the code block with the new line. That's a single line change isolated in one place that would affect all instances of the component, and ensure all future code blocks get equal treatment.

That said, I want to state clearly here that **this is NOT a change we need or want**. If you send another PR with the changes above it will be denied as well. I'm making this as clear as possible: please do not do this.

Given that we've now had a few instances here where instruction was not followed and PRs were not up to our expectations I'm going to ask you kindly refrain from submitting new PRs for the time being. We welcome you in the community, your feedback, and we encourage you to continue learning and understanding how Skeleton was built and open source contributions are handled for a project of this size. But at this time these PRs are eating up valuable time for work we cannot accept, so we're gonna have to ask for a timeout for a bit.

If you have questions about this please free to respond here or in the `#contributor` channel on Discord and I'll respond. Thanks for your understanding.

dirk-bester:
I chose the Draft PR type which according to the guidelines can kick off discussion. It is a large change, I definitely have no expectation for it to be in 1.0. Or even to be accepted at all.

In that previous PR, I was specifically asked to fix trailing spaces. That lead me to discover why pasting of the sample code is weird and lacks the return at the end one expects.

I did think about handling it in the CodeBlock code. But late night and tired made me think that would be less acceptable. That would certainly be a better way to accomplish the same thing.

endigo9740:
Just so we're clear, our guidelines are:

- Discussion first
- Get approval from a core team member
- THEN submit draft PR

Our suggestion for sending a draft PR early is only mean when you get to that draft step. Because some folks opt to spend days or even weeks working on changes without sharing them so we can provide guidance and feedback. We need to be able to course correct if things are off track once development begins.

I stated this before, but a draft PR is not prior discussion. Believe or not, we're actually trying to help you out there. We do not want people working on features we can't or won't accept. Not only is a waste of everyone's time, but then it puts us in a hard place and folks may feel there's some level of obligation to accept work that's already been done. But we don't want to be put in that situations. I hope you can understand. It's a small courtesy that helps all involved.

PR #1115, opened 2023-03-08
Update theme generator so no error occurs while typing in hex colors 

### LINKED ISSUE #1111: Theme generator does not update when a hex value is written into it
### Current Behavior

There seems to be a bug with the theme generator, so that when you type a hex colour value into the visible input field that is less than 6 hex values it won't update the colour, even once you finish typing all 6 hex values. When you have 6 values manually typed in, you need to refresh to get the changes. If less than 6 values are typed in it just doesn't work, even on a refresh. The console shows an invalid RGB value error:

image

### Steps To Reproduce

_No response_

### Anything else?

_No response_

bennymi:
## Before submitting the PR:

## What does your PR address?

Changes made:
- added a validity checker for hex values
- added a contrastReports list variable that calls the `getPassReport` function only when hex value is valid
- html now gets generated with this new list instead of using `{@const contrastReport = ...}` calculation
- added some types
- updated an import in `Swatch.svelte` which was pointing at the wrong type
- added missing `semanticNames` (`'error'` and `'surface'`) to the `tailwind.ts` file

This would Close #1111 

### Tips

### DIFF
--- a/src/docs/layouts/DocsThemer/DocsThemer.svelte
+++ b/src/docs/layouts/DocsThemer/DocsThemer.svelte
@@ -15,9 +15,9 @@
 
 	// Local Utils
 	import { storePreview } from './stores';
-	import type { ColorSettings, FormTheme } from './types';
+	import type { ColorSettings, FormTheme, ContrastReport } from './types';
 	import { inputSettings, fontSettings } from './settings';
-	import { type Palette, generatePalette, generateA11yOnColor, getPassReport } from './colors';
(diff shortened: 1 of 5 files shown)


endigo9740:
@bennymi Rhys implemented the color contrast ratio checker features, and while they work, they do seem to divert from the approach I was using for most functionality. After reviewing a bit I think a refactor of this tool may be in store for our near future. I know he and another contributor were talking about means to improve the color calibration, so we may try to tackle all this at once.

But testing your PR, your change seems to provide a stopgap solution that should get us by for now. I'll go ahead and merge. But let me know if you come across any other odd behaviors.

Thanks for the quick turn around!

PR #1124, opened 2023-03-09
Add some tests, make case insensitive

jameslounds:
Was playing around writing some tests for the component as is, and realised the input is case sensitive. This may well be intentional, so feel free just to ignore this PR, and feel free to use any and all code in this PR without merging it.

I also changed `filteredOptions` to an array, not a function. The name `filteredOptions` sounds like an array, whereas something like `filterOptions()` or `getFilteredOptions()` sounds like a function.

I left in my test around keyboard interaction as it took me some time to figure out lifting the component state and I figured it might be helpful further down the road.


jameslounds:
Sorry, made this on the wrong repo - my bad

PR #1131, opened 2023-03-11
Fix: Optimized sponsor images

AdrianGonz97:
The size of images for sponsors were causing a "jerk" effect when trying to smooth scroll on the home page. Those images have been shrunken from 136kB -> 40kB (Brain & Bones) and 654kB -> 30kB (Luke).


endigo9740:
@AdrianGonz97 ahh, thank you!

PR #1145, opened 2023-03-13
Update `get-started` docs to be a bit more explicit on manual install for stylesheets

jimmyhsu:
## Before submitting the PR:

## What does your PR address?

Hey Skeleton team!

I've been enjoying Skeleton thus far and I wanted to maybe suggest a small doc change! The manual install for stylesheets in the `Get Started` section was mildly confusing as I'm newer to Svelte/SvelteKit (just learned it today!) so I was thinking this could help some folks who may have potentially run in to the same issue.

I assume this won't be an issue to anyone more experienced to Svelte as they'd immediately know, but for newbies it could help feel a bit more grounded (like the `tailwind.config.cjs` example does) so they're a little more assured they're doing the correct thing.

It's a minor change and didn't trip me up for too long (maybe a minute max), but I was thinking reducing every bit of friction could help for adoption!

### Tips

### DIFF
--- a/src/routes/(inner)/docs/get-started/SectionStylesheets.svelte
+++ b/src/routes/(inner)/docs/get-started/SectionStylesheets.svelte
@@ -37,15 +37,19 @@
 					</div>
 				</aside>
 				<CodeBlock
-					language="ts"
+					language="html"
 					code={`
-// Your selected Skeleton theme:
-import '@skeletonlabs/skeleton/themes/theme-skeleton.css';\n
-// This contains the bulk of Skeletons required styles:
-import '@skeletonlabs/skeleton/styles/all.css';\n
-// Finally, your application's global stylesheet (sometimes labeled 'app.css')
-import '../app.postcss';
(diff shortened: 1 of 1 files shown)


endigo9740:
@jimmyhsu as we discussed on Discord, sometimes being more verbose is useful. But in this case all relevant information is already present in the current iteration:

1. We explain the location of the file is `/src/routes/+layout.svelte` (a Svelte file)
2. We explain that this is using `Typescript` per the code block
3. There's only one place TS code can be inserted in Svelte components - inside a `<script lang="ts">` tag

This follows convention through the rest of the app. Which is to provide the minimal information needed. Which as an added benefit makes the information easier to maintain for a team of our size.

The addition of `<slot />` is a bit opinionated, as some folks may have a different layout structure by the time they reach this documentation. For example the CLI install can implement this file using an App Shell. Folks might worry they're in the wrong place or doing something wrong if the code is not a 1:1 match.

PR #1174, opened 2023-03-17
Add `focus` event to `popup`

Cucumberrbob:
## What does your PR address?

Adds a `focus` event to the options passed to `use:popup`. This allows us to trigger the menu when an element is either clicked or tabbed to, and close it when focus moves outside the menu.
In `focus-click`, we do the same, but add a listener to the `click` event on the trigger node, so when the node already has focus, it can still be toggled by clicking the button.

It might be worth adding a notice at the top of the page that this might not work unless you enable the "Press tab to highlight each option on a webpage" toggle in Safari.

### DIFF
--- a/src/lib/utilities/Popup/popup.ts
+++ b/src/lib/utilities/Popup/popup.ts
@@ -31,7 +31,7 @@ export function popup(node: HTMLElement, args: PopupSettings) {
 	let autoUpdateCleanup: any;
 
 	// Local A11y Variables
-	const elemWhitelist: string = 'a[href], button, input, textarea, select, details, [tabindex]:not([tabindex="-1"])';
+	const elemWhitelist = ':is(a[href], button, input, textarea, select, details, [tabindex]):not([tabindex="-1"])';
 	let activeFocusIdx: number;
 	let focusableElems: HTMLElement[];
 
--- a/src/lib/utilities/Popup/popup.ts
+++ b/src/lib/utilities/Popup/popup.ts
@@ -2,9 +2,22 @@ import { get, writable, type Writable } from 'svelte/store';
 
 // Types
 import type { PopupSettings } from './types';
+import type { Middleware } from '@floating-ui/dom';
 
 // Store
-export const storePopup: Writable<any> = writable(undefined);
+type PopupStore = {
+	computePosition: typeof import('@floating-ui/dom').computePosition;
+	autoUpdate: typeof import('@floating-ui/dom').autoUpdate;
+	flip: typeof import('@floating-ui/dom').flip;
+	shift: typeof import('@floating-ui/dom').shift;
+	offset: typeof import('@floating-ui/dom').offset;
+	arrow: typeof import('@floating-ui/dom').arrow;
+};
+export const storePopup: Writable<PopupStore | undefined> = writable(undefined);
+
+function isNode(el?: unknown): el is Node {
+	return el instanceof Node;
+}
 
 // Action
 export function popup(node: HTMLElement, args: PopupSettings) {
@@ -20,14 +33,16 @@ export function popup(node: HTMLElement, args: PopupSettings) {
 	if (!event || !target) return;
 
 	// Local
-	const { computePosition, autoUpdate, flip, shift, offset, arrow } = get(storePopup);
+	const storeValue = get(storePopup);
+	if (!storeValue) throw Error('You need to initialise the popup store before using the popup action.');
+	const { computePosition, autoUpdate, flip, shift, offset, arrow } = storeValue;
 	const elemPopup: HTMLElement | null = document.querySelector(`[data-popup="${target}"]`);
 	const elemArrow: HTMLElement | null = elemPopup?.querySelector(`.arrow`) ?? null;
-	let isVisible: boolean = false;
-	let autoUpdateCleanup: any;
+	let isVisible = false;
+	let autoUpdateCleanup: () => void;
 
 	// Local A11y Variables
-	const elemWhitelist: string = 'a[href], button, input, textarea, select, details, [tabindex]:not([tabindex="-1"])';
+	const elemWhitelist = 'a[href], button, input, textarea, select, details, [tabindex]:not([tabindex="-1"])';
 	let activeFocusIdx: number;
 	let focusableElems: HTMLElement[];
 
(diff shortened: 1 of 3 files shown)


src/lib/utilities/Popup/popup.ts
```diff
@@ -182,6 +183,7 @@ export function popup(node: HTMLElement, args: PopupSettings) {
 		if (autoUpdateCleanup) autoUpdateCleanup();
 	}
 	function closeOnFocusOut(e: FocusEvent) {
+		console.log('called');
```
endigo9740, line 186: I realize you're still working on this so I thought I'd call this out while I saw it. Careful about leaving in console calls ;)

Cucumberrbob: whoops! Thanks for pointing that out!

Cucumberrbob: that being said, I think everything is done now , I just realised some of my testing around non-interactive elements with tabindexes vs interactive elements was wrong. They're still a little different, in that clicking on a non interactive element with tabindex=0 will open the popup, but a button wouldn't. I think a note in the docs would be better than checking the element type of the node, but I'm open to a better idea

Cucumberrbob:
pushed a couple changes to fix some issues others have found
wasn't really sure where to put them, as only the Combobox change is really related to the `focus` event, but I figured a new PR would just be more effort.

endigo9740:
@Cucumberrbob I noted the additional for a fix in another GitHub issue. Please keep PRs isolated in scope, otherwise it's makes it that much harder to review and understand what's going on.

We have denied PRs in the past for too broad a scope in the past. If you wish to see this merged, I suggest you leave this be and wait for an audit.

Cucumberrbob:
> @Cucumberrbob I noted the additional for a fix in another GitHub issue. Please keep PRs isolated in scope, otherwise it's makes it that much harder to review and understand what's going on.
> 
> We have denied PRs in the past for too broad a scope in the past. If you wish to see this merged, I suggest you leave this be and wait for an audit.

Cool, I'll leave it now as I think it's ready.
Would you like me to remove the unrelated changes and force push them away, and open a new PR with those changes?

endigo9740:
@Cucumberrbob Yeah I'd appreciate you reverting your latest changes and submitting no further PR for now. I'll take your suggestion into account, but I've assigned the issue to myself and would like to review this myself as I was handling the original issue with the poster in Discord.

I appreciate the enthusiasm, but remember you're working with lots of other people here. It's easy to step on toes if you insert yourself everywhere. The appropriate way to help here is to say "hey I have an idea, explain the idea, then make suggest of sending a PR". To which we will either tell you yes or no.

We don't need heroes here, we need need to work together, ok?

endigo9740:
Ok @Cucumberrbob, after reviewing with a couple core team members, here's were I'm going to fall on this PR. First off all, I'm going to strongly encourage you try to keep PRs razor focused on one particular goal at a time. I know it's temping to fix this and that while you have the file open but this has several major downsides in a collaborative environment:

- It provides immediate complexity for the review, making it hard to know the intention of the change. I don't know if a particular change is fixing the issue described in the PR, or something added along the way.
- This makes makes it VERY difficult to for me to review and properly test for regressions. If change A is working as intended, but change B is throwing an error, its not always obvious that change B is the culprit. We might even go so far as to reject a PR due to this issue, even though change A was working as intended the whole time.
- And just for future reference, we provide strict guidelines for how we expect our PRs submitted, which ALWAYS involves prior discussion. We highlight this both in the PR checklist in the canned original message above (which looks to have been nuked), as well as in our [contribution guidelines here](https://www.skeleton.dev/docs/contributing/requirements). This is important because it helps prevent these kind of scope creep issues.

So that said, here's what I'm seeing modified in this PR:

1. Types are being directly import from Floating UI directly into the Popup action
2. You're throwing an error when the Floating UI dependency store is not populated 
3. You've updated and improved typing in various locations throughout the script
4. You've implemented the `focus` and `focus-click` handlers

Of those, only number 4 had prior discussion and is described as the intended goal of this PR. In most cases this would be grounds for me to deny and reject the PR. But given that there's a lot of good work here I'm going to let most of this slide. The only exception being number 1 (type imports).

I know we tested this and were not able to replicate the issues I had with Highlight.js imported straight into Code Blocks. But given this is going to be the first release after dropping v1, where I went on and on about this ushering in stability and no longer including breaking, I feel very uncomfortable sending this out THIS release.

HOWEVER, talking with @AdrianGonz97 I do think this is something that will work fine in production. But I want to test this implemented in isolation away from other changes. If you will please revert those changes and move those to another PR, we'll aim to test this as part of the release after next instead. I think they're good changes, I'd just rather be safe than sorry here.

All that said, there are a number of additional issues I wanted to highlight that will need to be fixed before I can review again and this can be merged:

- You're importing `tick` at the top of the action, though this is not used. (Auto-import accident maybe?)
- Event listeners are being added for `focusin`, `focusout`, and `mousedown`, but not removed in the `destroy()` call. The action always needs to clean up events here.
- Thank you for updating the documentation alongside your changes, but I'm gong to suggest we drop focus/focus-click from the primary demo at the top and and instead create a new section (above Combobox?) dedicated to explaining and showcasing the proper use case of these - say using an input instead of a button. I think they need a little more context.
- The word `initialise` is misspelled in the store error, this should be be `initialize`
- (these are on me) but there's a few misspellings in the comments for `Middlware`, `Visbility`, and `Listners`. If you don't mind helping me grab those too, I would appreciate the help!

Btw if you insist on keeping focus in the primary demo, that's fine, just make sure you account for responsive sizing on small screens while you're at it please!

Screenshot 2023-03-22 at 1 45 12 PM

Cucumberrbob:
Thank you for the detailed response, it was super helpful!
I have:
* removed all the non focus-related stuff
* removed event listeners on destroy
* fixed the spelling issues
* updated docs - removed extra buttons from `sandbox`, and added a `Focus Event` section including two inputs, with `focus` and `focus-click` respectively
I'm not sure I've matched the register the docs normally use, so I'd especially appreciate feedback around the wording I've used.

endigo9740:
Thanks @Cucumberrbob, I'll give this another review tomorrow. If anything needs to be adjusted I'll handle then aim to merge shortly after. Thanks!

endigo9740:
@Cucumberrbob I submitted one final commit cleaning up the docs and implementing a handful of minor changes. Mostly nitpick stuff, nothing to stress over. Review the commit for details.

I'll go ahead and merge this to be included as part of the next release next week!

Cucumberrbob:
Awesome. Thank you so much for your patience. I know this PR kinda sucked in the ways it developed and was submitted, but you were really nice about it, and worked constructively with me to make it better.

PR #1177, opened 2023-03-18
PROPOSAL (DON'T MERGE) - Improved color calibration via Chroma.js

ryceg:
## What does your PR address?

Making a start on the Chroma stuff, simplifies the color logic. 

### Things That Are Working
* Color is generated via the Chroma api 
* You can input a CSS color (aliceblue), hex, rgb etc and it will interpret it correctly and automatically convert to hex
* Text color of the hex value on hover is always accessible

### Things That Aren't Working Quite Right

* Hovering over colors makes them expand, but it is not smooth.
* At smaller viewports there are gap issues on the right hand side while hovering over colors
* You can hover over a color and have it expand, but the mouse must be inside the color to show the hex code

### Out of Scope

* Drag'n'drop rearranging


endigo9740:
@ryceg just checking status on this

endigo9740:
Hey @ryceg, so I've had a chance to run through this. Unfortunately I worried this PR might be a bit overstuffed to do a proper review in time before the monorepo goes into place. The title of the PR denotes the goal was to update to use Chroma.js for color calculation. But browsing through I see:

1. Chroma is implemented as expected, which is great
2. The introduce of the `analogous`, `triadic`, and `tetratic` modes
3. A new way to hover over and reveal the hex color per each shade
4. A means to copy for each shade on click
5. Fairly notable increase in the complexity of the Swatch component logic
6. Changes to the way the CSS snippet is generated

(and probably more I haven't even got to yet)

You pointed out a number of issues in the original post above as well, but my concern here is how dramatically this alters what the Theme Generator is and how it operates. Both from a functionality standpoint and from a UX point of view.

What I'm thinking, and let me know if you agree here, is that we keep the PR in a draft state, but we do not move forward with merging it. Instead, when the monorepo structure is implemented we look to spin the theme generator off to it's own standalone application and having a meeting of the minds to determine the goals of the this tool. Then use this PR as a direct reference for how we should implement there.

Going forward there's a lot of new considerations we need to make for this tool, including but not limited to:

1. Keeping this simple and friendly for your average user (the majority of our audience)
2. Providing a way to toggle on optional power user features (for folks like you and me)
3. Providing support for the expanded 11 shade Tailwind palettes with 950 added
4. Determine how we determine the "main" shade per each color (shade 500)
6. How we generate the palette so the brand color is 1:1 with shade 500 - a hard requirement
7. Improve the UX so this feels much less overwhelming. We're overloading folks here, even power users.
8. Consider alternative ways to display information, such as columns for colors instead of rows
9. Review options for exporting variables in a TS/JS format in addition to CSS
10. Review options for importing and re-editing an existing theme (using the JS file?)
11. Review support for margin and padding CSS properties, and potentially more

(Plus a whole host of other features I haven't even mentioned here)

Given all this, I think a ground up rewrite might be our best solution. If we do that we can plan out the UI and UX before a line of code is written, we can implement Chroma from day one, and we can build this will the full insight of all knowledge already accrued and everything we wish to integrate. All with a clean base that doesn't feel constrained within the documentation site.

When we do this I'd like to be in charge of the interface (UI/UX), and I'd like to task you with the color generation engine. That way we kind of tag team and and lean into our strengths. What do you think?

ryceg:
Hey Chris! Thanks for the review, I am definitely in agreement re: not merging this and instead doing it piecemeal. Some of the changes rely on other changes (hence why the PR got bloated), but I can bash out an order of operations. Re: point 5, I managed to fix that issue, shade 500 is now the inputted colour. 

I am cautious of a ground up rewrite as the color logic in there is all pretty tight (longest function is ~ 20loc), don't wanna throw the baby out with the bathwater, but I really like the idea of the theming interface being a little better defined. I think that a strong vision for what the theming engine should be capable of is needed, at the moment we've (I've) been just slapping features on as they come.

endigo9740:
@AdrianGonz97 @niktek Per the discussion above this will not be moving forward for the monorepo setup. I think with this we're good to move forward with merging that when you two gives the thumbs up.

endigo9740:
@ryceg given this PR won't be moving forward I'm going to go ahead and close it. We can still reference it from the lit of closed PRs though. I don't have an ETA for when I'll return to the standalone theme generator, which will introduce Chroma.js from the ground up. But I'll try to ping you when we get to that point.

PR #1220, opened 2023-03-28
Font Awesome local install

### LINKED ISSUE #1190: Embed Font Awesome locally
### Link to the Page

_No response_

### Describe the Issue

For GDPR compliance we'll need to remove the remote embed of Font Awesome and instead reference this locally. This will also include performance benefits.

### Are you able to create a Pull Request with the fix?

Yes

endigo9740:
Fixes #1190

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,18 +1,19 @@
 {
 	"name": "@skeletonlabs/skeleton",
-	"version": "v1.0.0",
+	"version": "v1.1.0",
 	"lockfileVersion": 2,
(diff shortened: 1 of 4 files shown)


Sarenor:
At least as far as GDPR is concerned that seems fine.
Otherwise I'll need to poke holes in this tomorrow, finally got time for games today :D

Sarenor:
Haven't checked every page on the site, but homepage and a couple of components I've scrolled over look good 👍

PR #1223, opened 2023-03-29
Removed problem causing dot from Font instructions.

Sarenor:
Not an issue but the new Theme-Docs had a small mistake that we all missed:

The font-family src-import needs to be ``src: url('/fonts/${f.file}');`` instead of ``src: url('./fonts/${f.file}');``

### DIFF
--- a/src/app.postcss
+++ b/src/app.postcss
@@ -6,22 +6,22 @@
 /* Google Fonts */
 @font-face {
 	font-family: 'Quicksand';
-	src: url('./fonts/Quicksand.ttf');
+	src: url('/fonts/Quicksand.ttf');
 	font-display: swap;
 }
 @font-face {
 	font-family: 'Space Grotesk';
-	src: url('./fonts/SpaceGrotesk.ttf');
+	src: url('/fonts/SpaceGrotesk.ttf');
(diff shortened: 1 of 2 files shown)


endigo9740:
Every version of the fonts are working with the `./` on the doc site currently:

Screenshot 2023-03-29 at 10 25 29 AM

Maybe both are acceptable?

endigo9740:
@Sarenor this is so weird, I tested multiple conditions and this was the one that worked in the local dev server. We'll need to fix this for the doc site itself too.

Screenshot 2023-03-29 at 10 25 29 AM

endigo9740:
@Sarenor testing the dev site, all custom fonts as shown above are working, minus Quicksand for some reason - which is used for Modern and Gold Nouveau. Additionally Vintage has the wrong font applied, it's using the same one Seafoam uses (Playfair) rather than Abril. Not sure why.

endigo9740:
@Sarenor this PR targets master so I cannot accept it. We need to maintain PR -> dev -> master, even if we have to cherry pick to echo the change from dev to master. I'm going to close this out and recreate it to follow that flow now.

PR #1232, opened 2023-03-30
use namespaced `Skeleton.ModalMeta`

### LINKED ISSUE #1231: Improve types for modal metadata

I've had an interesting idea (may well just be an idea) around the typing of modalStore[number]['meta']. Similar to what Sveltekit does with the App namespace, Skeleton could declare a namespace, and reference types from that namespace. This lets the user override types where applicable. This seems useful as you would now get type errors like with props for passing data to your modal components.

If the user creates a discriminated union type with some key like _type to distinguish between types, then TS can ensure not just that it fits the general type of metadata, but the specific type said component needs:

```ts
// file: skeleton.d.ts in user's repo
type ChangeUsernameModalMeta = {
    currentUsername: string;
    _type: 'change-username';
};
type VerifyEmailModalMeta = {
    email: string;
    _type: 'verify-email';
};
declare namespace Skeleton {
    type ModalMeta = ChangeUsernameModalMeta | VerifyEmailModalMeta;
}
```
and in utilities/Modal/types.ts:
```ts
export interface ModalSettings {
  ...
    meta?: Skeleton.ModalMeta;
}
```

to clarify, this is fully backward compatible since `Skeleton.ModalMeta` will resolve to `any` when not overridden

Enhancement

_No response_

Cucumberrbob:
## Before submitting the PR:

## What does your PR address?

Fixes  #1231
By adding a namespaced type `Skeleton.ModalMeta`, users of Skeleton can overwrite the type to fit their application, similar to SvelteKit's `App.Locals` etc.
This will cause type errors similar to props, and if using typeguards, intellisense for the specific modal component, not just all modal components

### DIFF
--- a/src/lib/ambient.d.ts
+++ b/src/lib/ambient.d.ts
@@ -1,0 +1,3 @@
+declare namespace Skeleton {
+	type ModalMeta = any;
+}
--- a/src/lib/utilities/Modal/types.ts
+++ b/src/lib/utilities/Modal/types.ts
@@ -1,5 +1,5 @@
 // Modal Types
-
+import '../../ambient';
 export interface ModalComponent {
 	/** Import and provide your component reference. */
 	ref: any;
@@ -39,5 +39,5 @@ export interface ModalSettings {
 	/** Override the Submit button label. */
 	buttonTextSubmit?: string;
 	/** Pass arbitrary data per modal instance. */
-	meta?: any;
+	meta?: Skeleton.ModalMeta;
 }


endigo9740:
@ryceg @AdrianGonz97 as our resident TS experts I'd love you guys to give this a once over and let me know what you think.

endigo9740:
@Cucumberrbob it's an interesting idea, but seems a bit like "solution in search of a problem". I think it would be a it of an anti-pattern if folks are sending so much data through the `meta` key that typing would be beneficial. I think when you get to that point that you may have outgrown the intended use for that. For now I think we'll stick to what we have, but thanks for the PR and suggestion!

PR #1242, opened 2023-03-31
bugfix: only dispatch `backdrop` event if backdrop clicked

### LINKED ISSUE #1240: Modal on:backdrop Triggers Inside Modal
### Current Behavior

I'm on Skeleton 1.1.0 and experiencing unexpected behavior with the Modal component. Using the `on:backdrop` event causes the handler to fire whenever any region of the modal is clicked while the modal is open, not just when clicking on the backdrop.

### Steps To Reproduce

1. Create a Modal component in the root `+layout.svelte`.
2. Add the event `on:backdrop`, with any function call. E.g. `on:backdrop={console.log("Hello world."}`
3. Trigger the modal, and click anywhere within its area. Observe that the event is fired within or without the modal's area.

### Anything else?

I believe this to be unintended behavior. I would like to handle clicking outside the modal separately from clicking within it. For instance, if a user clicks outside a modal, this closes it, and I want to react to this event to change the application state.

Cucumberrbob:
## Before submitting the PR:

## What does your PR address?

Fixes #1240 
Before, the `backdrop` event was dispatched whenever an element inside the modal component was clicked.
Now, it only dispatches the event when the backdrop is clicked

### DIFF
--- a/src/lib/utilities/Modal/Modal.svelte
+++ b/src/lib/utilities/Modal/Modal.svelte
@@ -102,10 +102,12 @@
 
 	function onBackdropInteraction(event: MouseEvent | TouchEvent): void {
 		if (!(event.target instanceof Element)) return;
-		if (event.target.classList.contains('modal-backdrop')) onClose();
-		if (event.target.classList.contains('modal-transition')) onClose();
-		/** @event {{ event }} backdrop - Fires on backdrop interaction.  */
-		dispatch('backdrop', event);
+		const classList = event.target.classList;
+		if (classList.contains('modal-backdrop') || classList.contains('modal-transition')) {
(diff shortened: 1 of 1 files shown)


endigo9740:
@Cucumberrbob seems to work well, thanks Bob!

Just FYI I'm going to reach out to the original user requesting this change and ask them their use case. The `on:backdrop` only fires from the `<Modal />` component, which makes it fairly useless in most scenario. Given that this component should only exist in one place - the root layout. I need to ensure they understand they shouldn't be implementing the modal component all over the place. That would be an anti-pattern here.

I believe the Discord discussion is linked in the original GitHub issue if you want to see how that pans out.

PR #1249, opened 2023-03-31
Use `cspell` to fix spelling issues

Cucumberrbob:
## Before submitting the PR:

## What does your PR address?

Fixes hopefully all the errant spelling issues.
There are quite a lot of changes here, so it may take a while to get a manual regression test done to ensure nothing has been broken.

Adds `cspell` as a dependency, but doesn't add a command to `package.json`. Run it using `{npx|pnpm exec} cspell '**/**.*'`

Cspell does have the functionality to allow us to split the ignored words list up into a few files (contributor names, packages Skeleton uses, etc), but this may be more clutter than it's worth


src/routes/(inner)/components/steppers/+page.svelte
niktek, line 1: this shouldn't have wrapped onto new lines- might be worth doing a format on this branch as well to see what happens ?

src/routes/(inner)/components/radio-groups/+page.svelte
niktek, line 1: Absolute nitpicking here - the shortening made both locals the same length, making the shortening horiz is still not spelt right and now of a different length to it's sibling.  I'd probably prefer that it was actually the full name of 
timeHorizontal
timeVertical
As it is consistent with the rest of the codebase.  But it's also a bunch of local vars of no consequence whatsoever.

src/routes/(inner)/elements/breadcrumbs/+page.svelte
niktek, line 1: I'd probably adjust #Each to #each in the h2 as it is quoting a syntax

src/routes/(inner)/elements/cards/+page.svelte
niktek, line 1: Double check the line wrapping here as well

src/routes/(inner)/elements/core/+page.svelte
niktek, line 1: line wrapping

niktek:
Looks good overall.  I'm thinking that it might be better to just add the lorem ipsum vocabulary into the dictionary itself so that a) don't have to worry about escaping it and b) in case the extra escapes are messing with the formatting logic

I think there should also be an update at https://www.skeleton.dev/docs/contributing/requirements that adds the cspell VS code extension as a requirement

Cucumberrbob:
> Looks good overall. I'm thinking that it might be better to just add the lorem ipsum vocabulary into the dictionary itself so that a) don't have to worry about escaping it and b) in case the extra escapes are messing with the formatting logic
> 
> I think there should also be an update at https://www.skeleton.dev/docs/contributing/requirements that adds the cspell VS code extension as a requirement

I was loathed to add the big blocks of `cspell:disable`, but it seemed like the best solution. [Each lorem that emmet generates is different](https://docs.emmet.io/abbreviations/lorem-ipsum/), so my thinking was that demonstrating turning off spell checking there would save time and energy if more sample text ever needs to be added. I'm open to a better solution, but the comments felt cleaner to me than a long word ignore list.

Regarding the requirement of the vscode extensions, I know @endigo9740 has expressed concerns over forcing contributors to use anything except the minimum required tools. Perhaps adding it as a recommended vscode extension would work better?

I'll get the line wrapping fixed now

src/routes/(inner)/components/radio-groups/+page.svelte
Cucumberrbob: `timeHorz` made me think about hours, which is why I changed it rather than adding it to ignore list - which to be fair is outside the scope of this PR.

I've changed them to horizontal/vertical for clarity

Sarenor:
I agree on leaving it as a recommendation. People will learn once their PRs are rejected because of spelling mistakes.

endigo9740:
I'm really not a fan of altering the code source so that a spelling tool doesn't yell at us. We need variable length `lorem ipsum` at different points, so abstracting a variable isn't the right solution here. We don't want to end up with a dozen or so versions or so we have to maintain, and we don't want o have to bounce around the app to simply change text copy.

I think the comments will more than suffice. We get the flexibility of writing whatever we need, and it provides a simple way to disable any warnings. We're already using the comment approach for TS and Prettier throughout the app. It's not an unusual convention for the project.

As for the extension - a friendly recommendation is fine. A hard requirement is not. First it's one requirement, and then another, and so on until people have to install a million different extensions they may not need or want to even contribute. That's a hard pass. We want to keep contributions as accessible as possible. We will never deny a PR because of a spelling mistake either.

endigo9740:
@Cucumberrbob this looks great. There's a lot of great corrections here. It's funny how many of these you can amass in the course of a little over a year!

One thing I'm going to recommend is a follow up PR that documents the presence and usage of cspell, similar to what we do with Prettier and Vitest here:
- https://www.skeleton.dev/docs/contributing/requirements#code-linting--formatting
- https://www.skeleton.dev/docs/contributing/requirements#automated-tests

We don't need any more than what's shown there - a note that it exists and what it does, a link to the documentation, and the primary command(s) needed to run it. Just so we have a quick reference for contributors please.

PR #1271, opened 2023-04-04
Fixed #1264 (SyntaxError in popup close query) with early return

### LINKED ISSUE #1264: Recieve error when using form action to redirect page
### Current Behavior

I am currently using a form hidden within a popup to allow the user to logout. This submits a form action which triggers a redirect. After the user is redirect, the popup no longer exists since you are now on a page without the popup element existing.

### Steps To Reproduce

1. Clone example repo [here](https://github.com/saturnonearth/my-skeleton-demo) (can't be done in Stackblitz as they don't allow form submissions)
2. Open console
3. Click _Dashboard_ in upper right
4. Click, _Click Me!_ button, then click Sign Out
5. See console for error

### Anything else?

Error as shown: 

`DOMException: Failed to execute 'querySelectorAll'(popup.js:105 Uncaught DOMException: Failed to execute 'querySelectorAll' on 'Element': The provided selector is empty.)`

Sarenor:
## What does your PR address?

Fixed #1264 (SyntaxError in popup close query) with early return

### DIFF
--- a/src/lib/utilities/Popup/popup.ts
+++ b/src/lib/utilities/Popup/popup.ts
@@ -94,7 +94,7 @@ export function popup(node: HTMLElement, args: PopupSettings) {
 	// Close popup if element matching closeQuery is clicked
 	function closeOnQueryClick(clickedEl: Node) {
 		if (!clickedEl) return;
-		const interactiveMenuElems = elemPopup?.querySelectorAll(args.closeQuery || '');
+		const interactiveMenuElems = elemPopup?.querySelectorAll(args.closeQuery || 'a[href], button');
 		if (!interactiveMenuElems?.length) return;
 		interactiveMenuElems.forEach((elem) => {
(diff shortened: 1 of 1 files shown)


endigo9740:
@Sarenor without stepping through the history I cannot be sure this was every set properly, but the goal was always to have a default value instead of `''` there. The default for closeQuery should be anchors and buttons, which could be `a[href], button`. We note this in the docs currently:

Screenshot 2023-04-04 at 9 16 48 AM

I know Bob did some work on this, and he may have altered that by mistake. The problem is we want those default, otherwise some of our doc features won't work as expected - such as our app bar menus. Those should close as you select the anchor links within the popup. But currently the menu stays open even through navigation - which is quite annoying.

If you could please revert your change and try the anchor/button query string instead.

endigo9740:
Perfect @Sarenor - that not only resolves the original bug issue, but now our docs are working again. A two-fer! :D

PR #1277, opened 2023-04-05
Feat/migrate twconfig format

### LINKED ISSUE #1273: Tailwind v3.3: Tailwind Config Improvements
- Per: https://github.com/skeletonlabs/skeleton/issues/1238

---

https://tailwindcss.com/blog/tailwindcss-v3-3#esm-and-typescript-support

One of the most error prone parts of using Tailwind is the config file, primarily due to the lack of feedback when a mistake is made in the config file. We should aim to embrace this new format and the benefits it provides asap. This should cut down a slew of support issues on our end.

This ticket will revolve around testing and updating the doc to support this as our official recommenation.

(@niktek I will notify you when we settle on a solution, so that the CLI may follow along)

niktek:
Fixes #1273

Converted tailwind.config to TS and updated the Skeleton TW plugin as well.

- [ ] Need to update manual install docs to match new format and filenames
- [ ] Types could probably be improved generally
- [ ] Line 15 of tailwind/intellisense.ts the typing of the 'all' var needs some typing


src/lib/tailwind/settings.ts
```diff
@@ -1,8 +1,13 @@
...
-module.exports = {
-	colorNames: ['primary', 'secondary', 'tertiary', 'success', 'warning', 'error', 'surface'],
-	colorShades: [50, 100, 200, 300, 400, 500, 600, 700, 800, 900],
+interface ColorPairing {
+	light: number;
+	dark: number;
+}
+
+export default {
+	colorNames: ['primary', 'secondary', 'tertiary', 'success', 'warning', 'error', 'surface'] as const,
```
Sarenor, line 9: I've never seen ``as const`` used in Typescript, what does it do and is it necessary?

Cucumberrbob: [Matt Pocock's video explanation (2 mins)](https://www.youtube.com/watch?v=jxSh-RZTCws)
[Typescript release notes (2 mins)](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions)
[Logrocket blog post (5 mins)](https://blog.logrocket.com/complete-guide-const-assertions-typescript/)

In short, it tells typescript that these values are hard-coded: they will never change. So the `typeof colorNames` is a string literal of `'primary' | 'secondary' | ...`. This gives us intellisense autocompletion for these color names

AdrianGonz97:
Just wanted to list a couple of really important notes:
- We don't really need to change anything to support users using the TW config with TS, they just need to import our plugin like this (we can put this in the documentation): 
```ts
import forms from '@tailwindcss/forms';
import skeleton from '@skeletonlabs/skeleton/tailwind/skeleton.cjs';
import { join } from 'path';
import type { Config } from 'tailwindcss';

export default {
	darkMode: 'class',
	content: [
		'./src/**/*.{html,js,svelte,ts}',
		join(require.resolve('@skeletonlabs/skeleton'), '../**/*.{html,js,svelte,ts}'),
	],
	theme: {
		extend: {},
	},
	plugins: [ ...skeleton({ intellisense: true }) ],
} satisfies Config;
```
- Converting our plugin to TS is going to cause some issues:
    - We need to *ensure* that the transpiled JS for our plugin is formatted as a `.cjs` file, otherwise this is going to be a <ins>**breaking change**</ins>. One way to get around this is to prefix the `.ts` with a `.cts`, telling the compiler to emit a particular file as CommonJS. You can read more about it [here.](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-7.html#new-file-extensions)
    - This is also going to break our build scripts, specifically `compile-css-to-js.cjs` and `generate-jss.js`. Both of them rely on importing the core plugin `/src/lib/tailwind/core.cjs` and our TW config `tailwind.config.cjs`, respectfully. <ins>If our plugin and TW config file are going to be written as TS, we need to convert these build scripts to TS as well.</ins> In addition to that, we're going to need to either add another build step to transpile the build scripts first, or we can use something like [`tsx`](https://www.npmjs.com/package/tsx) to run the scripts directly.

endigo9740:
@niktek @Sarenor @AdrianGonz97 it seems like we've stalled out on this one. Should we call it quits here? Given the number of roadbumps we've encountered I'm ok with returning this a bit further down the road.

Sarenor:
Sounds reasonable, yeah

endigo9740:
Sounds like Dom and I are in favor of pausing this. @niktek do you want to close this out and come back this update at a later time? Give it's your PR I'll let you handle this.

endigo9740:
Gonna make an executive decision here to close this out per the reasons stated above.

PR #1280, opened 2023-04-05
Delete role + tabindex from non-interactive labels

niktek:
This gets rid of the SK warnings and seems to make sense in that the role and tab index were on the label as well as the actual active input.

### DIFF
--- a/src/lib/components/ListBox/ListBoxItem.svelte
+++ b/src/lib/components/ListBox/ListBoxItem.svelte
@@ -71,31 +71,35 @@
 	$: classesLabel = `${cLabel}`;
 </script>
 
-<!-- WARNING: avoid click handlers on <label>; will fire twice -->
-<label
-	class="listbox-item {classesBase}"
-	aria-selected={selected}
-	data-testid="listbox-item"
-	on:keydown={onKeyDown}
-	on:keydown
-	on:keyup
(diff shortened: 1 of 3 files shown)


endigo9740:
I'm not sure this is the right solve for this issue. The issue is the roles don't fit the element, but the roles are important if we're going to call Skeleton an a11y friendly library. Ideally we would change the template to satisfy the needs of both.

See the examples here:
https://www.w3.org/WAI/ARIA/apg/patterns/tabs/

Long story short, those roles are not accepted on Label elements. But we need the Label elements to handle the natural selection. So the simplest solution would be to wrap each tab in a div and move the roles up to that. But keep the labels inside so the click handling works as expected, without us having to reinvent that from the ground up.

I'm fine to take this if need be, just create an issue for it.

endigo9740:
@niktek I've reworked this in a way that should meet all requirements.

- The label is still present for natural selection
- I've inserted a child div that handles the aria roles, tab selection, and keyboard navigation
- I've made a couple minor style changes to ensure focus outlines and such are displayed as expected

It's a bit more verbose than I want, but not sure there's a better way to go about it without dropping the labels and implementing an `on:click` on the parent `div`. I feel the answer is always less JS though, so I went this route.

PR #1295, opened 2023-04-07
Update focusTrap.ts

bartduisters:
onDestory -> onDestroy


endigo9740:
Great catch @bartduisters thanks!

PR #1303, opened 2023-04-09
Add ability to set modal animation duration In/Out seperately

### LINKED ISSUE #1301: Ability to adjust modal animation open/close duration seperatately

Imo, modal animations are good, but when closing, often should be instant, so the user does not have to wait. Since the animation duration is shared this is not currently possible. Could we get a separate prop for open *and* close?

Enhancement

_No response_

saturnonearth:
## Before submitting the PR:
- [x] Did you update and run tests before submission using `npm run test`? (only un-related errors found)

## What does your PR address?

Add the ability to set the modal duration as Number or Array. If set as Array, use the first number as the transition duration In, and the second number as the transition duration Out. If single non-array number is used, use for both In/Out durations.

I added this as an optional Array, rather than splitting the duration prop into two separate Number props, such as durationIn, and durationOut, for a couple of reasons. One being that it would be a breaking change. And the second reason is that I like keeping values that are related to one another, together. It makes more sense and reduces the amount of props you have to pass. I know it's not entirely within the pattern of Skeleton's current prop syntax, so feel free to change it.

Fixes #1301.

### DIFF
--- a/src/lib/utilities/Modal/Modal.svelte
+++ b/src/lib/utilities/Modal/Modal.svelte
@@ -23,7 +23,7 @@
 	// Props (transitions)
 	// Set JSDoc annotation for duration below manually so it's picked up properly by Sveld.
 	/**
- 	* Set the open/close animation duration. Set '0' (zero) to disable, or set as Array to control In/Out duration seperately.
+ 	* Set the open/close animation duration. Set '0' (zero) to disable, or set as Array to control In[0]/Out[1] duration seperately.
 	* @type {number | [number, number]} 
 	*/
 	export let duration: number | [number, number] = 150;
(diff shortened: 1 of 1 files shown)


saturnonearth:
> control In[0]/Out[1] duration seperately.

Done!

endigo9740:
@saturnonearth as noted in the original ticket, I believe there's some overlap with a more sweeping change we're looking to implement in regards to transitions:
https://github.com/skeletonlabs/skeleton/issues/1322

This is heavily inspired by your feedback and PRs. I believe there's some overlap and conflict in the direction of this PR in relation to what's outlined in the ticket linked above. If you agree, then let's close this out in favor of the broader effect incoming.

endigo9740:
Closing this out in favor of: https://github.com/skeletonlabs/skeleton/issues/1322

PR #1309, opened 2023-04-11
Popups display not being correctly set on Close()

### LINKED ISSUE #1305: Using multiple popup setting objects with Autocomplete
### Current Behavior

Using multiple popup setting objects with Autocomplete creates odd behaviors where hitting tab causes input to focus on the div containing autocomplete list, followed by all the items in the list. 

Whereas when multiple autocompletes share the same setting object, and therefor the same name, hitting tab moves to the next input as I would expect it to. 

### Steps To Reproduce

Reproduced using this stackblitz link.

StackBlitz: https://stackblitz.com/edit/skeletonlabs-repl-22yqvw?file=src/routes/+page.svelte

### Anything else?

_No response_

### LINKED ISSUE #1306: Autocomplete: <button> element causes tab to over elements when using.
### Current Behavior

Discovered from #1305 
> Additional issue related to this, Stackblitz should be updated to show a third input field in the first input group, when focused onto the first input and tab is pressed, it skips focusing the second input and moves straight to third.

This happens when using popup, but I would assume popup shouldn't affect this and it would still happen? I can create a separate stackblitz later when I get a chance.

But if I delete the button (during my tests I was deleting it from the else conditional) I was able to tab to the second field, but if I placed the button back in, it would start skipping the second input. 

--- 
## Example

```html
<!-- From my local skeleton testing sandbox -->
<div class="autocomplete {classsesBase}" data-testid="autocomplete">
	{#if optionsFiltered.length > 0}
		<nav class="autocomplete-nav {classesNav}">
			<ul class="autocomplete-list {classesList}">
				{#each optionsFiltered as option, i (option)}
					<li class="autocomplete-item {classesItem}" animate:flip={{ duration }} transition:slide|local={{ duration }}>
						<button class="autocomplete-button {classesButton}" type="button" on:click={() => onSelection(option)} on:click on:keypress>
							{@html option.label}
						</button>
					</li>
				{/each}
			</ul>
		</nav>
	{:else}
        <nav class="autocomplete-nav {classesNav}">
            <ul class="autocomplete-list {classesList}"> 
                <li class="autocomplete-item {classesItem}">
                    <button class="autocomplete-button {classesButton}" type="button">
                        {@html emptyState}
                    </button>
                </li>
            </ul>
        </nav>  
	{/if}
</div>
```

### Steps To Reproduce

Stackblitz: https://stackblitz.com/edit/skeletonlabs-repl-22yqvw?file=src/routes/+page.svelte

### Anything else?

_No response_

JustBarnt:
Fixes #1305
Fixes #1306 

## Before submitting the PR:

## What does your PR address?
`elemPopup.style.display` was being set to `hidden` instead of `none`

This causes an issue when using the autocomplete with popup, because its list items are buttons, and when it wasn't actually being set to `display:none` properly it was causing focus to be given to the button even though the popup wasn't visible to the users.

Updated `elemPopup.style.display` to `none` as `hidden` is a tailwind class name for `display:none`. 

Also updated so that `Tab` and `Escape` close the popup. To match the default functionality for tabbing in an input. 

@endigo9740 ~~If you think its better, I can resubmit this PR with tab only closing the popup is `node` is an instance of `HTMLInputElement` instead?~~ 
Actually after thinking about this for like 10 minutes, I decided also checking if node is an input element is probably better anyways.

### DIFF
--- a/src/lib/utilities/Popup/popup.ts
+++ b/src/lib/utilities/Popup/popup.ts
@@ -158,7 +158,6 @@ export function popup(node: HTMLElement, args: PopupSettings) {
 		render(); // update
 		elemPopup.style.display = 'block';
 		elemPopup.style.opacity = '1';
-		elemPopup.style.pointerEvents = 'initial';
 		isVisible = true;
 		stateEventHandler(true);
 		// Utilize autoUpdate ONLY when the popup is.
@@ -171,7 +170,6 @@ export function popup(node: HTMLElement, args: PopupSettings) {
 		const cssTransitionDuration = parseFloat(window.getComputedStyle(elemPopup).transitionDuration.replace('s', '')) * 1000;
 		setTimeout(() => {
 			elemPopup.style.display = 'none';
-			elemPopup.style.pointerEvents = 'none';
 			isVisible = false;
 			stateEventHandler(false);
 		}, cssTransitionDuration);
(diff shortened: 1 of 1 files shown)


endigo9740:
FYI I've implemented a couple quick changes discussed on Discord. Namely:

1. Removed pointer-event on show/hide, as these are redundant with display none.
2. Adjusted the keyboard interaction if/else logic, simplified this a bit
2. Set a new condition to close the popup if the target node is selected and Tab is pressed

We'll address any other remaining issues in follow up PRs.

PR #1312, opened 2023-04-11
Switch to crossfade instead of fly, increase duration

saturnonearth:
## Before submitting the PR:

## What does your PR address?
I originally wanted to fix the way Toasts would just "popout" when destroyed:
CleanShot 2023-04-11 at 17 52 16

but ended up changing the animation entirely to this:
CleanShot 2023-04-11 at 17 53 05

I added a crossfade animation and increased the default duration of Toasts to 250ms (from 150) to make them appear a bit smoother. This is kind of a preference thing in regards to the animation.

I definitely don't have much experience when it comes to animations so if anyone wants to tweak it so it's more refined, be my guest! I took inspiration from [Svelte French Toast](https://svelte-french-toast.com/) (yes that library name is amazing).

They use straight CSS via:

```CSS
transition: all 230ms cubic-bezier(0.21, 1.02, 0.73, 1);
```
and
```CSS
transform: translateY(calc(var(--offset, 16px) * var(--factor) * 1px));
```

### DIFF
--- a/src/lib/utilities/Toast/Toast.svelte
+++ b/src/lib/utilities/Toast/Toast.svelte
@@ -1,5 +1,6 @@
 <script lang="ts">
-	import { fly } from 'svelte/transition';
+	import { crossfade } from 'svelte/transition';
+	import { cubicInOut} from 'svelte/easing';
 	import { flip } from 'svelte/animate';
 
 	// Types
@@ -16,7 +17,7 @@
 	/** Maximum toasts that can show at once. */
 	export let max = 3;
 	/** The duration of the fly in/out animation. */
-	export let duration = 150;
+	export let duration = 250;
 
 	// Props (styles)
 	/** Provide classes to set the background color. */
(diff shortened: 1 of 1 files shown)


Sarenor:
Tested it in the preview and works as expected while looking beautiful.

PR's looking good as well!

saturnonearth:
> Tested it in the preview and works as expected while looking beautiful.
> 
> PR's looking good as well! (I just don't have the rights to merge it ;) )

Thx! I just released I need to make some slight adjustments and add on a transform origin, will do that when I have time

endigo9740:
@saturnonearth I was kind of curious how this would work with other positions, such as top-left or bottom right, given there's a slight motion to the effect. But I see you're taking the axis values into account here! That's really cool! I love that.

My only question here is whether we think we might reuse this effect in the future. Right now the send/receive are hardcoded directly into the Toast component. Perhaps we should consider abstracting these for potential reuse in the future? My suggestion would be something like:

`/src/lib/transitions/crossfade.ts`

Then move the customized crossfade animation there and allow for import in the Toasts or any other component that could use this. This might align well with our effort in to make animations more configurable:
https://github.com/skeletonlabs/skeleton/issues/1322

Thoughts? Perhaps that's premature abstraction?

saturnonearth:
> @saturnonearth I was kind of curious how this would work with other positions, such as top-left or bottom-right, given there's a slight motion to the effect. But I see you're taking the axis values into account here! That's really cool! I love that.
> 
> My only question here is whether we think we might reuse this effect in the future. Right now the send/receive are hardcoded directly into the Toast component. Perhaps we should consider abstracting these for potential reuse in the future? My suggestion would be something like:
> 
> `/src/lib/transitions/crossfade.ts`
> 
> Then move the customized crossfade animation there and allow for import in the Toasts or any other component that could use this. This might align well with our effort in to make animations more configurable:
> 
> * [Implement dynamic animations for all components #1322](https://github.com/skeletonlabs/skeleton/issues/1322)
> 
> One of my long term goals has always been to introduce a set of custom transitions via Skeleton. As it's the perfect complement to our library.
> 
> Thoughts? Perhaps that's premature abstraction?

Yea I like this idea!

endigo9740:
@saturnonearth I'm going to go ahead and merge this. It's a great change, but I also don't want this blocking or overlapping with @JustBarnt effort on:
https://github.com/skeletonlabs/skeleton/issues/1322

For now we'll keep the customized crossfade localized within component, but once the other changes are complete I would welcome abstracting this out to it's own reusable transition. Perhaps we use that opportunity to introduce a few transition animations to make it a set. This would include a new documentation section, such as Utilities > Transitions that would demo and detail the use of these animations.

So one step at a time :)

PR #1314, opened 2023-04-12
Fix: apply sort to filtered list

ch-lukas:
## What does your PR address?

The Utility DataTable search doesn't work when column sort is applied. To replicate sort any col and then attempt to search. Then to fix it, remove the sort and the search once again works.

Very simple fix was to replace store.base with store.filtered in sortOrder.

### DIFF
--- a/src/lib/utilities/DataTable/DataTable.ts
+++ b/src/lib/utilities/DataTable/DataTable.ts
@@ -80,7 +80,9 @@ export function createDataTableStore<T extends Record<PropertyKey, any>>(source:
 
 /** Listens for changes to `$dataTableModel` and triggers: search, selection, sort, and pagination. */
 export function dataTableHandler<T extends Record<PropertyKey, any>>(model: DataTableModel<T>): void {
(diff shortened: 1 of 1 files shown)


Sarenor:
DataTables aren't in active development at the moment and aren't officially supported.
Just, as a warning to you.

endigo9740:
@ch-lukas I know it sounds crazy to turn down free work, but we're pretty much dead set on retiring the data table component. I've detailed the reasons for this here:
https://github.com/skeletonlabs/skeleton/issues/538#issuecomment-1419517817

I have not had the opportunity to find and vet a proper recommendation yet, but we're leaning towards the use of a third party library if you need extensive use of data tables in your project. Here's one that seems to have a lot of promise and should pair well with Skeleton's Table element styles:
https://github.com/vincjo/datatables

Given this, I'm going to go ahead and close this out. I may look to more forward with deprecating and removing our data table solution in the next week or so. As time allows. So keep an eye on that ticket above.

PR #1334, opened 2023-04-16
Feat/Add onInvalid hook to input chips

Metaphysics0:
## Before submitting the PR:
--> There is currently 1 breaking test that is not related (`autocomplete.ts:134`)
- 
## What does your PR address?

- Adds an optional `onInvalid` hook to the inputChips components + added an example in the docs for this
- Added some typings to the exported
onInvalid-example
 attributes


src/routes/(inner)/components/input-chips/+page.svelte
```diff
@@ -114,6 +124,34 @@
 				</svelte:fragment>
 			</DocsPreview>
 		</section>
+		<section class="space-y-4">
+			<h2>Custom Hooks</h2>
```
Metaphysics0, line 128: since it's only one 'custom hook' for now, i wasn't sure if this title was the best, but i think it's good or if/when we want to add more.

Sarenor:
Nice and good looking addition, thanks for that!
Can be merged as far as I can tell :)

endigo9740:
@Metaphysics0 I really like this feature, the concept is sound. But there were a few issues with the PR as submitted:

1. You were implementing additional types for the input chip props. We're in the process of removing primitive types like this, as Typescript can assume these automatically. No need to explicitly list them. This is the recommended practice. This change was also outside the scope of the PR.
2. You were implementing a custom function to run when the invalidate status is reach, however Svelte has a great mechanism in place for this scenario - using dispatched events! So I've dropped the function in place of a new event
3. Additionally I updated the verbiage and naming for a couple things. This is just me being nitpick-y.

You can review my last commit to see the changes in place. But this event offers a few benefits:

1. It's automatically documented in the `Events` tab on the docs page.
2. It uses the same general language where `onInvalid` becomes `on:invalid`
3. It allows us more easily package up more data in the event details if needed

With these modifications I think this is in a good state, so I'm going to merge now. This will be part of the release next week. Thanks for your contribution!

PR #1342, opened 2023-04-18
Bugfix: Resolved index bug causing first index items to get skipped with `focus` or `focus-click` events.

### LINKED ISSUE #1337: Popup: Hitting the down arrow after display list of items causes it skip the first item
### Current Behavior

When hitting the down arrow to start navigating items in the popup, it automatically jumps to the second item. 

From testing this is only an issue with using `focus` or `focus-click`.

I am fairly positive it is because those two events force focus on the input, and when the input is rendered but its a `focus` or `focus-click` event the focus does not get applied to the starting element, therefor when the down arrow is pressed, index moves to 1 which causes the focus to move to the second item.

This should be easily fixable by moving `activeFocusInx = 0` in the same if conditional that sets the focus immediately, and then updating the `ArrowDown` conditional to also work if index is -1. 

_No response_

### Anything else?

_No response_

JustBarnt:
## Before submitting the PR:
- [X] Does your PR reference an issue? Fixes #1337 

## What does your PR address?
The variable `activeFocusIdx` was being set to 0 no matter what, when it should only be set to 0 if the event is not a `focus` or `focus-click` event just like call to focus the first `focusableElems` is. This caused the first item to get skipped when the down arrow was pressed when using a `focus` or `focus-click` event.

Moved `activeFocusIdx = 0` into the same if that contains `focusableElems[0].focus();`

### DIFF
--- a/src/lib/utilities/Popup/popup.ts
+++ b/src/lib/utilities/Popup/popup.ts
@@ -82,11 +82,11 @@ export function popup(node: HTMLElement, args: PopupSettings) {
 		focusableElems = Array.from(elemPopup?.querySelectorAll(elemWhitelist));
 		// reset the focus index
 		activeFocusIdx = -1;
-		// Automatically focus the element if openWithFocus is true (for example if
-		// the menu was opened with Enter instead of with a click
-		activeFocusIdx = 0;
 		// if the popup was triggered via focus, we don't want to move that focus
+        // we also only want to set activeFocusIdx to 0 if the popup wasn't triggered via `focus` or `focus-click`
 		if (args.event !== 'focus' && args.event !== 'focus-click') {
+            // Automatically focus the element if the event is not `focus` or `focus-click`
+            activeFocusIdx = 0;
 			focusableElems[0]?.focus();
 		}
 	}
--- a/src/lib/utilities/Popup/popup.ts
+++ b/src/lib/utilities/Popup/popup.ts
@@ -83,10 +83,10 @@ export function popup(node: HTMLElement, args: PopupSettings) {
 		// reset the focus index
 		activeFocusIdx = -1;
 		// if the popup was triggered via focus, we don't want to move that focus
-        // we also only want to set activeFocusIdx to 0 if the popup wasn't triggered via `focus` or `focus-click`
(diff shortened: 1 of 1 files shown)


src/lib/utilities/Popup/popup.ts
```diff
@@ -82,11 +82,11 @@ export function popup(node: HTMLElement, args: PopupSettings) {
...
 		// reset the focus index
 		activeFocusIdx = -1;
-		// Automatically focus the element if openWithFocus is true (for example if
-		// the menu was opened with Enter instead of with a click
-		activeFocusIdx = 0;
 		// if the popup was triggered via focus, we don't want to move that focus
+        // we also only want to set activeFocusIdx to 0 if the popup wasn't triggered via `focus` or `focus-click`
 		if (args.event !== 'focus' && args.event !== 'focus-click') {
+            // Automatically focus the element if the event is not `focus` or `focus-click`
+            activeFocusIdx = 0;
```
Sarenor, line 89: If you fix the intendations, I'll even give you a ✅  @JustBarnt  ;)

JustBarnt: I could perhaps do that ;)

JustBarnt: @Sarenor Was it because I use spaces instead of tabs? I just ran prettier nothing looked different so I assume prettier is set to use tabs 🤣

endigo9740:
@JustBarnt @Sarenor I hate to discard completed work, but FYI I've slated an upcoming refactor to this component here:
https://github.com/skeletonlabs/skeleton/issues/1350

Until that's completed I will not be accepting any non-trivial updates or changes to the feature, so I'm going to close this PR rather than moving forward with this. However, I have referenced the original issues and will likely follow your lead here when I get to this point.

But for now, I'd rather tackle all changes "as a whole" rather than piecemeal. If we see any other relevant popup-related changes, please try to list them in the refactor ticket rather than creating new tickets. Any new tickets will be closed and marked duplicate and redirected to there as well.

Thanks!

JustBarnt:
Sounds good!

Issue #1356, opened 2023-04-23
Scrollbar customizations cause bug in Chrome on macOS


michenriksen:
### Current Behavior

The customizations done to the scrollbar in [core.css](https://github.com/skeletonlabs/skeleton/blob/dev/src/lib/styles/core.css#L29-L46) causes the scrollbar “track” to always be visible and take up space from the content, instead of floating on top/overlay the content, inside elements with a defined height and `overflow-y-auto`. 

As an example, I have taken the [sticky position](https://tailwindcss.com/docs/position#sticky-positioning-elements) demo code from Tailwind's documentation and put it in a JSFiddle with Tailwind loaded: https://jsfiddle.net/x2vq8skn/

Notice how the scrollbar floats nicely on top/overlays the content on macOS Chrome:

![vanilla-tailwindcss-scrollbars](https://user-images.githubusercontent.com/304361/233832781-8d277a79-fd8b-48ce-9c71-6e7bf1fad7f5.gif)

If I generate a new Skeleton project and copy-paste the same demo code into `+page.svelte`, the scrollbar area is permanently visible, creating an ugly gap instead of overlaying the content:

![skeleton-scrollbars](https://user-images.githubusercontent.com/304361/233833036-a20933a9-09ea-47cc-98f9-d9e41a27b79b.gif)

### Reproduction / Steps To Reproduce

1. Generate a new Skeleton app with `npm create skeleton-app@latest my-skeleton-app` with default options.
2. Open `src/routes/+page.svelte` and replace content with HTML code from https://jsfiddle.net/x2vq8skn/
3. Run `npm run dev` and open the dev URL in Chrome on macOS. Notice that the scrollbar customizations cause the scrollbar area to always be visible instead of overlaying the content like in the JSFiddle demonstration.

### Anything else?

I am aware that this issue has been [reported and closed as `wont-fix`](https://github.com/skeletonlabs/skeleton/issues/1304); however, I'm opening this issue hoping you will reconsider your `wont-fix` decision. I think it looks awful, and could possibly be a show-stopper for many with the current behavior.

The workaround mentioned in #1304 unfortunately doesn't work. 

I'm sure there is a good reason you introduced the scrollbar customizations, so maybe @hackape's proposal to make it opt-in is a viable solution? If it's possible, having a `.no-scrollbar-customization` to add to elements when needed would also be a nice way to handle it!

hackape:
[...]

PR #1360, opened 2023-04-24
Apply same height to scrollbars as width

bartduisters:
Before change

Screenshot 2023-04-24 at 11 35 49

After change

Screenshot 2023-04-24 at 11 39 03

Note: I've overwritten the behavior of CodeBlock to showcase the before and after - by default there is no overflow on the CodeBlock.

### DIFF
--- a/src/lib/styles/core.css
+++ b/src/lib/styles/core.css
@@ -28,6 +28,7 @@
 
 	::-webkit-scrollbar {
 		@apply w-2;
+		@apply h-2;
 	}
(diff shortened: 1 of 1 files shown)


Sarenor:
Makes sense and looks good to me!

endigo9740:
@bartduisters thanks for this change, note that we may be revising the scrollbar style per this request soon:
https://github.com/skeletonlabs/skeleton/issues/1356

However, I'll go ahead and accept this for now as I do feel it's an improvement. Thanks!

Issue #1286, opened 2023-04-05
Investigate if moving CSS-Imports has any detrimental side effects


Sarenor:
### Describe what feature you'd like. Pseudo-code, mockups, or screenshots of similar solutions are encouraged!

We've found that Tailwind behaves differently when ``skeleton.css`` etc. are imported in ``app.postcss`` instead of the root ``+layout.svelte``.
So far, we don't officially recommend moving the imports there since we fear unknown side effects and things not working as expected with the way Tailwind behaves.

On the other hand, potentially having layer-directives & @apply available for Skeleton & End user classes could be a very valuable enhancement so we should devote some time to investigation. 

### What type of pull request would this be?

Enhancement

### Any links to similar examples or other references we should review?

https://discord.com/channels/1003691521280856084/1093117288875569192

MLNW:
Relates to https://github.com/skeletonlabs/skeleton/issues/1152

endigo9740:
@MLNW just to make sure we're on the same page - the relation is that by moving the CSS imports into the stylesheet, that folks are free to use the automatic import reordering feature of VS code without issue. Is that correct?

If so I agree with that take. And we should weight that with considering this change.

MLNW:
Yes, that's correct

PR #1361, opened 2023-04-24
rename docSell to docShell

bartduisters:
To have more clarity `docSellDefaults` was renamed to `docShellDefaults`. The placement swap of the imports is due to import sorting. Can be reverted if not wanted.

I do prefer automatic import sorting on projects with multiple contributors to have consistent import orders.

### DIFF
--- a/src/docs/layouts/DocsShell/DocsShell.svelte
+++ b/src/docs/layouts/DocsShell/DocsShell.svelte
@@ -11,13 +11,13 @@
 	// Partials
 	import Header from '$docs/layouts/DocsShell/partials/Header.svelte';
 	import PanelProps from '$docs/layouts/DocsShell/partials/PanelProps.svelte';
-	import PanelParams from './partials/PanelParams.svelte';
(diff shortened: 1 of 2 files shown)


endigo9740:
The base change is fine, but...

> The placement swap of the imports is due to import sorting. Can be reverted if not wanted.
> I do prefer automatic import sorting on projects with multiple contributors to have consistent import orders.

We don't support this right now. It's not a matter of preference, this causes detrimental behavior where stylesheet imports can be sorted in an incorrect order. Order matters for our stylesheet imports. We're actually adding a new FAQ (last item) that details this here:
https://dev.skeleton.dev/docs/introduction#faqs

We believe if we move our CSS imports to occur within the global stylesheet, this might avoid behavior per:
https://github.com/skeletonlabs/skeleton/issues/1286

But currently this is untested. We may need to deny PRs that change import order going forward until this situation is sorted.  At any for any changes to the root layout. Please use discretion when utilizing the reordering please.

Generally speaking only change the item described in the PR. Anything else it out of scope and can be harder to review.

PR #1362, opened 2023-04-24
Remove inferred JSDoc type annotation.

juangiordana:
Remove inferred JSDoc type annotation.


endigo9740:
@juangiordana I'm sure you mean well, and you're technically correct this can be inferred for type checking. However, this will break out auto-documentation of the prop when using Sveld. You can learn more about this here:

https://www.skeleton.dev/docs/contributing/documentation

Essentially non-primitive types are not inferred by Sveld, which means they would not be documented properly on the "Props" tab on the documentation page. It's a quirk of Sveld.

juangiordana:
Sorry about that. I'll make sure to review the docs next time :)

Issue #1189, opened 2023-03-22
Opt-In Typography Styles


endigo9740:
### Describe what feature you'd like. Pseudo-code, mockups, or screenshots of similar solutions are encouraged!

We've had a lot of folks struggling with Typography styles overrides, either locally or globally. Let's look into a means make it easier to opt-in. We found a lot of success with inputs, so let's follow the lead there.

### What type of pull request would this be?

Enhancement

endigo9740:
Noting this Tailwind v3.3 feature for later reference:
https://tailwindcss.com/blog/tailwindcss-v3-3#new-line-height-shorthand-for-font-size-utilities

endigo9740:
Note: font sizes should be built into theme settings, design tokens added

endigo9740:
FYI I've started a new PR here: https://github.com/skeletonlabs/skeleton/pull/1374

I've made a few modifications so far:

1. I fixed some HTML elements within the "Explore" drop down in the app bar (unrelated)
2. Implemented the Tailwind Typography plugin for use within our blog
3. Update our blog to utilize the `.prose` class by default instead of custom styles
4. Reimplemented the way we use `.prose` styles in `typography.css`. This was all setup incorrectly before.
5. Simplified many of the existing typography selectors

Per number 5, here's a before and after:

```css
h1:not(.unstyled):is(:not(.prose *)) { ... }
```

```css
h1:not(.unstyled) { ... }
```

Note that current updates do not address the primary issue yet - typography is not yet opt in. But I'll return to this tomorrow. I'm testing several ideas, with the goal minimizing regressions for existing users. If there's a migration process, I'll keep it as simple as possible.

endigo9740:
The new opt-in typography system is now available in the PR. Note that I've only generated the appropriate stylesheets and documented how it operates, and how to migrate. I have not yet migrated the documentation site to utilize the new system. View the notes below to learn how this works.

## Stylesheet Updates

A number of updates have been made to stylesheets:

- A lot of loose partial stylesheets have (ex: tailwind.css, core.css, etc) have been moved to `/styles/partials`
- The `all.css` stylesheet remains, but now imports `skeleton.css` - it's future successor
- A new `skeleton-next.css` stylesheet has been introduced, this mirrors `skeleton.css` but uses the opt-in typography
- A new `skeleton-minimal.css` stylesheet has been introduced which excludes `core.css` and `typography-x.css`

[...]

PR #1374, opened 2023-04-25
Improve Typography System

endigo9740:
Addresses #1189

As well as:

Addresses #1376
Addresses #1356


src/routes/(inner)/components/autocomplete/+page.svelte
```diff
@@ -222,9 +222,9 @@ const flavorOptions: AutocompleteOption[] = [
 		<section class="space-y-4">
 			<h2>Input Chip</h2>
 			<p>We've provided a demo of using Autocomplete alongside a Skeleton Input Chip component below.</p>
-			<pre>inputChip: {JSON.stringify(inputChip, null, 2)}</pre>
-			<pre>flavorOptions: {JSON.stringify(flavorOptions, null, 2)}</pre>
-			<pre>inputChipList: {JSON.stringify(inputChipList, null, 2)}</pre>
+			<!-- <pre>inputChip: {JSON.stringify(inputChip, null, 2)}</pre> -->
```
bartduisters, line 225: Is this on accident?
If they are no longer necessary, could be removed

src/routes/+layout.svelte
```diff
@@ -2,7 +2,7 @@
 <script lang="ts">
 	// Dependency: Highlight JS
 	import hljs from 'highlight.js';
-	import '$lib/styles/highlight-js.css'; // was: 'highlight.js/styles/github-dark.css';
+	import '$docs/styles/highlight-js.css'; // was: 'highlight.js/styles/github-dark.css';
```
bartduisters, line 5: The comment should now be `// was: '$lib/styles/highlight-js.css'`, or be removed - is it important what it used to be?

src/routes/(inner)/components/autocomplete/+page.svelte
endigo9740: I need to check the git blame on this, but someone left these in and it made it all the way to production darn it. You're right though, should just be axed. Handling that now.

src/routes/+layout.svelte
endigo9740: The comment isn't really useful any more so I'm dropping it completely.

endigo9740:
@bartduisters thanks, I've implemented changes based on your suggestions above.

PR #1421, opened 2023-05-02
Popup refactor

### LINKED ISSUE #1282: Popup does not work after page change when on AppBar
### Current Behavior

Popup behavior is odd when placed onto AppShell/AppBar. It stops working when either there is a hot-reload on dev mode, or when you change the page.

You can see the button on each page that is outside the AppBar always works.

chrome_wbHapXbohs

### Steps To Reproduce

1. Load Stackblitz link
2. Click "Authenticated" in top nav bar
5. Click button on top right of AppBar - it should "flash".

### Anything else?

Note that each page has its own layout with its own AppShell. This issue may be related to #1281

StackBlitz link: https://stackblitz.com/edit/skeletonlabs-repl-8scgxj?file=src/routes/(public)/+layout.svelte

### LINKED ISSUE #1336: Popup: Allow passing a reference to an element instead of a string id

At the moment we need to specify a target that matches a `data-popup` attribute:
```
let popupSettings: PopupSettings = {
	// Set the event as: click | hover | hover-click | focus | focus-click
	event: 'click',
	// Provide a matching 'data-popup' value.
	target: 'examplePopup'
};
<button ... use:popup={popupSettings}>Trigger</button>
<div ... data-popup="examplePopup">(popup)</div>
```

As far as I can tell we are then getting a reference to the element as essentially the first thing in the popup action
```
	const elemPopup: HTMLElement | null = document.querySelector(`[data-popup="${args.target}"]`);
```

For my use-case (a drop down menu I want to be able to use in different places), it would be far more convenient if I could simply pass the reference directly. 

It could look something like the following. Notice that I've renamed `target` to `targetRef` so both options are still possible.
```
let popup;
let popupSettings: PopupSettings = {
	// Set the event as: click | hover | hover-click | focus | focus-click
	event: 'click',
	// Provide a matching 'data-popup' value.
	targetRef: popup
};
<button ... use:popup={popupSettings}>Trigger</button>
<div bind:this={popup}>(popup)</div>
```

Unless I'm missing something, it is probably a one or two line change to make that functionality available.

Enhancement

_No response_

### LINKED ISSUE #1337: Popup: Hitting the down arrow after display list of items causes it skip the first item

### LINKED ISSUE #1338: Popup hover: target popup disappears when mouseout on parent element
### Current Behavior

Popup disappears when hovering on it. I think there maybe an option to showing target popup when hovering over it. As an example implementation: change `mouseout` to `mouseleave` and hide only after some delay (300-500 ms should be fine I think)

_No response_

### Anything else?

_No response_

### LINKED ISSUE #1345: Differentiate popup navigation list focus and selection styles
### Current Behavior

Hello!

I'm trying to highlight the nav list item which corresponds to the current page in my AppBar navigation lists.

I'm using the same code as sketelon.dev, which works well for highlighting the current page on SideBar and Drawer.

For AppBar nav lists, I'm using something similar, as suggested in the docs: "To highlight active state, we recommend conditionally applying a background color to the anchor tag" (the documentation should be updated to use "bg-primary-active-token" instead of "!bg-primary-500" . It gives better results in dark mode.

```
$: classesActive = (href: string) => {
		return $page.url.pathname === href ? 'bg-primary-active-token' : '';
	}
```

The current page is highlighted. The bug is that the first item of each navlist always remains highlighted. The same thing happens in skeleton.dev (first item of "Explore" menu is always highlighted: "Home").

Interestingly, this does not happen on the Theme menu, where only the current theme is highlighted.

If I remove ```class="list-nav"``` from the nav element, highlighting works correctly but styling is lost. :(

Thank you! :)

_No response_

### Anything else?

_No response_

### LINKED ISSUE #1350: Popup refactor and extension
We've had a number of issues and feature requests reported for the current popup implementation. Likewise we've had a bit of "feature creep" since the inception. Given this I'm going to aim for a refactor of this component utilizing all information we have available to us for this component.

This may include a number of new feature requests:

- [x] https://github.com/skeletonlabs/skeleton/issues/1014
- [x] Make it easier to create tooltips (reference [#1332](https://github.com/skeletonlabs/skeleton/pull/1332))
- [x] https://github.com/skeletonlabs/skeleton/issues/1338
- [x] https://github.com/skeletonlabs/skeleton/issues/1336

I'll also aim to address a number of reported bugs:

- [x] https://github.com/skeletonlabs/skeleton/issues/1345
- [x] https://github.com/skeletonlabs/skeleton/issues/1282
- [x] https://github.com/skeletonlabs/skeleton/issues/1337
- [x] https://github.com/skeletonlabs/skeleton/issues/1379 ([potential PR fix](https://github.com/skeletonlabs/skeleton/pull/1415))

New issues reported in the thread below:

- [x] Make considerations for popups used within a loop ([reference](https://discord.com/channels/1003691521280856084/1048010639219642419/threads/1099704955478212628))
- [x] consider a way to support multiple popups on one trigger - ex: tooltip + popup ([reference](https://discord.com/channels/1003691521280856084/1100073004815351939))
- [x] The closeQuery: `''` selector should default all child element closing the popup
- [x] Hover popups flicker when using an icon inside a button instead of text (CHRIS: `btn-icon` works best)
- [x] Document: The target in popup settings must be unique, otherwise this may lead to undesired behavior

If you have new ideas or suggestions related to the popup I'm going to ask that you focus them here for now. If the issues is large enough to require a dedicated ticket I'll split it off as need. But I think we need a "wholistic" approach to these updates to ensure everything continues to work in unison as expected.

---

Not implemented / Not supported

> NOTE: I'll move items down here that we don't intend to move forward with or cannot support. I'll response in each ticket with a detailed follow up for this designation 

- [x] https://github.com/skeletonlabs/skeleton/issues/1327
- [x] https://github.com/skeletonlabs/skeleton/issues/1323

### LINKED ISSUE #1379: Popup is not working fine on live as in dev env.
### Current Behavior

Popup is working fine on dev env but when it comes in live, that popup disappears when I hover and move around on trigger div.
And I'm using popup inside a loop, here is the main code

```html
<script lang="ts">
	// skeleton and floating-ui
	import { popup } from '@skeletonlabs/skeleton';
	import type { PopupSettings } from '@skeletonlabs/skeleton';
	import { computePosition, autoUpdate, flip, shift, offset, arrow } from '@floating-ui/dom';
	import { storePopup } from '@skeletonlabs/skeleton';

	storePopup.set({ computePosition, autoUpdate, flip, shift, offset, arrow });

	let popupSettings: PopupSettings = {
		event: 'hover', // event
		target: 'my_list_popup' // data-popup value
	};
</script>

<div>
	<div>
		{#each my_list as anime}
			<div
				use:popup={popupSettings}
			>
				<div>
                                        <!-- trigger code -->
                                </div>
			</div>
                        <div
				data-popup="my_list_popup"
			>
				<!-- popup code -->
				<div class="arrow bg-surface-50" />
			</div>
		{/each}
	</div>
</div>
```

_No response_

### Anything else?

[Screencast from 2023-04-26 20-34-04.webm](https://user-images.githubusercontent.com/114811070/234635596-3531c646-bd9f-4cae-b647-275d3fc63cbf.webm)

1st one is the live site and 2nd one is the dev env ( localhost )
The problem is with the 1st one ( live )

endigo9740:
Closes #1350

As well as:

Fixes #1338
Fixes #1336
Fixes #1345 
Fixes #1282
Fixes #1337
Fixes #1379

### DIFF
--- a/src/lib/utilities/Popup/popup.ts
+++ b/src/lib/utilities/Popup/popup.ts
@@ -16,8 +16,14 @@ export function popup(triggerNode: HTMLElement, args: PopupSettings) {
 	let focusablePoupupElements: HTMLElement[];
 	const documentationLink = 'https://www.skeleton.dev/utilities/popups';
 	// Elements
-	const elemPopup: HTMLElement | null = document.querySelector(`[data-popup="${args.target}"]`);
-	const elemArrow: HTMLElement | null = elemPopup?.querySelector(`.arrow`) ?? null;
+	let elemPopup: HTMLElement;
+	let elemArrow: HTMLElement;
+
+	function setDomElements(): void {
+		elemPopup = document.querySelector(`[data-popup="${args.target}"]`) ?? new HTMLElement();
+		elemArrow = elemPopup?.querySelector(`.arrow`) ?? new HTMLElement();
+	}
+	setDomElements(); // init
 
 	// Render Floating UI Popup
 	function render(): void {
@@ -104,7 +110,7 @@ export function popup(triggerNode: HTMLElement, args: PopupSettings) {
 		// Focus the first focusable element within the popup
 		focusablePoupupElements = Array.from(elemPopup?.querySelectorAll(focusableAllowedList));
 	}
-	function close(): void {
+	function close(callback?: any): void {
 		if (!elemPopup) return;
 		// Set transition duration
 		const cssTransitionDuration = parseFloat(window.getComputedStyle(elemPopup).transitionDuration.replace('s', '')) * 1000;
@@ -116,9 +122,11 @@ export function popup(triggerNode: HTMLElement, args: PopupSettings) {
 			// Update the DOM
 			elemPopup.style.opacity = '0';
 			elemPopup.style.pointerEvents = 'none';
+			// Cleanup Floating UI autoUpdate (close only)
+			if (popupState.autoUpdateCleanup) popupState.autoUpdateCleanup();
+			// Trigger callback
+			if (callback) callback();
 		}, cssTransitionDuration);
-		// Cleanup Floating UI autoUpdate (close only)
-		if (popupState.autoUpdateCleanup) popupState.autoUpdateCleanup();
 	}
 
 	// Event Handlers
@@ -192,8 +200,11 @@ export function popup(triggerNode: HTMLElement, args: PopupSettings) {
 	// Lifecycle
 	return {
 		update(newArgs: PopupSettings) {
-			args = newArgs;
-			render();
+			close(() => {
+				args = newArgs;
+				render();
+				setDomElements();
+			});
 		},
 		destroy() {
 			// Trigger Events
--- a/src/lib/utilities/Popup/popup.ts
+++ b/src/lib/utilities/Popup/popup.ts
@@ -148,27 +148,19 @@ export function popup(triggerNode: HTMLElement, args: PopupSettings) {
 		if (popupState.open === false) return;
 		// Handle keys
 		const key: string = event.key;
-		// On ESC key
+		// On Esc key
 		if (key === 'Escape') {
 			event.preventDefault();
 			triggerNode.focus();
 			close();
 			return;
 		}
-		// On TAB key
-		if (document.activeElement === triggerNode && key === 'Tab') {
-			if (args.event === 'focus-blur') return;
-			event.preventDefault();
-			triggerNode.focus();
-			close();
-			return;
-		}
-		// On Arrow Down key
-		if (popupState.open && document.activeElement === triggerNode && key === 'ArrowDown') {
+		// On Tab or ArrowDown key
+		const triggerMenuFocused: boolean = popupState.open && document.activeElement === triggerNode;
+		if (triggerMenuFocused && (key === 'ArrowDown' || key === 'Tab')) {
 			event.preventDefault();
 			if (focusableAllowedList.length > 0) focusablePoupupElements[0].focus();
 		}
-		// TODO: On Arrow Up key
 	};
 
 	// Event Listeners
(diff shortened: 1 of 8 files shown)


src/lib/utilities/Popup/popup.ts
```diff
@@ -70,205 +66,131 @@ export function popup(node: HTMLElement, args: PopupSettings) {
...
-		node.addEventListener('mouseover', show, true);
-		node.addEventListener('mouseout', close, true);
+	switch (args.event) {
+		case 'click':
+			triggerNode.addEventListener('click', toggle, true);
+			window.addEventListener('click', onWindowClick, true);
+			break;
+		case 'hover':
+			triggerNode.addEventListener('mouseover', open, true);
+			triggerNode.addEventListener('mouseleave', close, true);
```
Sarenor, line 160: 👍  :D Thanks! :)

endigo9740: I gotchu :D

endigo9740:
I've done one more pass through the documentation to improve the instruction. I'm going to go ahead and merge this, but welcome continued testing through the remainder of this week.

PR #1436, opened 2023-05-05
Draft suggestions for additional popup improvements

danielimmke:
…definition for popup settings

## What does your PR address?

This is some work I did to get tooltips working more smoothly on my own project today. Coincidentally, the popover file on the dev branch was worked on today. I decided to pull latest and submit my changes as a speculative PR. The main two issues I was trying to fix were: needing to add a direction class to the container so I could style the "arrow" element to have a border (box shadow) and there not being a fade in transition to go with the fade out transition.

Additionally, I noticed an issue where when you had an item set to "Hover" it was spam invoking the open/close methods so I added an extra guard to stop that.

I definitely don't expect this to get merged as is, but I figured it might be worth looking at.

[Demo](https://user-images.githubusercontent.com/1744366/236364105-2d8e8830-748a-4e81-8a6f-c03fc6f5d23c.mov)

### DIFF
--- a/src/lib/utilities/Popup/popup.ts
+++ b/src/lib/utilities/Popup/popup.ts
@@ -1,5 +1,6 @@
 import { get, writable, type Writable } from 'svelte/store';
-import type { Middleware, PopupSettings } from './types';
+import type { Placement } from '@floating-ui/dom';
+import type { PopupSettings } from './types';
 
 // Use a store to pass the Floating UI import references
 export const storePopup: Writable<any> = writable(undefined);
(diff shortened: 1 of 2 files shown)


endigo9740:
Hey @danielimmke, yeah I've been busy working on bit refactor this week per:
- https://github.com/skeletonlabs/skeleton/issues/1350

This addresses a number of issues reported by users, with a few feature expansions, as well as much improved documentation.

With a quick glance of your PR I am seeing a couple issues that will prevent us from merging as is, but I am interested to see how you implemented these. The border styling for arrows seems cool - I'm wanting implement a means to allow the Svelte action to generate the arrow element dynamically via a parameter fan. Probably default on with `arrow: false` to disable.

> Additionally, I noticed an issue where when you had an item set to "Hover" it was spam invoking the open/close methods so I added an extra guard to stop that.

I'm very curious about this point. This hasn't come up previously and I haven't observed this, so I'll look a bit further into how you resolved this.

It's late evening for me, so I may not get to a proper review until tomorrow, but I do look forward to checking it out. Thanks for the submission!

src/lib/utilities/Popup/popup.ts
```diff
@@ -1,5 +1,6 @@
 import { get, writable, type Writable } from 'svelte/store';
-import type { Middleware, PopupSettings } from './types';
+import type { Placement } from '@floating-ui/dom';
```
endigo9740, line 2: Real quick though, here's one of the issues that's a no-go for us. We cannot depend on the dependencies being installed by default. We very much want to make all third party dependencies like this (highlight.js, etc) optional and opt in.

By embedding this import inside the `popup.ts` it's no longer optional and will throw errors for users that install Skeleton without this present, even if they're not using popups. We ran into this very same issue with Highlight.js in the past.

endigo9740:
Reiterating what I mentioned above - we can't use this in it's current state given the way imports are being handled. But I will keep this in mind for future expansion. Thanks!

PR #1437, opened 2023-05-05
Feature | Added LineNumbers Property to Code Block

### LINKED ISSUE #1241: New CodeBlock prop: lineNumber

The default would be `0` (hide).
If specified, such as `42`, each line in the code block would be numbered sequentially on the left, starting with the specified value.

This is two features in one:
- The ability to add line numbers in the code block
- The ability to customize the _starting_ line number.

This would be a standout feature, since most implementations of line numbering has no option to adjust the starting number.

New Feature

_No response_

LukeHagar:
Closes #1241


endigo9740:
@LukeHagar I made a number of changes:

1. Moved the CSS to Element styles, to match conventions with the rest of the library
2. Improved the line number spans to have a fixed width. Otherwise each digit would shift things to the right by that amount. (ex: 10, 100, 1000 would be uneven)
3. Adjusted the demo text just slightly and implemented the opt-in typography
4. Dropped a `const` no longer used in the codeblock logic

Per the typography, this is a new addition this release. Be sure you're doing your PRs against the lastest dev updates, it seemed you were a bit behind so the title and code tag showed like this:

Screenshot 2023-05-05 at 12 38 10 PM

Issue #1459, opened 2023-05-08
Horizontal AppRail


austerj:
### Describe what feature you'd like. Pseudo-code, mockups, or screenshots of similar solutions are encouraged!

Hi, thanks for making Skeleton!

I'd like to make use of the `AppRail` / `AppRailTile` components as a horizontal navbar on mobile. I'm currently tinkering with a local version of the component and will make a pull request if I end up at something that seems appropriate. Also mentioned at https://github.com/skeletonlabs/skeleton/issues/790, however as pointed out the tabs have the disadvantage of not being proper anchors and somewhat unnecessary bindings for the purpose of navigation, while the `AppRail` makes nice use of `svelte:element` to dynamically create the appropriate HTML element.

Currently leaning towards making a dedicated horizontal component for a couple reasons. Since the default props are provided as strings (instead of e.g. `height ?? 'h-full'`), there wouldn't be a way to provide sensible defaults dispatched on something like a boolean `isHorizontal` flag without breaking with the conventions. I also don't think there's a ton of value gained by trying to force this into one component when weighed against the extra complexity of branching default props.

On a related note, the `AppRailTile` forces a square aspect ratio, while a bottom navbar should probably default to evenly distributing the full screen width when possible. I would argue this could be a nice generalization for the standard `AppRailTile` component as well, as I don't see any particular reason why the square aspect ratio should be mandatory on desktop either. Ideally such a generalization could be made and the same component type could be slotted into both the vertical- and horizontal `AppRail`.

### What type of pull request would this be?

New Feature

PR #1460, opened 2023-05-08
Horizontal AppRail

austerj:
## Before submitting the PR:

## What does your PR address?

Addresses feature request in https://github.com/skeletonlabs/skeleton/issues/1459, starting with enabling non-square `AppRailTiles`.

### DIFF
--- a/src/lib/components/AppRail/AppRail.svelte
+++ b/src/lib/components/AppRail/AppRail.svelte
@@ -58,6 +58,6 @@
 	<div class="app-bar-lead {regionLead}"><slot name="lead" /></div>
 	<!-- Slot: Default -->
 	<div class="app-bar-default {regionDefault}"><slot /></div>
-	<!-- Slot: lead -->
+	<!-- Slot: trail -->
 	<div class="app-bar-trail {regionTrail}"><slot name="trail" /></div>
 </div>
--- a/src/lib/components/AppRail/AppRailTile.svelte
+++ b/src/lib/components/AppRail/AppRailTile.svelte
@@ -55,8 +55,9 @@
 	}
 
 	// Reactive
+	$: classesDiv = `${width} ${height}`;
 	$: classesActive = String($selected) && String(value) && $selected === value ? `${active}` : '';
-	$: classesBase = `${cBase} ${width} ${height} ${hover} ${classesActive} ${$$props.class || ''}`;
+	$: classesBase = `${cBase} ${hover} ${classesActive} ${$$props.class || ''}`;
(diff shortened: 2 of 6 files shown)


austerj:
I noticed some apparent inconsistencies in the naming in the existing `AppRail` component (`regionLead`, `regionDefault` and `regionTrail` should be `slotLead`, `slotDefault` and `slotTrail` right?), but that would be a breaking change in case someone uses those props as they're named now. So was not sure if this should be left as is.

austerj:
https://stackblitz.com/edit/skeletonlabs-repl-vrvxsv?file=src%2Froutes%2F%2Blayout.svelte

Small demo of the updated `AppRailTile` with the existing `AppRail` and demoing the added `HorizontalAppRail`. It should be pretty easy to make a responsive layout from this with media queries.

The most drastic change I made was in the `AppRailTile`, where I added `width` and `height` parameters. Because there's an outer `div` handling events, I found that passing the `width` and `height` directly to this `div` instead and just always filling out that with the inner element gave the best results, and as far as I can tell it behaves similar to the original one with the default props, but with the added benefit that you can do things like this:

image

endigo9740:
@austerj I appreciate your enthusiasm and the PRs, but there's a few problems here. The main issue is you've not followed our contribution instructions. We explicitly request prior discussion for non-trivial issues. And we prefer this before a line of code is written. This appears in two places:

1. In the original post above, which you checked as completed. Even though this did not occur:
Screenshot 2023-05-09 at 10 44 40 AM

2. In our contribution guidelines:
https://www.skeleton.dev/docs/contributing/requirements

Screenshot 2023-05-09 at 10 46 02 AM

We try to make this very clear so that folks don't spend time working on issues we can't accept. Unfortunately that seems to be the scenario that's played out there. You've spent time building this new feature that we won't be able to move forward with for the reasons stated below:

1. We would never recommend duplicating components for an orientation change within a component. This should be an extension to the current component to keep maintenance low. Doubling components doubles the maintenance cost.
2. The app rail is a "rail", not a "bar", which means it it is not intended for a horizontal layout. We've borrowed this concept from Material Design and they state the guidelines for this very well:

https://m3.material.io/components/navigation-rail/overview
> There's one type of navigation rail in Material. This component works best on larger screens and can be swapped with a navigation bar on smaller screens.

3. We already have a solution for a "navigation navigation" bar, [which is provided using styled](https://www.skeleton.dev/components/tabs#navigation).

Again, these are all things that either myself or another member of the team could have explained if you created an issue first, reach out on Discord, or otherwise conversed with us upfront. Again I'm sorry but we cannot accept this.

PR #1482, opened 2023-05-11
added mutationObserver to focusTrap

### LINKED ISSUE #1423: Accordions in Drawers (or in general?) have wonky Tab shortcuts
### Current Behavior

When inside of a Drawer, attempting to navigate down an AccordionsItems ends up becoming impossible **unless** they were open already. However, when you close the Accordion in this scenario (with the accordion already open prior to tabbing through it), you are able to tab outside of the drawer.

https://user-images.githubusercontent.com/32039708/236020851-c7b51531-f146-4217-b3fd-c817e8eef1c8.mp4

https://stackblitz.com/edit/skeletonlabs-repl-kh4qc3?file=src/routes/+layout.svelte

### Anything else?

Discord Thread: https://discord.com/channels/1003691521280856084/1103386196350095400/1103386196350095400

### LINKED ISSUE #1444: The search for the Skeleton website loses focus trap after searching anything.
### Current Behavior

After searching anything without going somewhere you are able to tab out of the search modal into the backdrop and tab through the site.

1. Go to https://www.skeleton.dev/
2. Hit CMD/CTRL + K
3. Type "tables" for example
4. Hit tab until you tab outside of the modal into the actual site whilst being in the search modal

### Anything else?

It's interesting that if you only type 1 letter, then hit backspace it still allow you to get out so I feel like it has to do with how autocomplete filters

Mahmoud-zino:
## Fixes

Fixes #1444
Fixes #1423 

## Before submitting the PR:
I have also tested the focusTrap manually with the following components:
1. DocsSearch
2. Modal
3. Accordion
4. Drawer
5. Drawer with Accordions with anchor links

## What does your PR address?
As described here: https://github.com/skeletonlabs/skeleton/issues/1444#issuecomment-1541461421 and following the second solution, I implemented a MutationObserver to monitor the children of the node and execute the `onScanElements` function.
### Notes
1. Changed function names: (`onInit` -> `onScanElements`, `onDestroy` -> `onCleanUp`) because these functions are also used outside the lifecycle logic and I thought these names describe them better. (I would happily change them back if necessary)
2. The `onScanElements` function now receives a boolean indicating if the call was executed by the observer or at initialization of the `focusTrap`, this boolean is used to prevent the function from focusing on the first element when called by the observer.
In other words: "to not focus on the first element when the children are changed.

### DIFF
--- a/src/lib/actions/FocusTrap/focusTrap.ts
+++ b/src/lib/actions/FocusTrap/focusTrap.ts
@@ -1,5 +1,4 @@
 // Action: Focus Trap
-
 export function focusTrap(node: HTMLElement, enabled: boolean) {
 	const elemWhitelist = 'a[href], button, input, textarea, select, details, [tabindex]:not([tabindex="-1"])';
 	let elemFirst: HTMLElement;
@@ -21,7 +20,7 @@ export function focusTrap(node: HTMLElement, enabled: boolean) {
 		}
 	}
 
-	const onInit = () => {
+	const onScanElements = (fromObserver: boolean) => {
 		if (enabled === false) return;
 		// Gather all focusable elements
 		const focusableElems: HTMLElement[] = Array.from(node.querySelectorAll(elemWhitelist));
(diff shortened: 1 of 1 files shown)


src/lib/actions/FocusTrap/focusTrap.ts
```diff
@@ -21,36 +20,47 @@ export function focusTrap(node: HTMLElement, enabled: boolean) {
...
+	// When children of node are changed (added or removed)
+	const onObservationChange = (mutationRecords: MutationRecord[], observer: MutationObserver) => {
+		if (mutationRecords.length) {
+			onCleanUp();
+			onScanElements(true);
+		}
+		return observer;
+	};
+	const observer = new MutationObserver(onObservationChange);
+	observer.observe(node, { childList: true, subtree: true });
```
Sarenor, line 54: looking good, but I think we need to add an observer.disconnect() in the ``onCleanup()``

Mahmoud-zino: you are absolutely correct, I forgot about it. will add it immediately.

endigo9740:
@Mahmoud-zino just did some test against this and it's freaking awesome! Seems like a perfect fix for this issue. Thank you so much for this. It seems I need to learn more about MutationObserver!

Also - if you're a member of our Discord would you reach out to me on there. Given how many members we have I struggle to keep everything's accounts straight if the username doesn't match. But you've submitted some useful PRs lately so I want to try to keep you in mind for the future!

PR #1486, opened 2023-05-11
App Rail Refactored

### LINKED ISSUE #1478: Revisit the App Rail and App Rail Tiles components

These seem overly complicated and are not working as desired in real world use cases. I might propose we split the tiles into two forms:

- A version that's meant to work with a data driven state or store
- A version that's simply a visual wrapper for an anchor tag

This would make it a bit simpler to mix and match tile types within a single instance.

Additionally, we should also look at moving to the wrapped checkbox format of lists, tabs, and other components. This would still support stores, but not require them.

Enhancement

_No response_

endigo9740:
Fixes #1478


src/routes/(inner)/components/app-rail/+page.svelte
```diff
@@ -96,54 +114,83 @@ const storeValue: Writable<number> = writable(0);
...
 `}
 			/>
 		</section>
+		<section class="space-y-4">
+			<h2 class="h2">Horizontal Orientation</h2>
+			<!-- prettier-ignore -->
+			<p>
+				The App Rail is intended for vertical orientations only. For mobile interfaces, considering using a <a class="anchor" href="/components/tabs#navigation">navigation tab component</a>.
+			</p>
+		</section>
```
Sarenor, line 194: Loving this addition :)

Sarenor:
Looking good and working well as far as I can tell!

PR #1502, opened 2023-05-14
New setup for mono repo with support for ad hoc multi repos

niktek:
This sets up the monorepo for Skeleton.

Just to cover some of the thoughts so far: The decision of what to leave in the mono repo and what should be added add-hoc is determined by an OSS vs private/personal repo distinction.  So skeleton labs.co can add a detached repo into the sites/ folder and it will not be mixed in with the main Skeleton repo.

The other consideration around this is that it is not just the repo, but all the attendant GH Issues tracker, CI/CD, deployments etc.  It's felt that at this point it is better to keep as much together in the one repo as possible until some need changes that.

However - the approach that has been taken of using a Vite alias to link to the Skeleton lib means that these other add-hoc repos need to be setup uniquely to work.  CSA has been extended to automatically add these mods for new sites that are created within this mono repo - `pnpm site <name>` will automatically stick the new site in the sites/ folder and it's README.md covers the next steps and differences from a normal CSA site.

There is no need to do `pnpm dev` from the top level folder like traditional pnpm workspaces, you only need to go into the site you are working on and run `pnpm dev` to work as you normally did before.

No other features of a mono repo have been added as yet, such as a shared config for TS/linting/formatting/githooks/changesets.  This is purely to get a mono repo in place for people to evaluate and ask questions.

@AdrianGonz97 and I have both managed to deploy this to our own Vercel accounts successfully, so we need to wait for @endigo9740 to give us a view into the project settings for docsite and whatever is in the logs.  Most likely it will be the Root Folder setting in the project settings that needs to be changed to sites/skeleton.dev - I'm not sure whether we can do this as a preview branch with a custom root url either.  Prob easiest to do this on a new Vercel project even under a different URL.  Also, should take a look at the project framework setting as they have a Svelte-kit legacy setting that my personal one was one and a new one available.


endigo9740:
Very cool. Will give this a proper review tomorrow. Let's plan to meet and discuss if you're available then.

AdrianGonz97:
I noticed a slight problem with generating new sites and sticking them in `sites` with `pnpm site <name>`. Whenever a new site is added, the root `pnpm-lock.yaml` file is modified as the deps from the generated site are added. We should really modify the `pnpm-workspace.yaml` file to only include `packages/skeleton` and `sites/skeleton.dev`.

Additionally, we need to remove `create-skeleton-app` from the `devDependencies` of the root `package.json` as it interferes with initial installation. `pnpm install` (on a fresh clone) is prevented because `create-skeleton-app` doesn't exist in the beginning, which means that we can't execute the `scripts/checkout.cjs` scripts because we can't install any dependencies due to the error.

To get around this, we can modify the `pnpm site <name>` script for the following:
```ts
// I replaced "site" for "csa" to perhaps make it more clear 
"csa": "node ./packages/create-skeleton-app/src/index.js --types=typescript --prettier --eslint --playwright=false --vitest=false --codeblocks=true --popups=true --typography=false --forms=false -t=skeleton --skeletontemplate=welcome -p=sites --monorepo"
``` 

With this, we can remove the `"create-skeleton-app": "workspace:*"` from `devDependencies` and still have full functionality of CSA.

PR #1530, opened 2023-05-20
feat/paginator-more-options

### LINKED ISSUE #1514: More options for the paginator

I am working with a table that has tens of thousands of entries and server-side pagination. I would like to use Skeleton's paginator, but the way it currently works, it is almost impossible to get to the last page of information or any of the middle pages.

More specifically, I would very much like a "double-arrow" sort of thing that goes either to the last or the first page. And in addition to that, a field where I can input the desired page would be very nice.

Enhancement

_No response_

Mahmoud-zino:
closes #1514 

## Before submitting the PR:
- [x] Does your branch follow our [naming convention](https://www.skeleton.dev/docs/contributing)?

## What does your PR address?
1. Added First and Last buttons in the paginator to navigate to the first or last page.
2. Added page input to navigate to a specific page.

### Notes
1. I didn't bind the input value to handle invalid values without affecting the `settings.offset`.
2. I am unsure if the input needs an indicator like "page" written beside it or using an input group.
3. the page input is of type number so we can handle the input easily, however, the input number adds its own spin box which is redundant because we have the next, and previous buttons. I left it like this to hear your opinion about it and change it if necessary.

Please tell me if I can change or improve anything.

### DIFF
--- a/packages/skeleton/src/lib/components/Paginator/Paginator.svelte
+++ b/packages/skeleton/src/lib/components/Paginator/Paginator.svelte
@@ -29,6 +29,10 @@
 	export let buttonTextPrevious: CssClasses = '&larr;';
 	/** Set the text label for the Next button. */
 	export let buttonTextNext: CssClasses = '&rarr;';
+	/** Set the text label for the First button. */
+	export let buttonTextFirst: CssClasses = '&laquo;';
+	/** Set the text label for the Last button. */
+	export let buttonTextLast: CssClasses = '&raquo;';
 
 	// Base Classes
 	const cBase = 'flex flex-col md:flex-row items-center space-y-4 md:space-y-0 md:space-x-4';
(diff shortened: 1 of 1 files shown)


endigo9740:
@Mahmoud-zino Also, I'm fine if we want to split this task up - limited this to the first/last button so we can get something out for tomorrow's release, then revisit the abbreviated number row mechanic next release. Again, up to you since we're limited on time!

Mahmoud-zino:
@endigo9740 Yeah understandable, splitting the tasks sounds good, I would also love to work on the numbers row feature because I am new to responsive design and would like to learn more (this means I will need your guidance also 🙏).

endigo9740:
@Mahmoud-zino sounds great - drop the second feature for now, then you can revisit the numeric row approach next release. Go ahead and create a dedicated feature request ticket for that when you can please. You're welcome to copy the image I shared above for reference.

yazer79:
The paginator can handle both pages pagination and offset not just offset like now.
Pages can be provided as an array or from api for example:
number of pages - total pages

PR #1533, opened 2023-05-21
feat/dynamic-transitions

### LINKED ISSUE #1322: Implement dynamic animations for all components

There will be multiple parts to this process:

- [ ] We need to determine which components use transitions, so we know which we're working with. @JustBarnt create a list and drop it in a comment below so we can review.
- [ ] We then need to update these components to use the new prop approach, preferably using the split in/out transitions like this PR - https://github.com/skeletonlabs/skeleton/pull/1303
- [ ] We need to mark the old `duration` props as depreciated. Keep them for backwards compatibility for now, but they will no longer be used. Removing them would be a breaking change.
- [ ] Then we need to test and implement the means make transitions opt-in and interchangeable. My theory is we can do this by allowing users to pass their own transitions (either Svelte-provided or custom), similar to how we allow dynamic actions/actionParams for the Avatar component:
https://github.com/skeletonlabs/skeleton/blob/dev/src/lib/components/Avatar/Avatar.svelte#L18

Per that last point, I believe transitions are functions, like actions. I also anticipate both the transition and it's params will need to be passed, like so:

- `transitionIn: [transition, {params}]` - handles the entry transition and params.
- `transitionOut: [transition, {params}]` - handles the exit transition and params.

I'm not opposed to the default `transitionX` props using the same canned Svelte animations they do currently. This would mean that users get the same "eye candy" out of the box they do now.

I'd expect the default value for the param props would be an object containing only the duration value, as I believe this is the only standard key between all transitions:

`export let transitionIn = [slide, { duration: 200 }];`

If you have any questions about this please let me know.

Enhancement

Here's how we handle passing an action/actionParams as part of the Avatar component:
https://github.com/skeletonlabs/skeleton/blob/dev/src/lib/components/Avatar/Avatar.svelte#L14

Here's Svelte's tutorial about the split in/out transition attributes:
https://svelte.dev/tutorial/in-and-out

Here's the Svelte tutorial about custom transitions. They are functions, similar to Actions:
https://svelte.dev/tutorial/custom-css-transitions

### LINKED ISSUE #1694: Drawer: disable opacity transition on demand

Add a boolean prop to disable the opacity animation of the fly transition in the drawer.

Enhancement

As discussed in Discord: https://discord.com/channels/1003691521280856084/1121757065329770657

Mahmoud-zino:
## Linked Issue

Closes #1322
Closes #1694

## Description

1. Added dynamic Transition types.
2. replaced accordion's transitions with dynamic transitions.

## Changsets

feat: Implemented dynamic transitions to the Accordion component.

## Checklist


Mahmoud-zino:
@endigo9740 
As we talked about before, I've applied the changes to the accordion to check with you if the changes are ok.
## Some Notes and clarifications about the changes
1. some of the implementation details are not exactly as we discussed because I tried them and found the current implementation somewhat better. (please let me know if I have to change anything)
2. I used `any[]` for the transition function parameters. is it ok or is there a better solution maybe?
3. I feel like the transition changes need more clarification in the docs, but I don't know if it is a good idea to include a transition section in every component we change or just include it somewhere else.
4. I had to change the `AccordionItem.test` to include the transition parameter because it can't get it from the parent context. Is it correct or should I change it and how?

Thank you for taking the time to check this.

endigo9740:
@ryceg is based near Australia so he'll be on in a few hours - let's wait for his feedback and then I'd like to do one more full audit with all changes in place. Once that's settled we can move forward with other components.

packages/skeleton/src/lib/components/Accordion/AccordionItem.svelte
```diff
@@ -41,8 +41,14 @@
...
 	export let active: Writable<string | null> = getContext('active');
-	/** Set the drawer animation duration. */
+	/** DEPRECATED: use transition or transitionIn, transitionOut instead. */
 	export let duration: number = getContext('duration');
+	/** Set the In/Out transition. */
+	export let transition: [(...args: any[]) => TransitionConfig, {}] = getContext('transition');
+	/** Overrides the In transition with it's params. */
+	export let transitionIn: [(...args: any[]) => TransitionConfig, {}] = getContext('transitionIn');
+	/** Overrides the Out transition with it's params. */
+	export let transitionOut: [(...args: any[]) => TransitionConfig, {}] = getContext('transitionOut');
```
ryceg, line 51:
```suggestion
	/** Set the In/Out transition. */
	export let transition: [(...args: Parameters<typeof slide>) => TransitionConfig, {}] = getContext('transition');
	/** Overrides the In transition with its params. */
	export let transitionIn: [(...args: Parameters<typeof slide>) => TransitionConfig, {}] = getContext('transitionIn');
	/** Overrides the Out transition with its params. */
	export let transitionOut: [(...args: Parameters<typeof slide>) => TransitionConfig, {}] = getContext('transitionOut');
```

packages/skeleton/src/lib/components/Accordion/Accordion.svelte
```diff
@@ -4,15 +4,31 @@
...
+	/**
+	 * Overrides the In transition with it's params.
+	 * @type {[(...args:any[]) => TransitionConfig, {}]|undefined}
+	 */
+	export let transitionIn: [(...args: any[]) => TransitionConfig, {}] | undefined = undefined;
+	/**
+	 * Overrides the Out transition with it's params.
+	 * @type {[(...args:any[]) => TransitionConfig, {}]|undefined}
+	 */
+	export let transitionOut: [(...args: any[]) => TransitionConfig, {}] | undefined = undefined;
```
ryceg, line 31:
```suggestion
	/**
	 * Set the In/Out transition with its params.
	 * @type {[(...args: Parameters<typeof slide>) => TransitionConfig, {}]}
	 */
	export let transition: [(...args: Parameters<typeof slide>) => TransitionConfig, {}] = [slide, { duration: 200 }];
	/**
	 * Overrides the In transition with its params.
	 * @type {[(...args: Parameters<typeof slide>) => TransitionConfig, {}]|undefined}
	 */
	export let transitionIn: [(...args: Parameters<typeof slide>) => TransitionConfig, {}] | undefined = undefined;
	/**
	 * Overrides the Out transition with its params.
	 * @type {[(...args: Parameters<typeof slide>) => TransitionConfig, {}]|undefined}
	 */
	export let transitionOut: [(...args: Parameters<typeof slide>) => TransitionConfig, {}] | undefined = undefined;
```

ryceg (changes requested):
Awesome stuff! Just a couple typing fixes, plus the `@deprecated` syntax

endigo9740:
Thanks @ryceg!

Mahmoud-zino:
@endigo9740 I have implemented the fixes mentioned by @ryceg .
please let me know if I missed something 👍.

packages/skeleton/src/lib/components/Accordion/Accordion.svelte
```diff
@@ -4,15 +4,31 @@
...
+	import { slide, type TransitionConfig } from 'svelte/transition';
 
 	// Types
 	import type { CssClasses } from '../..';
 
 	// Props
 	/** Set the auto-collapse mode. */
 	export let autocollapse = false;
-	/** Set the drawer animation duration in milliseconds. */
+	/** DEPRECATED: use transition or transitionIn, transitionOut instead. */
```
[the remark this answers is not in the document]
Mahmoud-zino: I had to skip this because SVELD won't interpret it correctly.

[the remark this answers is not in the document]
ryceg: Ah, that's a pain, the intellisense for the deprecated flag is super handy. Comes up as a warning. Might be something that needs to be fixed upstream in Sveld. in that case lgtm!

packages/skeleton/src/lib/index.ts
```diff
@@ -14,6 +15,11 @@ export type { PopupSettings } from './utilities/Popup/types';
 // This type alias is to identify CSS classes within component props, which enables Tailwind IntelliSense
 export type CssClasses = string;
 
+export interface TransitionSettings {
+	transition: (...args: Parameters<typeof slide>) => TransitionConfig;
```
endigo9740, line 19: @ryceg I like that @Mahmoud-zino has created a global shared type for reuse, but I'v noted we're using `slide` here as the default. Is there a more generic value we can use here? I believe transitions are just functions by default - do we provide a generic function here?

This may tie into another issue I'll raise in the PR thread in regards to removing animations.

endigo9740:
Additionally @Mahmoud-zino @ryceg - we should make considerations for accessibility here - which includes the ability for users to disable animations. Should we allow a a `null` or `undefined` value to be passed in favor of the object? I'm not sure which the transition directives can support off hand

packages/skeleton/src/lib/index.ts
Mahmoud-zino: @endigo9740 yeah I looked into it, this will work with most transitions but not `draw` or `crossfade` because they take different parameters. I think the only way to make it really generic is by using `any`...

ryceg:
So under the hood, the transitions apply CSS as `t` moves from `0` to `1`. If we implemented a `$prefersReducedMotion` store as Geoff Rich suggests, we can simply use a ternary to apply the correct `duration` for the fallback property. i.e.
```ts
export let transitionIn: TransitionSettings = { transition: slide, params: { duration: $prefersReducedMotion ? 0 : 200 } };
```

Thus, no tweening occurs between `0` and `1`, so the duration would be instant- no motion sickness.

This allows people to still customize the transition behaviour to account for the media query manually, but protects end-users if they just go with the default.

packages/skeleton/src/lib/index.ts
ryceg: We can use a union type;
 
```ts
transition: (...args: Parameters<typeof blur | fade | fly | slide | scale>) => TransitionConfig
```

It's not pretty, and it doesn't support custom transitions that include additional params (although you could slap a ` | Record<string | any>` on the end to accommodate that) but it does work

Mahmoud-zino: @endigo9740 @ryceg Is using `object` a bad idea? I achieved the same functionality by changing the params to object and it works.
```ts
export interface TransitionSettings {
	transition: (node: HTMLElement, params: object) => TransitionConfig;
	params: object;
}
```

ryceg: Typescript works by narrowing wider types; the `object` type simply tells Typescript that it's not one of the other primitives. Both are technically "correct", but you're going to be more _accurate_ with a union type (accuracy in the sense of trying to make your types resemble the corresponding data). 

Try it out- with the union type you'll get intellisense on the object, and it'll narrow down the possible choices as you add props that don't exist on some of the interfaces.

endigo9740:
@Mahmoud-zino @ryceg I commented on the type thread above. I do like the suggestion of the `$prefersReducedMotion` store. I do think we should go ahead and implement that. We don't currently have a location for a global store yet, so I'm leaning towards `/src/lib/stores.ts`.

```ts
import { writable, type Writable } from 'svelte/store';

export const prefersReducedMotion: Writable<boolean> = writable(false);
```

As discussed previously, we'll need a new "Animations" or "Transitions" page in the docs that covers the process of modifying transitions. This store can be covered there.

packages/skeleton/src/lib/index.ts
Mahmoud-zino: @endigo9740 @ryceg 
This is from Svelte\types... and they are using `any` to define the params.
```ts
declare type TransitionFn = (node: Element, params: any, options: TransitionOptions) => TransitionConfig;
export declare function create_in_transition(node: Element & ElementCSSInlineStyle, fn: TransitionFn, params: any): {
    start(): void;
    invalidate(): void;
    end(): void;
};
```
regarding the Union, it works perfectly as long as the user provides params from one of the transitions (which I really think are more than enough).
I tried this after @ryceg mentioned the benefits of Union and it works really well.
```ts
export type TransitionParams = BlurParams | FadeParams | FlyParams | SlideParams | ScaleParams;
export interface TransitionSettings {
	transition: (node: HTMLElement, params: TransitionParams ) => TransitionConfig;
	params: TransitionParams;
}
```
I just tested it with Svelte's custom spinning transition from the [tutorial](https://learn.svelte.dev/tutorial/custom-css-transitions).

endigo9740: Rhys will roll up a newspaper and bop us on the head if we use `any` :P I'll let him comment on the param type though.

Another thought per disabling transitions - one of the primary reasons I wanted to move forward with user-provided transitions was providing a way not to just make the transition have a duration of 0, but to remove the runtime cost of animations, for performance.

The perfect example is the autocomplete component - some folks are adding hundreds or thousands of options to that, and the calculations for the transitions are running per each row. I ended up dropping animations for this component because of this. How do we cover this use case? Do we need to be able to inform users how to essentially remove the transition calculation through the passed value.

Thoughts on how we do that?

Mahmoud-zino: Since the users in these specific cases have to pass undefined to both In and Out transitions separately which IMO doesn't look good, I suggest creating an extra Prop to disable the transitions.
and we can benefit from that by not reducing the duration to 0 if `$prefersReducedMotion` was set, but really just disabling the transitions. This is a lot easier to understand and I think will be cleaner in code.

endigo9740: @Mahmoud-zino I'm not sure another prop would help since the transition directives are still applied to the elements. We need to alter the value passed to the `transition:in` and `transition:out`, correct?

Mahmoud-zino: yes and no, we can do something like [this repl](https://svelte.dev/repl/fa6cdc5ef80d4be29b35c112a1b9c659?version=3.28.0).

endigo9740: Hmm, I like it, but there's a lot of boilerplate. We definitely don't want that to live within every component. It would need to be abstracted for reuse by all components.

I'm was thinking something more like how we handle dynamic actions:
https://github.com/skeletonlabs/skeleton/blob/dev/packages/skeleton/src/lib/components/Avatar/Avatar.svelte#L20

If the action is not present, we still need to pass something to `use:` directive, so we pass an empty function:
https://github.com/skeletonlabs/skeleton/blob/dev/packages/skeleton/src/lib/components/Avatar/Avatar.svelte#L59

In practice that would look something like this:

```html
<SomeComponent
    transitionIn={{transition: () => {}, params: {}}}
    transitionOut={{transition: () => {}, params: {}}}
/>
```

I don't like this though, we're then forcing end users to pass a bunch of extra noise.

This makes me wonder if we build in and include our own custom transaction called `disabled` doesn't doesn't do anything:

```js
function disabled(node, { duration }) {
    return { duration: 0, css: t => `` };
}
```

Then allow users to provide this like so:

```html
<SomeComponent
    transitionIn={{transition: disabled, params: {}}}
    transitionOut={{transition: disabled, params: {}}}
/>
```

Then if we can make params optional, it could be:

```html
<SomeComponent
    transitionIn={{transition: disabled}}
    transitionOut={{transition: disabled}}
/>
```

But I'm not sure this guarantees there won't be a performance hit, even if it's minor.

Really stumped on this one, it's a tough one to figure out.

Mahmoud-zino: @endigo9740 
I like the disabled Idea, but we don't have to return anything from it. Svelte only cares about the function parameters having the correct type.
I don't know how yet but I will try to benchmark both approaches and check the performance.
This might take some time, but I will report back when I have something.

ryceg: I'm not a huge fan of `disabled` being a function- traditionally `disabled` is a boolean, and there's less friction if we don't require them to import the `disabled` function just to pass it in. We could always add in a `disabled` flag in an extended interface very easily, and this would also afford the ability to toggle it.

ryceg:
```ts
export type DefaultTransitionParams = Parameters<typeof blur | typeof fade | typeof fly | typeof slide | typeof scale>[1] &
	AdditionalTransitionProps;
export type TransitionParams = (DefaultTransitionParams | Record<string, any>) & AdditionalTransitionProps;

interface AdditionalTransitionProps {
	/** Disable the transition completely. */
	disabled?: boolean;
	/**
	 * By default, animations will be disabled if the media query `prefers-reduced-motion` is present.
	 * There are many reasons why you might want to override this behavior, such as for minor animations.
	 */
	ignoreReducedMotion?: boolean;
}

export interface TransitionSettings {
	transition: (node: Element, params?: TransitionParams) => TransitionConfig;
	params?: TransitionParams;
}
```

This is an opinionated way of doing it, opting for people having to explicitly set to ignore reducedMotion- this would be my preference, as accessibility should be the default, and people should opt _in_ to patterns that may reduce accessibility. The `disabled` prop also allows for it to be toggled on an off, an escape hatch.

Mahmoud-zino:
@ryceg 
> export type DefaultTransitionParams = Parameters<typeof blur | typeof fade | typeof fly | typeof slide | typeof scale>[1] & AdditionalTransitionProps;

Svelte has `BlurParams`, `FadeParams` .... etc, which we can use instead of taking the parameters from the function and then extracting the type of the second parameter.

I like the idea of extending the interface to take disabled and ignoreReducedMotion. I will implement these changes and wait for you to take a look at them in the code. Thanks

Mahmoud-zino:
@ryceg @endigo9740 
I just added the prefersReducedMotion store and have some notes:
1. I think having `disabled` in the AdditionalTransitionProps doesn't make sense, because to disable the transition the user will have to use something like: `transitionIn={{transition: slide, params:{disabled: true}}}  transitionOut={{transition: slide, params:{disabled: true}}}`, and like Chris said: 
   > I don't like this though, we're then forcing end users to pass a bunch of extra noise.
2. I still think the `ignoreReducedMotion` is a very nice prop to have.
3. I added the function `disabledTransition()` in `index.ts` (I don't know where should we define it) just to test it and see if it makes sense and I really dig it. I know Rhys doesn't like It but it is easier to use if the user wants to disable the transition: `transitionIn={{transition: disabledTransition}} transitionOut={{transition: disabledTransition}}`.

Any thoughts on this?

ryceg:
Agreed, Params may not be the best place to put it, the ergonomics aren't awesome. I'm thinking that being able to pass a `false` to `transition` would be the best; nice and declarative, and allows for an easy ternary operator. A union type would let us do this, easy.

AdrianGonz97:
Just want to note the revised `TransitionSettings` type here:
```ts
type Transition = (node: Element, params?: any) => TransitionConfig;

export type TransitionParams<T extends Transition> = Parameters<T>[1] & AdditionalTransitionProps;

interface AdditionalTransitionProps {
    ignoreReducedMotion?: boolean;
}

export interface TransitionSettings<T extends Transition> {
    transition: T;
    params?: TransitionParams<T>;
}
```
This allows for custom transition functions to maintain type safety for their `params` along side the built-in transitions.

Mahmoud-zino:
@AdrianGonz97 
can you please provide an example of how is this going to be used?
I am trying `export let transitionIn: TransitionSettings = { transition: transitions ? slide : () => {}, params: { duration: 200 } };` but getting different errors regarding the type.
if I do `let transitionIn: TransitionSettings` => `Generic type 'TransitionSettings<T>' requires 1 type argument(s).ts(2314)`
and if do `let transitionIn: TransitionSettings<typeof slide>` => `Type 'void' is not assignable to type 'TransitionConfig'.ts(2322)` 
and just for the sake of trying, I tried `let transitionIn: TransitionSettings<any>` and got `Object literal may only specify known properties, and 'duration' does not exist in type 'AdditionalTransitionProps'`

AdrianGonz97:
Alright, here's the revised version of implementing it on the component side:
```ts
type TransitionOut = $$Generic<Transition>;
type TransitionIn = $$Generic<Transition>;

export let transitions = true;

export let transitionIn: TransitionSettings<TransitionIn> = { transition: slide as TransitionIn, params: { duration: transitions ? 200 : 0 } };

export let transitionOut: TransitionSettings<TransitionOut> = { transition: slide as TransitionOut, params: { duration: transitions ? 200 : 0 } };
```

Unfortunately, we have to use `as` to assert that `slide` is a `Transition`. But since we are defining the default here, we can always guarantee that `slide` is in fact a `Transition`, so it's okay to use in this case.

Mahmoud-zino:
@AdrianGonz97 Thank you for taking the time to fix it.
but we discussed in Discord yesterday that we will use:
`export let transitionIn: TransitionSettings<TransitionIn> = { transition: transitions ? slide : () => {}, params: { duration: 200 } };`
instead of changing the duration...
with the solution you provided we can do:
`export let transitionIn: TransitionSettings<TransitionIn> = { transition: transitions ? slide as TransitionIn : (() => {}) as TransitionIn, params: { duration: 200 } };` and it would work but do you think it's clean enough to implement? please keep in mind we will have to repeat this for every component sometimes even 4 times per component when the component has 2 different transitions.

packages/skeleton/src/lib/stores.ts
```diff
@@ -0,0 +1,22 @@
+import { browser } from '$app/environment';
+import { readable } from 'svelte/store';
+
+/** Prefers reduced motion  */
+const reducedMotionQuery = '(prefers-reduced-motion: reduce)';
+const getInitialMotionPreference = () => {
+	if (!browser) return false;
+	return window.matchMedia(reducedMotionQuery).matches;
+};
+export const prefersReducedMotion = readable(getInitialMotionPreference(), (set) => {
```
endigo9740, line 10: @Mahmoud-zino @ryceg per this store - correct me if I'm wrong, but wasn't it determined that this is something that will need to be enforced within each transition itself? While CSS transitions will support it automatically do to `prefers-reduced-motion`?

I'm worried that we're trying to tackle too many things at once here. Custom transitions and their typing should be the first consideration, then we can circle back to a11y strategies. Basically make it work, then make it work better.

Thoughts?

AdrianGonz97:
@Mahmoud-zino 
> with the solution you provided we can do:
`export let transitionIn: TransitionSettings<TransitionIn> = { transition: transitions ? slide as TransitionIn : (() => {}) as TransitionIn, params: { duration: 200 } };`

Oh my, I don't like that at all 😆. I must've missed it, but what was the reason for not going with duration modification? I looked back in the discussion on Discord, but saw no mention of a definitive method. 

With the exported prop now being realized with it's type, I think it makes more sense than before to go with modifying the duration since it's a much cleaner implementation. 

this:
```ts
export let transitionIn: TransitionSettings<TransitionIn> = {
	transition: slide as TransitionIn,
	params: { duration: transitions ? 200 : 0 }
};
```

versus this:
```ts
export let transitionIn: TransitionSettings<TransitionIn> = {
	transition: transitions ? (slide as TransitionIn) : ((() => {}) as TransitionIn),
	params: { duration: 200 }
};
```

packages/skeleton/src/lib/components/Accordion/AccordionItem.svelte
```diff
@@ -64,6 +68,15 @@
...
 	export let regionCaret: CssClasses = getContext('regionCaret');
+	// ---
+	/** DEPRECATED: use transition or transitionIn, transitionOut instead. */
+	export let duration: number = getContext('duration');
+	// Silence warning about unused props:
+	const deprecated = [duration];
+
+	// Local
+	const { transition: trIn, params: trInParams } = transitionIn;
+	const { transition: trOut, params: trOutParams } = transitionOut;
```
AdrianGonz97, line 79: Would be another breaking change for those that implement `duration` for this component. 
```suggestion
	/** DEPRECATED: use transition or transitionIn, transitionOut instead. */
	export let duration: number = getContext('duration');

	// Local
	const { transition: trIn, params: trInParams } = transitionIn;
	const { transition: trOut, params: trOutParams } = transitionOut;
```

...(continued below in the next comment)

packages/skeleton/src/lib/components/Accordion/Accordion.svelte
```diff
@@ -44,13 +66,21 @@
 	/** Provide arbitrary classes to the caret icon region. */
 	export let regionCaret: CssClasses = '';
 
+	/** DEPRECATED: use transition or transitionIn, transitionOut instead. */
+	export let duration = 200; // ms
+	// Silence warning about unused props:
+	const deprecated = [duration];
```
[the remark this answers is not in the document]
endigo9740: I agree, this is technically correct, but what's the alternative here? We need to be able to deprecate and move forward on certain features. Redundancy is not always practical, especially when it's a small team like ours maintaining a project of this scale.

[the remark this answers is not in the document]
endigo9740: FYI Adrian provided a suggested fix in the next comment (point it out as I missed it originally too)

packages/skeleton/src/lib/index.ts
```diff
@@ -14,6 +15,17 @@ export type { PopupSettings } from './utilities/Popup/types.js';
...
+interface AdditionalTransitionProps {
+	ignoreReducedMotion?: boolean;
+}
+export type Transition = (node: Element, params?: any) => TransitionConfig;
+export type TransitionParams<T extends Transition> = Parameters<T>[1] & AdditionalTransitionProps;
+export interface TransitionSettings<T extends Transition> {
+	transition: T;
+	params?: TransitionParams<T>;
+}
+
```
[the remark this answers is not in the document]
endigo9740: Any suggestions on where and how we name this?

.changeset/calm-dryers-peel.md
```diff
@@ -0,0 +1,5 @@
+---
+"@skeletonlabs/skeleton": minor
+---
+
+feat: dynamic-animations - Added Transition types, replaced accordion transitions with dynamic transitions.
```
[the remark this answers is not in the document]
endigo9740: Since this will likely drop in isolation - before other components are updated to the standard. I might suggestion:

> feat: Implemented dynamic transitions to the Accordion component.

packages/skeleton/src/lib/index.ts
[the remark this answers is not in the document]
AdrianGonz97: It's an awkward one since it can be used in places like in both `lib/components/` and `lib/utilities/`. The easiest would be to just move all the Transition types along with the `CssClasses` type alias into `src/lib/types.ts`. We can use this file to stick all the types that are used rather often throughout the project.

A bit icky, but it keeps our entry point clean.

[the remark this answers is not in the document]
endigo9740: Hmm, I don't like the idea of dropping it in the component or utilities directories given scope can be shared between both.

I believe `/lib/types.ts` was my original proposal for this, but it somehow got merged into `index.ts`. We're likely going to continue having "global" types added over time, so best to decide where we want those to live sooner than later.

I'd say we start there and feel it out.

endigo9740:
I've found a few items of note in the current implementation:

1. Setting `transitions={false}` on the parent Accordion component does nothing. The animation still plays in full even though I've confirmed `duration: 0` is set.
2. Right now users can pass and object to `transitionIn|Out` that contains keys `{transition}` or `{transition, params}`, but they cannot pass `{params}` as `transition` is required. There may be scenarios where users just want to modify the `duration`, `x`, or `y` params. Can we still support this?
3. I'm not sure I like the duration being appended to the params in the HTML template - this feels too far removed. Perhaps we do something like this?

```ts
// Local
let { transition: trIn, params: trInParams } = transitionIn;
let { transition: trOut, params: trOutParams } = transitionOut;
// DEPRECATED: drop the following when duration is removed
trInParams = { ...trInParams, duration }
trOutParams = { ...trOutParams, duration }
```

Then the directives in the template can be simplified to:
```html
in:trIn|local={trInParams}
out:trOut|local={trOutParams}
```

Mahmoud-zino:
@AdrianGonz97 sorry I misunderstood what you meant with the duration order.

@endigo9740 
> Right now users can pass an object to transitionIn|Out that contains keys {transition} or {transition, params}, but they cannot pass {params} as transition is required. There may be scenarios where users just want to modify the duration, x, or y params. Can we still support this?

I really don't think we should do this, it will make things complex for very few benefits.
in the docs there will always be an indication of what the default transition is, so the user will be able to provide the same transition with different params. this might be repetitive in these specific situations but it will save us a lot of work.

AdrianGonz97:
> Right now users can pass an object to transitionIn|Out that contains keys {transition} or {transition, params}, but they cannot pass {params} as transition is required. There may be scenarios where users just want to modify the duration, x, or y params. Can we still support this?


I agree with @Mahmoud-zino, it really would make things way more complex than it has to be. If the user wants fine-grain control over the transition, they can import it and customize it from there.

packages/skeleton/src/lib/components/Paginator/Paginator.svelte
```diff
@@ -1,7 +1,8 @@
 <script lang="ts">
 	import { createEventDispatcher } from 'svelte';
 	// Types
-	import type { CssClasses, PaginationSettings } from '../../index.js';
+	import type { PaginationSettings } from '../../index.js';
```
AdrianGonz97, line 4: While not relevant to this PR, might as well take the opportunity to fix this import.
```suggestion
	import type { PaginationSettings } from './types.js';
```

packages/skeleton/src/lib/components/Table/Table.svelte
```diff
@@ -4,7 +4,8 @@
 	import { tableA11y } from '../../utilities/DataTable/actions.js';
 
 	// Types
-	import type { CssClasses, TableSource } from '../../index.js';
+	import type { TableSource } from '../../index.js';
```
AdrianGonz97, line 7: Same here
```suggestion
	import type { TableSource } from './types.js';
```

endigo9740:
> I really don't think we should do this, it will make things complex for very few benefits.
in the docs there will always be an indication of what the default transition is, so the user will be able to provide the same transition with different params. this might be repetitive in these specific situations but it will save us a lot of work.

> I agree with @Mahmoud-zino, it really would make things way more complex than it has to be. If the user wants fine-grain control over the transition, they can import it and customize it from there.

Unfortunately this is a regression of functionality guys. In literally every component that utilizes transitions we provide a means to quickly modify the duration. This will no longer be possible unless we keep the `duration` prop and add even more.

If we don't modify this behavior the only way for folks to modify the params of the existing param will be to import and pass the exact same transition again, just so they can modify the props. Are we ok with that?

Mahmoud-zino:
@endigo9740 @AdrianGonz97 
## Status update:
1. I have converted all components to use dynamic transitions.
2. we ignored Backdrop transitions as discussed on discord
3. also we ignored the drawer transition
4. for the ignored components I still added the transitions prop so that users could turn off transitions and/or prefersReducedMotion to still have effect on all transitions (since drawer slide for example is not a minor transition).
5. I will resolve the conflict with the new bugfix regarding the drawer in/out transition whenever v2 is updated to div (I wrote Coka on Discord in this regard)
6. we still have an issue with crossfade transition, but @AdrianGonz97 was going to look at it when he has time.

endigo9740:
@Mahmoud-zino thank you so much for the quick summary - this was super helpful.

Just did a pass through the project and almost everything is there. Just a few odds and ends.

1. A few of the components still contain the `duration` property. Remember as part of this targeting a v2 release we need to permanently remove this property from all components. The new transition system replaces this.

2. Modals has a number of transitions properties in addition to duration. These need to be removed. This includes:

```
// Props (transitions)
/** The open/close animation duration. Set '0' (zero) to disable. */
export let duration = 150;
/** Set the fly transition opacity. */
export let flyOpacity = 0;
/** Set the fly transition X axis value. */
export let flyX = 0;
/** Set the fly transition Y axis value. */
export let flyY = 100;
```

These are also passed to the `$: parent` declaration, so make sure that's removed as well. This allows access from any custom modals - so let's do a quick audit and make sure none of these are used in our example or other modal components. I don't believe this is the case though.

3. We need to implement some addition developer documentation that explains how this works for contributors. This will help folks contributing new components to know how this works and how to implement our new standard. Feel free to do the first draft @Mahmoud-zino and then I'll follow up. This should live within the Style Guide between the CSS Style Conventions and the Pitfalls sections:

https://www.skeleton.dev/docs/contributing/style-guide

@AdrianGonz97 @ryceg I'd also encourage you guys do a quick one-over to verify the typing was implemented to your preference.

Mahmoud-zino:
@endigo9740 
> A few of the components still contain the duration property. Remember as part of this targeting a v2 release we need to permanently remove this property from all components, so this breaking change is allowed. The new transition system replaces this.

these durations are required for animations like flip...

> Likewise the Modal component has a number of transitions properties in addition to duration. These need to be removed. This includes:

no idea why I missed that. I deleted them, thank you for the reminder.

> We need to implement some addition developer documentation that explains how this works for contributors. This will help folks contributing new components to know how this works and how to implement our new standard. Feel free to do the first draft @Mahmoud-zino and then I'll follow up. This should live within the Style Guide between the CSS Style Conventions and the Pitfalls sections:

the draft is done. 👍

endigo9740:
@Mahmoud-zino this is still on my list to review - just lacking the time this week. Thanks for your patience!

endigo9740:
@Mahmoud-zino I've done a follow up per the items mentioned in my prior review:

> (InputChips) uses these durations are required for animations like flip...

Gotcha, we can leave that for now then.

> (per Modal's use of `duration`) no idea why I missed that. I deleted them, thank you for the reminder.

Thanks for adjusting this - looks good.

> (per the contributor docs) the draft is done. 👍

Looks great, I just committed a few tweaks and adjustments to the text and provided a more detailed example snippet. But overall it does what it needs to do. Great job.

### Additional Items

1. I reviewed your solution for the Drawer `opacityTransition`. I think this will suffice for now. It's currently the exception to the rule of how transitions are handled, but probably fine for now.
2. I noted that Toasts still implement `duration` due the fact it implements the custom crossfade animation by (Cory) @saturnonearth. Unfortunately this is an exception to our new standard, so I'm leaning towards us reverting back to a standard animation in the meantime (fly I believe?). We shouldn't delete the custom transition, but it should be abstracted and removed from the component and moved to it's own standalone file in `/src/lib/transitions/crossfade.ts` for now.

If you want to take a stab at converting this the custom Crossfade transition to a proper reusable JS transition per the Svelte Docs you can @Mahmoud-zino, but otherwise I think we set it aside and revisit custom transitions in a dedicated pass in the future. Here's the docs if you want to reference how that works:
https://svelte.dev/tutorial/custom-js-transitions

If you can update the Toast transition I think this PR is officially ready for the v2 release!

saturnonearth:
> I noted that Toasts still implement duration due the fact it implements the custom crossfade animation by (Cory) @saturnonearth. Unfortunately this is an exception to our new standard, so I'm leaning towards us reverting back to a standard animation in the meantime (there is a canned crossfade transition available).


I agree

packages/skeleton/src/lib/index.ts
```diff
@@ -10,9 +10,8 @@ export type { ToastSettings } from './utilities/Toast/types.js';
 export type { TableSource } from './components/Table/types.js';
 export type { PaginationSettings } from './components/Paginator/types.js';
 export type { PopupSettings } from './utilities/Popup/types.js';
-
-// This type alias is to identify CSS classes within component props, which enables Tailwind IntelliSense
-export type CssClasses = string;
+export type { Transition, TransitionParams } from './internal/transitions.js';
+export type { CssClasses } from './types.js';
 export type SvelteEvent<E extends Event = Event, T extends EventTarget = Element> = E & { currentTarget: EventTarget & T };
```
AdrianGonz97, line 16: This is unrelated, but could you move these two over to the `types.ts` file and rexport them into `index.ts` (just like how we did it with `CssClasses`)? This PR introduces the new file so we might as well take advantage 😄

PR #1559, opened 2023-05-24
New Component: Counter

LukeHagar:
## Before submitting the PR:

## What does your PR address?

### Tips

### DIFF
--- a/packages/skeleton/src/lib/components/Counter/Counter.svelte
+++ b/packages/skeleton/src/lib/components/Counter/Counter.svelte
@@ -1,0 +1,117 @@
+<script lang="ts">
+	import { fly } from 'svelte/transition';
+	import { onDestroy, onMount } from 'svelte';
+
+	// Types
+	import type { CssClasses, Transition, TransitionFunction } from '../..';
+
+	// Props (actions)
+	/** Provide the array of values to tick through. */
+	export let values: any[] = [];
+	/** Bind to the current value of the counter. */
+	export let currentValue: any;
+	/** Specify the values array should tick backwards. */
+	export let direction: 'forward' | 'backward' | 'random' = 'forward';
+	/** Select items by index from the values array. */
+	export let index: number = direction === 'backward' ? values.length - 1 : 0;
+	/** Specify how often in milliseconds the value should update. */
+	export let interval: number = 1500;
+
+	// Props (transitions)
+	/** Provide the svelte transition to use when values move in. */
+	export let transition: Transition | undefined;
+	/** Provide the svelte transition to use when values move in. */
+	export let transitionIn: Transition | undefined;
+	/** Provide the svelte transition to use when values move out. */
+	export let transitionOut: Transition | undefined;
+
+	// Props (styles)
+	/** Provide classes to set background styles. */
+	export let background: CssClasses = 'bg-surface-400-500-token';
+	/** Provide classes to set counter height. */
+	export let height: CssClasses = 'h-8';
+	/** Provide classes to set counter width. */
+	export let width: CssClasses = 'w-24';
+	/** Provide classes to set border styles. */
+	export let border: CssClasses = '';
+	/** Provide classes to set rounded style. */
+	export let rounded: CssClasses = 'rounded-token';
+	/** Provide classes to set shadow styles. */
+	export let shadow: CssClasses = '';
+	/** Provide classes to set cursor styles. */
+	export let cursor: CssClasses = '';
+
+	// Base Classes
+	let cBase = 'flex flex-col text-surface-50 font-semibold justify-center items-center overflow-hidden isolate';
+
+	// Reactive Classes
+	$: classesBase = `${cBase} ${background} ${height} ${width} ${border} ${rounded} ${shadow} ${cursor} ${$$props.class ?? ''}`;
+
+	// Utility Functions
+	function prunedRestProps(): any {
+		delete $$restProps.class;
+		return $$restProps;
+	}
+
+	let transitionInterval: any, tIn: TransitionFunction, tInProps: any, tOut: TransitionFunction, tOutProps: any;
+
+	$: currentValue = values[index];
+
+	$: {
+		tIn = fly;
+		tInProps = { y: 100, duration: 100, delay: 100 };
+		tOut = fly;
+		tOutProps = { y: -100, duration: 100 };
+
+		if (transitionIn) {
+			tIn = transitionIn.transition;
+			tInProps = transitionIn.props;
+		}
+		if (transitionOut) {
+			tOut = transitionOut.transition;
+			tOutProps = transitionOut.props;
+		}
+		if (transition) {
+			tIn = transition.transition;
+			tInProps = transition.props;
+			tOut = transition.transition;
+			tOutProps = transition.props;
+		}
+	}
+
+	onMount(() => {
+		transitionInterval = setInterval(() => {
+			switch (direction) {
+				case 'forward':
+					if (index === values.length - 1 || index > values.length - 1) {
+						index = 0;
+					} else {
+						index = index + 1;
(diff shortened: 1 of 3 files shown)


luke-hagar-sp:
Issues have arisen with the markup of the example page

My findings are detailed further here:
https://discord.com/channels/1003691521280856084/1007673252992262215/1111297715641536613

Additionally there is an open thread with Svelte awaiting comment here:
https://discord.com/channels/457912077277855764/1111309554018439258

endigo9740:
@luke-hagar-sp

Per your issue above, I believe dual binding might not be needed here. Instead let's only bind `index`, but instead use an event emitter that triggers when the transition occurs. Something like `on:transition` or `on:change`, which can bundle any relevant information. If we do this I don't believe we need `currentValue` at all.

Here's a few other suggestions and questions to follow up on:

## Suggestions

- Make sure add a period to the end of the doc description at the top of the page.
- The component is called "Counter", we should probably make the primary example count up or down to sell the concept
- We should support background, rounded, and similar props, but not populate them. Keep it text-only by default. A single styled example is welcome.
- This should be an inline element by default, so favor `<span>` over `div` and drop the `p` tag. Like `This is some (counter) text in a line`
- We should avoid inline classes like `text-center` - move that up to the base classes in the script tag
- Your mix of Switch Case and If statements feels a bit verbose. Consider at least transferring the If statements to ternary
    - ex: `index = values.length - 1 || index > values.length - 1 ? 0 : index + 1;
    - NOTE: totally fine to leave this as is until the logic is sorted though. I do that too!
- I believe the easing of the reference library kind of make or break this - [Joy of Code](https://joyofcode.xyz/animation-with-svelte) covers how to implement Svelte's canned options

## Questions

- I'm not sure I understand the `$: {}` block for transitions the script, why not just set the default transition prop values instead of using `undefined`?
    - Reference: https://github.com/skeletonlabs/skeleton/blob/563873625f1a8f9a0fae6daacd7ef076b9f540dd/packages/skeleton/src/lib/components/Accordion/Accordion.svelte#L19
    - If this is for timing, we can maybe do `transitionInDuration + userProvidedDuration + transitionOutDuration` to know the full interval duration, right?
    - This might also help avoid the need for the `onMount`, which can be very slow
- What's the use case of restProps? That's typically reserved for input components that have a billion attributes
- What's the use case of the `cursor` prop? This is never interactive as far as I can tell

endigo9740:
@LukeHagar FYI I updated my notes above to a checklist - feel free to tick each item as these are completed. I'll push this back to a draft state, set to "ready to review" when you wish for me to take another look.

endigo9740:
@luke-hagar-sp @luke-hagar-sp 

Not sure if you've had a chance to keep up with the dynamic transition conversation on Discord. But FYI I would to get this component in next week's release. Unfortunately I don't think we'll have our standard established for transitions by then.

If you don't mind, let's may go ahead and nix the dynamic transitions and treat this like any other existing component - where the transitions are built in with some nice defaults. We can still allow a `duration` to follow current standards. But otherwise that should simplify a lot of this.

If you think you can tackle this by EOD tomorrow let me know, otherwise I might take a stab at this if you don't mind.

endigo9740:
@luke-hagar-sp I've implemented a whole slew of updates to this component based on the feedback above. Here's a quick rundown of the changes:

1. The transitions are now hardcoded - we'll revisit once the standard for this is finalized
2. I've removed the dual binding for `index` and `current` as this is redundant, we can retrieve both points of data purely from the `index`
3. Given this, I've removed the `current` proper altogether
4. However, I've also introduced a new `on:tick` event which conveys similar info and can be "actionable"
5. I've introduce a number of new style props that I think make sense (font weight, color, etc)
6. Refactored a lot of logic to account for the hardcoded transitions and readability
7. The `switch` statement is now much more compact for example

Additionally I've updated the documentation page to have a better flow - introducing more complex concepts as you scroll down. Updated the text a bit, and the examples themselves - with some minor cosmetic improvements too.

Things I'd still like to address:

1. The height - I think this should be auto-determined rather than user specified if possible.
2. Vertical alignment - this relates to height, but when a fixed height is defined the text should be centered vertically
3. Additional styled examples, maybe showcase flipping between an array of SVGs?
4. To support SVG we'll need to change to `{@html currentValue}`
5. Easing settings - some room for improvement here

Let me know if you have any other ideas though - we'll include those as well.

Thanks bud, hope you're doing well though!

Mahmoud-zino:
@endigo9740 
> The height - I think this should be auto-determined rather than user-specified if possible.

this is not possible because the transition will double the height, making it look weird. but If you know some CSS trick to solve this I would be happy to implement it. (I tried max-h-min with no effect 😢)

> Vertical alignment - this relates to height, but when a fixed height is defined the text should be centered vertically

done.

> Additional styled examples, maybe showcase flipping between an array of SVGs?
To support SVG we'll need to change to {@html currentValue}

also done.

> Easing settings - some room for improvement here

this is just a personal preference, but I disliked the bounciness of the transition so switched the easing to `sineInOut`, if you disagree I can switch it back.

> Horizontal support? - perhaps with dynamic transitions

I will have to test this once we have the dynamic transitions ready.

also, I don't have write access, so I can't mark the PR as ready for review or change the first comment to reflect the new PR template.

endigo9740:
> this is not possible because the transition will double the height, making it look weird. but If you know some CSS trick to solve this I would be happy to implement it. (I tried max-h-min with no effect 😢)

Sorry yes, I meant to say to leave that one to me! I have some ideas but will need to test this directly.

> this is just a personal preference, but I disliked the bounciness of the transition so switched the easing to sineInOut, if you disagree I can switch it back.

Hmm I do like that. I'm good with this.

> I will have to test this once we have the dynamic transitions ready.

Makes total sense. No worries.

> 
@endigo9740

The height - I think this should be auto-determined rather than user-specified if possible.

this is not possible because the transition will double the height, making it look weird. but If you know some CSS trick to solve this I would be happy to implement it. (I tried max-h-min with no effect 😢)

Vertical alignment - this relates to height, but when a fixed height is defined the text should be centered vertically

done.

Additional styled examples, maybe showcase flipping between an array of SVGs?
To support SVG we'll need to change to {@html currentValue}

also done.

Easing settings - some room for improvement here

this is just a personal preference, but I disliked the bounciness of the transition so switched the easing to sineInOut, if you disagree I can switch it back.

Horizontal support? - perhaps with dynamic transitions

I will have to test this once we have the dynamic transitions ready.

also, I don't have write access, so I can't mark the PR as ready for review or change the first comment to reflect the new PR template.

I'll leave it pending until I get a chance to revisit the height styling stuff above. Thanks @Mahmoud-zino!

endigo9740:
> this is not possible because the transition will double the height, making it look weird. but If you know some CSS trick to solve this I would be happy to implement it. (I tried max-h-min with no effect 😢)

I meant to mention I'll revisit this, I have some ideas. We'll leave the PR pending until then!

> this is just a personal preference, but I disliked the bounciness of the transition so switched the easing to sineInOut, if you disagree I can switch it back.

I like it, seems good to me!

> I will have to test this once we have the dynamic transitions ready.

Makes total sense. I figure we'll want to revisit this when that occurs.

endigo9740:
@LukeHagar @Mahmoud-zino This one has been languishing for a bit, which is on my shoulders as well. Just haven't had a chance to return to figure out the styling issue. I'm going to encourage you guys to take a look, and if we can't get this by next release I may go ahead and close this out. We can always refer to the PR in the future when we have more time.

With the big updates planning for Skeleton, it may make more sense to reintroduce this alongside those anyways.

Thanks guys

Mahmoud-zino:
@LukeHagar If you want me to take another look at this I can, I just need you to sync the branch with dev. It is refusing to sync for me and the branch is about 300 commits behind 😄

PR #1564, opened 2023-05-25
add toast type

ryceg:
## Before submitting the PR:

## What does your PR address?

Adds the CSSClasses autocomplete goodness to the toast


endigo9740:
@ryceg I believe @niktek raised a good point about this the other day - the custom `CssClasses` type is only relevant for component props. It's what enables his post-build script to gather those values and ensure IntelliSense works for those. In this case I think we should just call a duck a duck. The expected values is a string and the description notifies the user this should be a class.

PR #1573, opened 2023-05-26
feat: pagination-controls

### LINKED ISSUE #1536: pagination controls

we need a pagination control (row of page numbers with `...` to limit horizontal space) to navigate easily between pages.
something like this:

239981714-7839aa1b-20e7-412f-9c92-2d33ca2d92af

Mock-up and notes from @endigo9740 
Screenshot_2023-05-22_at_1 06 35_PM

> Seems very busy on the right, so we'll probably want to make sure those are small buttons.

> We'll also need to track and active state for which button is currently selected. This should be prop class.

New Feature

https://mantine.dev/core/pagination/
The animations here could be a nice addition also: https://nextui.org/docs/components/pagination

Mahmoud-zino:
## Linked Issue

Closes #1536 

## Description

- added Numeric Row with a prop to show it and a prop to define the max active page siblings.

## Changeset

feat: added pagination-control Numeric row

## Checklist

### DIFF
--- a/.changeset/empty-apricots-judge.md
+++ b/.changeset/empty-apricots-judge.md
@@ -1,0 +1,6 @@
+---
+"skeleton.dev": minor
+"@skeletonlabs/skeleton": minor
+---
+
+feat: added pagination-control Numeric row
--- a/packages/skeleton/src/lib/components/Paginator/Paginator.svelte
+++ b/packages/skeleton/src/lib/components/Paginator/Paginator.svelte
@@ -19,6 +19,8 @@
 	export let showFirstLastButtons = false;
 	/** Displays a numeric row of page buttons. */
 	export let showNumericRow = false;
+	/** Maximum number of active page siblings in the numeric row.*/
+	export let maxNumericSiblings = 1;
 	/** Provide classes to set flexbox justification. */
 	export let justify: CssClasses = 'justify-between';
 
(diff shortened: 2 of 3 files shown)


Mahmoud-zino:
@endigo9740 I just finished the logic for the paginator controls.
## Notes
1. The `getControlPages` function could be not as readable as other parts in the code. I would appreciate any tips on how to make the code cleaner.
2. for now I have the maxPagesToShow variable as const, but we could also make it a prop to let the user control how many buttons are displayed. (I also have a hard time finding a fitting name for the prop because it describes how many buttons are displayed in the controls including separators (...) and excluding the page first and last page numbers).
3. I deleted the functions (onPrev, onNext, onFirst, onLast) in favor of gotoPage which can handle all options.
4. I added a `showPageControls` just like `showFirstLastButtons` to not make it a breaking change.
5. I failed miserably in handling responsive design 😢. tried media queries but the sizing is not working for me, maybe you can point me in the correct direction🙏.

Mahmoud-zino:
@endigo9740 Done 👍.

endigo9740:
@Mahmoud-zino I've gone through and implement some design changes. This includes the following changes:

1. We now go all-in on the button group styling, which is set using a `regionControl` prop
2. The control also has a dedicate prop for variant and separator styles
3. I've reworked or introduce a couple other button props to support this
4. I've reworked several portions of the template to follow our conventions for class state, styling, etc
5. I added conditions to hide the select input when `settings.amount` is an empty array
6. I've added a new `showPreviousNextButtons` prop to allow you to toggle these like first/last
7. I've updated the documentation to showcase as much of this as I can.

I have NOT yet done a proper review of your numeric row logic - but I'll have to circle back to that next week due to time constraints today. That said, it's fairly close to how I imagined it, but I did drop from 5 items to 3 and I very much recommend we make that a prop when we can. The problem here is when you're dealing with a fixed screen width (mobile; small screen) even at a small size this can take up a lot of vertical room if the options are too high. Not sure how we tackle that yet!

Mahmoud-zino:
@endigo9740 It looks awesome.
I will implement the property and try to document it.

Mahmoud-zino:
@endigo9740 
I have added the prop `maxNumericSiblings` and changed the logic of `getControlPages` so it adapts to `maxNumericSiblings` changes.

endigo9740:
@Mahmoud-zino did another pass through this today. A few suggestions to address your requests above...

> The getControlPages function could be not as readable as other parts in the code. I would appreciate any tips on how to make the code cleaner.

Sometimes you're just going to have a lot of logic to implement a feature, but here's a few suggestions:

1. Keep your local vars within the function declared at the top
2. Consider splitting your single (monolith) function into smaller modular parts, which can help self-document
3. Consider implementing inline docs to explain each bit of logic ([example](https://github.com/skeletonlabs/skeleton/blob/dev/packages/skeleton/src/lib/utilities/Popup/popup.ts#L97)) 
4. Write simple `if` statements as single line, ex: `if (!isWithinLeftSection) pages.push(-1);`
5. Can any of your if/else logic be written with `else if` statements? Given a lot of it looks related. No worries if not.

> I have added the prop `maxNumericSiblings` ...

Our goal should be to keep props short and semantic so they are easier to remember. Perhaps we modify the follow two props to aid this?

- `showNumericRow` -> `showNumerals`
- `maxNumericSiblings` -> `maxNumerals`

This helps identify the props are related, and two words is definitely better than three here. Thoughts?

> I also added a note about it in the docs.

Looks good. I'd go ahead and remove the alert format and try to integrate this into the paragraph description above the example. Happy to help rewrite that if needed, just let me know.

endigo9740:
@Mahmoud-zino I've made a couple super nitpick-y changes, but otherwise this is good to go. Merging now! Great job on this one!

PR #1577, opened 2023-05-26
Bugfix: dark bg-variant-hover-token

### LINKED ISSUE #1574: `bg-primary-hover-token` on select option only works in lightmode
### Current Behavior

When applying the `bg-primary-hover-token` utility class to a select option, it only produces a visible result in lightmode.

```html
<select class="select" size="4">
  <option class="bg-primary-hover-token">Foo</option>
  <option class="bg-primary-hover-token">Bar</option>
  <option class="bg-primary-hover-token">Baz</option>
</select>
```

### Anything else?

_No response_

Mahmoud-zino:
fixes #1574

## Before submitting the PR:
- [x] Does your branch follow our [naming convention](https://www.skeleton.dev/docs/contributing)?

## What does your PR address?
`.dark .bg-variant-hover-token` was missing.

### DIFF
--- a/.changeset/purple-mice-fly.md
+++ b/.changeset/purple-mice-fly.md
@@ -1,0 +1,5 @@
+---
+"@skeletonlabs/skeleton": minor
+---
+
+Added missing .dark .bg-variant-hover-token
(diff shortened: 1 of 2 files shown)


endigo9740:
@Mahmoud-zino per my issue in the original ticket - I'm going to dig into this a bit next week before we move forward with this PR. This may be our fix, but I want to understand what's going on here. We technically shouldn't need this.

Mahmoud-zino:
@endigo9740 understandable, I will be happy to also learn what is going on👍

endigo9740:
@Mahmoud-zino unfortunately it's not looking like I'll have time to dig into this right now. We'll have to circle back. For now we'll implement this as the fix.

PR #1584, opened 2023-05-27
bugfix: add `.js` extension to all relative imports

### LINKED ISSUE #1501: Typescript import should be made with the .js extension
### Current Behavior

So it seems that without the extension it is fine for CJS, but not for ESM: https://www.typescriptlang.org/docs/handbook/esm-node.html#type-in-packagejson-and-new-extensions

But leaving off the .js extension is also causing some issues for some other people trying to get things to work in web pack.

It might be in our best interests to default to using .js on all appropriate imports by default to ensure the widest compatibility with environments we might not even be aware of.

@Firelight on discord pointed to this package that looks like it could do a global replace of those imports as a one off for us.

Clearly I would like to test this out more before giving a final opinion.
https://www.npmjs.com/package/ts-transform-esm-import

_No response_

### Anything else?

_No response_

AdrianGonz97:
## Linked Issue

Closes #1501 

## Description

Added `.js` extension to all relative imports. We'll now be adhering to Node's ESM algorithm as specified [here](https://kit.svelte.dev/docs/packaging#caveats) in SvelteKit's packaging docs.

## Changsets

## Checklist

### DIFF
--- a/.changeset/pink-vans-rule.md
+++ b/.changeset/pink-vans-rule.md
@@ -1,0 +1,5 @@
+---
+"@skeletonlabs/skeleton": patch
+---
+
+bugfix: added `.js` extension to all relative imports
--- a/packages/skeleton/package.json
+++ b/packages/skeleton/package.json
@@ -117,12 +117,5 @@
 		"./dist/tailwind/*",
 		"./dist/themes/*",
 		"!./dist/**/*.test.*"
-	],
-	"typesVersions": {
-		">4.0": {
-			"index": [
-				"./src/lib/index.ts"
-			]
-		}
-	}
+	]
 }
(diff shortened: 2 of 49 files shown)


endigo9740:
@AdrianGonz97 is this something the VS Code auto-importer can be configured to handle? I can't recall if it's native or an extension off hand, but I worry this won't be maintained over time. Especially by contributors that don't know this is needed.

AdrianGonz97:
@endigo9740 I added `"moduleResolution": "nodenext"` to the `tsconfig.json` so that the editor and linter will yell at you if you don't add it. Additionally, if you're importing something, the intellisense will suggest appending the `.js` extension at the end of the file name as an autocomplete.

PR #1587, opened 2023-05-28
bugfix: changed `typesVersions` map to point to `index.d.ts`

### LINKED ISSUE #1581: All Components and Utilites are auto imported from the wrong paths
### Current Behavior

When typing any component inside your markup without importing it first and importing it via autocomplete (VSCode in my case) will result in an import looking like:
```js
import { ... } from '@skeletonlabs/skeleton/index';
```
Whilst the correct import should be:
```js
import { ... } from '@skeletonlabs/skeleton';
```

### Expected Behavior

Autocompleted imports should be imported like:
```js
import { ... } from '@skeletonlabs/skeleton';
```

### Steps To Reproduce

1. Open vscode
2. Open a Skeleton project
3. Navigate to any component/page that does not have any skeleton imports
4. Type for e.g. <AppBar /> inside the markup
5. Hit ``alt`` + ``.``
6. Enter the option: ' Add import from "@skeletonlabs/skeleton/index" '
7. Observe the wrong component being imported (or atleast from the wrong path)

_No response_

### More Information

This will only happen if the current svelte component/page doesn't already have anything imported from skeleton.

AdrianGonz97:
## Linked Issue

Closes #1581 

## Description

Replaces `index` for `index.d.ts` from `typesVersions` map in `package.json`. Even though SvelteKit's [packaging docs](https://kit.svelte.dev/docs/packaging#typescript) states:
> Note that if you opt into typesVersions you have to declare all type imports through it, including the root import (which is defined as `"index": [..]`). 

I think this may actually be a typo in the SvelteKit docs.

Fortunately, this change won't negatively effect any users that may already be importing their components with the `@skeletonlabs/skeleton/index` path since it was never a valid import path in the first place. If they imported anything from this path, the app would just error.

I also added the type declaration path for our plugin as well since that was missing.

## Changsets

## Checklist

### DIFF
--- a/.changeset/smooth-rockets-know.md
+++ b/.changeset/smooth-rockets-know.md
@@ -1,0 +1,5 @@
+---
+"@skeletonlabs/skeleton": patch
+---
+
+bugfix: updated `typesVersions` map to fix auto imports paths
--- a/packages/skeleton/package.json
+++ b/packages/skeleton/package.json
@@ -85,13 +85,13 @@
 			},
 			"./styles/*": "./dist/styles/*",
 			"./themes/*": "./dist/themes/*",
-			"./tailwind/skeleton.cjs": "./dist/tailwind/skeleton.cjs"
+			"./tailwind/skeleton.cjs": {
+				"types": "./dist/tailwind/skeleton.cts",
+				"default": "./dist/tailwind/skeleton.cjs"
+			}
 		},
 		"typesVersions": {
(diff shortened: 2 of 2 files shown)


endigo9740:
@AdrianGonz97 test locally in a standalone Skeleton project using a tarball version. The import resolves without `/index` appended to the end. Great fix, merged!

PR #1604, opened 2023-05-30
Fixed tabbing issue

### LINKED ISSUE #1602: Docs Search allows for the container holding the options to be tabbed on in FireFox
### Current Behavior

When tabbing through the items from the search on the docs site, the first thing you tab onto it the container that holds all the searchable options, this doesn't server any purpose.

### Expected Behavior

When tabbing off the search you'd want to tab onto the first element in the list.

### Steps To Reproduce

1. Open Firefox
2. Go to https://www.skeleton.dev/
3. hit Ctrl + K
4. hit tab
5. Your tab focus should now be on the container holding the searchable elements

_No response_

### More Information

This is fixable by setting the tabindex to -1 on the container to make it avoid tabs, this shouldn't concern accessability since it does not serve any purpose

Hugos68:
## Linked Issue

Fixed #1602

## Description

Added ``tabindex="-1"`` to autocomplete element containers to ensure users on firefox do not tab onto the container itself. This basically removes the tab-ability from an element and make it un-tab-able.

## Changsets

## Checklist

### DIFF
--- a/.changeset/funny-plants-burn.md
+++ b/.changeset/funny-plants-burn.md
@@ -1,0 +1,5 @@
+---
+"@skeletonlabs/skeleton": patch
+---
+
+docs: Updated autocomplete documentation to demonstrate Firefox tab navigation fix for wrapping elements


Hugos68:
A problem I'm facing with applying the fix is that the ``tabindex="-1"`` needs to be applied to the container that gets focus, and although the ``<ul>`` in the autocomplete holds the elements this is actually not the container that gets focused on, I will investigate this but it really appears that the  ``tabindex="-1"`` needs to be applied outside the Autocomplete component which could cause trouble because not every end-developer will think of that. Not sure what to do here really.

endigo9740:
@Hugos68 so you've definitely found a tough issue here. I'm able to replicate the issue via the steps provided in your ticket.

That said, testing your updated in this PR does not resolve the issue and actually makes it somewhat worse - Instead of focusing into the search results, tabbing instead jumps to the background elements in the App Bar (explore, theme, etc).

I'm going to ping @Mahmoud-zino to review this as well since he recently made some modifications to the focus trap which *should* prevent this from ever occurring. So essentially we have two issues at play here:

1. In Firefox, focus is not being set to the search list result lists (list of anchors) as expected
2. In Firefox, the focus trap is not working as expecting, allowing you to escape the active modal

Unfortunately I have no guidance on this yet. It's honestly very frustrating to learn that Firefox is not operating as Chrome, which works exactly as expected.

endigo9740:
@Hugos68 given the short time remaining this week I'm going to go ahead and merge this. If you do note any issues let me know and we can patch. Thanks.

PR #1614, opened 2023-05-31
Trap focus no longer loses focus when clicking on non-tabbable elements

### LINKED ISSUE #1613: Focus Trap loses focus after mouse click (or any other type of focus not from tabbing)
### Current Behavior

A good example of the behaviour is when opening the docs search and clicking on any element that is not the docs search bar or elements. After doing that and hitting tab the focus trap basically lost focus and this is undesired behaviour.

Demonstration: https://www.youtube.com/watch?v=axzLbWhhxsM

### Expected Behavior

When selecting anything else inside the dialog and then hitting tab again I would expect the focustrap to stay.

### Steps To Reproduce

1. Using focustrap
2. Click on any non-tabbable element.
3. Hit Tab
4. Focus trap is lost

https://www.skeleton.dev/

### More Information

I have already found an pretty easy fix for this and will work on a PR immediately after submitting this issue.

Hugos68:
## Linked Issue

Fixed #1613

## Description

Instead of adding event listeners to each child, I added 1 keydown listener to the dom, this prevents adding and removing listeners every time the targeted node has changes (observed by the MutationObserver). The main reason I did this is to add 1 more check after ``Shift + Tab`` and ``Tab`` have been checked: Now when both tab scenario's were falsy focusTrap will do 1 additional check to check wether the currently active element is not inside the targeted node, if not it will focus the first child inside the targeted node again. This utilizes ``document.activeElement`` which will return the current active element (like the name suggests). This way when a user clicks on an element that is not tabbable and hits tab after, it will first check if that element is part of our focusableChilds list, if not it will prevent the user from tabbing (``e.preventDefault()``) and instead focus the firstFocusableChild (determined by the MutationObserver).

Edit (Important note):
After discussing with @endigo9740 and @Mahmoud-zino we've decided to make the behaviour I implemented optional by user, this ensures modals can be focused without escape and forms can be escaped if the users wishes to. Because that would introduce a breaking change we've decided to, for the time being, allow a ``data-focus-strict`` attribute to be set on the node focus trap is applied to, it will look like:
```html
<div use:focusTrap={true} data-focus-strict>
```
The focus trap action will respect this prop and focus strictly (no escape) if the attribute is set. This will be removed when Skeleton Version 2 hits and instead will be part of a parameter passed on to the focusTrap action. So it will convert from:
```ts
export function focusTrap(node: HTMLElement, enabled: boolean);
```
to:
```ts
export function focusTrap(node: HTMLElement, {enabled: boolean, strict?: boolean});
```
Read the discussion: https://discord.com/channels/1003691521280856084/1003699523522142399/1113897699687796848

## Changsets

## Checklist

### DIFF
--- a/packages/skeleton/src/lib/actions/FocusTrap/focusTrap.ts
+++ b/packages/skeleton/src/lib/actions/FocusTrap/focusTrap.ts
@@ -1,11 +1,9 @@
 // Action: Focus Trap
 export function focusTrap(node: HTMLElement, enabled: boolean) {
-	const elemWhitelist = 'a[href], button, input, textarea, select, details, [tabindex]:not([tabindex="-1"])';
 	let firstFocusableChild: HTMLElement;
 	let lastFocusableChild: HTMLElement;
 
 	function keydownHandler(event: KeyboardEvent) {
-
 		// We do not care for any keys other than tab so we return early, also return if focustrap is disabled
 		// Note: All code after this statement is with 'Tab is pressed' context in mind.
 		if (event.key !== 'Tab' || !enabled) return;
@@ -14,7 +12,7 @@ export function focusTrap(node: HTMLElement, enabled: boolean) {
 		if (event.shiftKey && document.activeElement === firstFocusableChild) {
 			event.preventDefault();
 			lastFocusableChild.focus();
-		} 
+		}
 
 		// If the currently active element is the lastFocusableChild, focus the first child
 		else if (!event.shiftKey && document.activeElement === lastFocusableChild) {
@@ -30,18 +28,10 @@ export function focusTrap(node: HTMLElement, enabled: boolean) {
 	}
 	document.addEventListener('keydown', keydownHandler);
 
-	// Get first and last focusable (and therefor tabbable) child from element
-	function getFirstAndLastFocusableChild(element: HTMLElement) {
-		const focusableChilds: HTMLElement[] = Array.from(element.querySelectorAll(elemWhitelist));
-		const first = focusableChilds.at(0);
-		const last = focusableChilds.at(-1);
-		return { first, last };
-	}
-
-	function determineFocusableChilds(fromObserver: boolean)  {
+	function determineFocusableChilds(fromObserver: boolean) {
 		if (!enabled) return;
 		// Gather all focusable elements
-		const focusableElems: HTMLElement[] = Array.from(node.querySelectorAll(elemWhitelist));
+		const focusableElems: HTMLElement[] = Array.from(node.querySelectorAll(focusableElementSelector));
 		if (focusableElems.length) {
 			// Get first and last focusable child from the node
 			const { first, last } = getFirstAndLastFocusableChild(node);
@@ -51,9 +41,8 @@ export function focusTrap(node: HTMLElement, enabled: boolean) {
 			// Auto-focus first focusable element only when not called from observer
 			if (!fromObserver) firstFocusableChild.focus();
 		}
-	};
+	}
 	determineFocusableChilds(false);
-	
 
 	// When children of node are changed (added or removed)
 	const onObservationChange = (mutationRecords: MutationRecord[], observer: MutationObserver) => {
@@ -67,11 +56,21 @@ export function focusTrap(node: HTMLElement, enabled: boolean) {
 	return {
 		update(newArgs: boolean) {
 			enabled = newArgs;
-			if (newArgs) determineFocusableChilds(false)
+			if (newArgs) determineFocusableChilds(false);
 		},
 		destroy() {
 			document.removeEventListener('keydown', keydownHandler);
 			observer.disconnect();
 		}
 	};
+}
+
+const focusableElementSelector = 'a[href], button, input, textarea, select, details, [tabindex]:not([tabindex="-1"])';
+
+// Get first and last focusable (and therefor tabbable) child from element
+function getFirstAndLastFocusableChild(element: HTMLElement) {
+	const focusableChilds: HTMLElement[] = Array.from(element.querySelectorAll(focusableElementSelector));
+	const first = focusableChilds.at(0);
+	const last = focusableChilds.at(-1);
+	return { first, last };
 }
--- a/packages/skeleton/src/lib/actions/FocusTrap/focusTrap.ts
+++ b/packages/skeleton/src/lib/actions/FocusTrap/focusTrap.ts
@@ -64,7 +64,7 @@ export function focusTrap(node: HTMLElement, enabled: boolean) {
 const focusableElementSelector = 'a[href], button, input, textarea, select, details, [tabindex]:not([tabindex="-1"])';
 
 // Get first and last focusable child from element
-function getFirstAndLastFocusableChild(element: HTMLElement): { first: HTMLElement | undefined, last: HTMLElement | undefined } {
+function getFirstAndLastFocusableChild(element: HTMLElement): { first: HTMLElement | undefined; last: HTMLElement | undefined } {
 	const focusableChilds: HTMLElement[] = Array.from(element.querySelectorAll(focusableElementSelector));
 	const first = focusableChilds.at(0);
 	const last = focusableChilds.at(-1);
--- a/packages/skeleton/src/lib/actions/FocusTrap/focusTrap.ts
+++ b/packages/skeleton/src/lib/actions/FocusTrap/focusTrap.ts
@@ -73,5 +73,7 @@ function getFirstAndLastFocusableChild(element: HTMLElement): { first: HTMLEleme
 
 // Return wether a element has the data-strict-focus attribute
 function isFocusStrict(element: HTMLElement): boolean {
-	return element.dataset.strictFocus !== undefined;
+	const strictFocus = element.dataset.strictFocus;
+	if (strictFocus === '' || strictFocus?.toLocaleLowerCase() === 'true') return true;
+	return false;
 }
--- a/packages/skeleton/src/lib/actions/FocusTrap/focusTrap.ts
+++ b/packages/skeleton/src/lib/actions/FocusTrap/focusTrap.ts
@@ -17,7 +17,7 @@ export function focusTrap(node: HTMLElement, enabled: boolean) {
 			event.preventDefault();
 			firstFocusableChild.focus();
 		}
-		// If strictFocus is enabled and the activeElement is currently outside our node focus the first child
+		// If strictFocus is enabled and the activeElement is currently outside our node: focus the first child
 		else if (isFocusStrict(node) && !node.contains(document.activeElement)) {
 			event.preventDefault();
 			firstFocusableChild.focus();
@@ -60,11 +60,11 @@ export function focusTrap(node: HTMLElement, enabled: boolean) {
 	};
 }
 
-// Selector to match any focusable (and thus tabbable) elements
+// Selector that matches any focusable (and thus tabbable) elements
 const focusableElementSelector = 'a[href], button, input, textarea, select, details, [tabindex]:not([tabindex="-1"])';
 
 // Get first and last focusable child from element
-function getFirstAndLastFocusableChild(element: HTMLElement) {
+function getFirstAndLastFocusableChild(element: HTMLElement): { first: HTMLElement | undefined, last: HTMLElement | undefined } {
 	const focusableChilds: HTMLElement[] = Array.from(element.querySelectorAll(focusableElementSelector));
 	const first = focusableChilds.at(0);
 	const last = focusableChilds.at(-1);
--- a/packages/skeleton/src/lib/actions/FocusTrap/focusTrap.ts
+++ b/packages/skeleton/src/lib/actions/FocusTrap/focusTrap.ts
@@ -17,8 +17,8 @@ export function focusTrap(node: HTMLElement, enabled: boolean) {
 			event.preventDefault();
 			firstFocusableChild.focus();
 		}
-		// If the activeElement is currently outside our node focus the first child
-		else if (!node.contains(document.activeElement)) {
+		// If strictFocus is enabled and the activeElement is currently outside our node focus the first child
+		else if (isFocusStrict(node) && !node.contains(document.activeElement)) {
 			event.preventDefault();
 			firstFocusableChild.focus();
 		}
@@ -69,4 +69,9 @@ function getFirstAndLastFocusableChild(element: HTMLElement) {
 	const first = focusableChilds.at(0);
 	const last = focusableChilds.at(-1);
 	return { first, last };
+}
+
+// Return wether a element has the data-strict-focus attribute
+function isFocusStrict(element: HTMLElement): boolean {
+	return element.dataset.strictFocus !== undefined;
 }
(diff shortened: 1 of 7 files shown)


Hugos68:
> @Hugos68 This looks like a pretty big rewrite of the action. I'm going to have to set aside some time to review in the near future. Unfortunately I can't provide an ETA, but this will be asap.
> 
> For now I'll ask @Mahmoud-zino to come in as a guest and peer review if/when he can. Just try to poke holes and highlight any obvious issues we might run into.
> 
> Thanks guys.

Ofcourse, I tried my best documenting the code why and when it would run, since I can understand rewrites like this coming from another person is sometimes hard to understand. No worries it's not a breaking bug, just a minor inconvenience thus no rush required.

endigo9740:
@Hugos68 I know this is still marked in the draft state, but wanted to provide a couple bits of feedback early on:

1. You've moved `focusableElementSelector`, `getFirstAndLastFocusableChild`, and `isFocusStrict` outside the Action function. Given these are all related directly to the action I'd recommend we keep those embedded within the function. Treat it a bit like methods for a Class. Best to co-locate those.
2. As a stylistic choice we group "add listeners" statements near the bottom of the action, just above the `return` statement. This provides some consistency and makes it easy to scan and ensure every listeners added is removed in the `destroy()` method.
3. Do me a favor and add a comment near the top of the file that says "DEPRECATED: replace the `enabled` argument with `{enabled: boolean, strict?: boolean}` in v2." I'll be scanning the doc for references to  `DEPRECATED` when we make this change so this is just a helpful reminder. Technically it's not deprecated, just a temporary solution.

Given a couple folks have had their hands on this, some of this may not be from you - but if you don't mind helping us correct this while you're in progress please!

Hugos68:
> @Hugos68 I know this is still marked in the draft state, but wanted to provide a couple bits of feedback early on:
> 
>     1. The `focusableElementSelector`, `getFirstAndLastFocusableChild`, and `isFocusStrict` constant and functions are outside the Action function. Given these are all related directly to the action I'd recommend we keep those embedded within the Action please. Treat these like properties and methods for a Class. Best to co-locate those.
> 
>     2. As a stylistic choice we group `addEventListener()` statements near the bottom of the action, just above the `return` statement. This provides some consistency and makes it easy to scan and ensure every listener added is removed in the `destroy()` method.
> 
>     3. Do me a favor and add a comment near the top of the file that says "DEPRECATED: replace the `enabled` argument with `{enabled: boolean, strict?: boolean}` in v2." I'll be scanning the doc for references to  `DEPRECATED` when we make this change so this is just a helpful reminder. Technically it's not deprecated, just a temporary solution.
> 
> 
> Given a couple folks have had their hands on this, some of this may not be from you - but if you don't mind helping us correct this while you're in progress please!

1) I moved them outside because I am pretty sure they are being created now for every instance of focus trap whilst in reality is a function that can be used by any focus trap, I'll happily move it back though, that is just my way of doing stuff.

2) Sure thing, on it.

3) Will do 👍

Hugos68:
Currently 2 elements that both have action: ``use:focusTrap={true}`` and attribute: ``data-focus-strict`` will break the tabbing ability of the user causing them to be stuck. To fix this issue (because it HAS to be fixed) I have 2 suggestions in mind:
## Suggestion A:
We implement a [Stack](https://www.geeksforgeeks.org/stack-data-structure/), this will basically ensure all elements on the current page, this will work like the following:
Element A with  action: ``use:focusTrap={true}`` and attribute: ``data-focus-strict`` gets added to the DOM the stack will then look like: 
```js
const stack = [A];
```
Then element B with  action: ``use:focusTrap={true}`` and attribute: ``data-focus-strict`` gets added to the DOM the stack will then look like:
```js
const stack = [A, B];
```
Then whenever the ``isFocusStrict();`` method gets called element B will take higher precedence over element A because it was added later to the stack, therefor only element B will be strictly focused. When one of these nodes are destroyed they are removed from the stack by calling the following method in the destroy function:
``removeStrictlyElement()``

## Suggestion B:
We still implement the same stack like in suggestion A but instead of preferring one over another based on when they were added to the DOM we prefer one over another based on ``z-index``, this will look something like this:
Element A with  action: ``use:focusTrap={true}``, attribute: ``data-focus-strict`` and ``z-index: 10`` gets added to the DOM the stack will then look like: 
```js
const stack = [A];
```
Then element B with action:``use:focusTrap={true}``, attribute: ``data-focus-strict`` and ``z-index: 5`` gets added to the DOM the stack will then look like:
```js
const stack = [A, B];
```
Then whenever the ``isFocusStrict();`` method gets called element A will take higher precedence over element B because element A has a higher ``z-index`` value than element B so Element A will be strictly focused. Again when one of these nodes are destroyed they are removed from the stack by calling the following method in the destroy function:
``removeStrictlyElement()``

I'd love to hear feedback or other solutions 👍, FYI I've implemented A for now since it firstly came to mind and tabs are broken with the current PR.

Mahmoud-zino:
@Hugos68 another suggestion to consider:
join all focusable elements from both focus traps (or all visible focus traps with `data-focus-strict`) in one array and handle them as one focus trap.

Hugos68:
> @Hugos68 another suggestion to consider: join all focusable elements from both focus traps (or all visible focus traps with `data-focus-strict`) in one array and handle them as one focus trap.
Lets say we have a modal with strict focus trap and another modal on top, how would that handle the focus trap in that scenario? Because we don't want the user to be able to tab behind the top modal, but from what I understand from your suggestion we basically combine both, but would that no cause the tab to go behind the modal in front?

Mahmoud-zino:
yeah, you are correct on this one, I didn't though about two modals on top of each other but about a modal having a form inside it. there is a lot to consider here and a huge amount of edge cases to cover.

Hugos68:
> yeah, you are correct on this one, I didn't though about two modals on top of each other but about a modal having a form inside it. there is a lot to consider here and a huge amount of edge cases to cover.

I can't think of any other "normal" patterns then modal on modal, any other situation the developer shouldn't really use 2 focus traps on the same screen for the user to see, so I would say any other edge case is just a anti pattern created by the develepor.

sites/skeleton.dev/src/routes/(inner)/actions/focus-trap/+page.svelte
```diff
@@ -24,7 +24,7 @@
 	<svelte:fragment slot="sandbox">
 		<DocsPreview>
 			<svelte:fragment slot="preview">
-				<form class="w-full card p-4 text-token space-y-4" use:focusTrap={isFocused}>
+				<form class="w-full card p-4 text-token space-y-4" use:focusTrap={isFocused} data-focus-strict>
```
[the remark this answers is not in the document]
Hugos68: What do you exactly mean? I'd rather only do the functionality and Chris or you can handle the presentation for consumers if that's ok.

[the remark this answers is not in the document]
Mahmoud-zino: yeah exactly and please leave the SlideToggle in the footer slot beside the enable FocusTrap slide toggle (you can check how the other toggle is implemented to get inspiration 😉 )

packages/skeleton/src/lib/actions/FocusTrap/focusTrap.ts
```diff
@@ -1,66 +1,78 @@
...
-		}
+	// Selector that matches any focusable (and thus tabbable) elements
+	const focusableElementSelector = 'a[href], button, input, textarea, select, details, [tabindex]:not([tabindex="-1"])';
+
+	// Get first and last focusable child from element
+	function getFirstAndLastFocusableChild(element: HTMLElement): { first: HTMLElement | undefined; last: HTMLElement | undefined } {
+		const focusableChilds: HTMLElement[] = Array.from(element.querySelectorAll(focusableElementSelector));
+		const first = focusableChilds.at(0);
+		const last = focusableChilds.at(-1);
+		return { first, last };
```
Mahmoud-zino, line 15: type: focusableChildren

packages/skeleton/src/lib/actions/FocusTrap/focusTrap.ts
```diff
@@ -1,66 +1,78 @@
...
-	function onLastElemKeydown(e: KeyboardEvent): void {
-		if (!e.shiftKey && e.code === 'Tab') {
-			e.preventDefault();
-			elemFirst.focus();
-		}
+	// Return wether a element has the data-focus-strict attribute
+	function isFocusStrict(element: HTMLElement): boolean {
+		const strictFocus = element.dataset.focusStrict;
+		if (strictFocus === '' || strictFocus?.toLocaleLowerCase() === 'true') return true;
+		return false;
```
Mahmoud-zino, line 22:
```ts
return strictFocus === '' || strictFocus?.toLocaleLowerCase() === 'true';
```

packages/skeleton/src/lib/actions/FocusTrap/focusTrap.ts
```diff
@@ -1,66 +1,78 @@
...
+		return { first, last };
 	}
 
-	// When the last item selected, tab pressed, jump to the first selectable item.
-	function onLastElemKeydown(e: KeyboardEvent): void {
-		if (!e.shiftKey && e.code === 'Tab') {
-			e.preventDefault();
-			elemFirst.focus();
-		}
+	// Return wether a element has the data-focus-strict attribute
```
Mahmoud-zino, line 18: typo: // Return whether an element has the data-focus-strict attribute

packages/skeleton/src/lib/actions/FocusTrap/focusTrap.ts
```diff
@@ -1,66 +1,78 @@
...
-			elemLast = focusableElems[focusableElems.length - 1];
-			// Auto-focus first focusable element only when not called from observer
-			if (!fromObserver) elemFirst.focus();
-			// Listen for keydown on first & last element
-			elemFirst.addEventListener('keydown', onFirstElemKeydown);
-			elemLast.addEventListener('keydown', onLastElemKeydown);
+	function keydownHandler(event: KeyboardEvent) {
+		// Note: All code after this statement is with 'Tab is pressed' context in mind
+		if (event.key !== 'Tab' || !enabled) return;
+		// If the user holds shift and the currently active element is the firstFocusableChild: focus the last child
```
Mahmoud-zino, line 28: redundant comment.

packages/skeleton/src/lib/actions/FocusTrap/focusTrap.ts
```diff
@@ -1,66 +1,78 @@
...
+		if (event.shiftKey && document.activeElement === firstFocusableChild) {
+			event.preventDefault();
+			lastFocusableChild.focus();
+		}
+		// If the currently active element is the lastFocusableChild: focus the first child
```
Mahmoud-zino, line 33: redundant comment.

packages/skeleton/src/lib/actions/FocusTrap/focusTrap.ts
```diff
@@ -1,66 +1,78 @@
...
+		else if (!event.shiftKey && document.activeElement === lastFocusableChild) {
+			event.preventDefault();
+			firstFocusableChild.focus();
+		}
+		// If strictFocus is enabled and the activeElement is currently outside our node: focus the first child
```
Mahmoud-zino, line 38: redundant comment.

packages/skeleton/src/lib/actions/FocusTrap/focusTrap.ts
```diff
@@ -1,66 +1,78 @@
...
-	};
-	onScanElements(false);
-
-	function onCleanUp(): void {
-		if (elemFirst) elemFirst.removeEventListener('keydown', onFirstElemKeydown);
-		if (elemLast) elemLast.removeEventListener('keydown', onLastElemKeydown);
 	}
+	function determineFocusableChildren(fromObserver: boolean) {
+		if (!enabled) return;
+		// Get first and last focusable child from the node
```
Mahmoud-zino, line 46: redundant comment.

packages/skeleton/src/lib/actions/FocusTrap/focusTrap.ts
```diff
@@ -1,66 +1,78 @@
...
+		const { first, last } = getFirstAndLastFocusableChild(node);
+		// If there is not first (and thus last node, we return early)
```
Mahmoud-zino, line 48: redundant comment.

packages/skeleton/src/lib/actions/FocusTrap/focusTrap.ts
```diff
@@ -1,66 +1,78 @@
...
+		if (!first || !last) return;
+		// Set the new found first and last focusable childs to the variables
```
Mahmoud-zino, line 50: redundant comment.

Mahmoud-zino (commented):
## Note
These changes are just a suggestion to the coding style, please skip any change if you don't feel it fits your style 👍

packages/skeleton/src/lib/actions/FocusTrap/focusTrap.ts
```diff
@@ -1,66 +1,78 @@
...
-	function onFirstElemKeydown(e: KeyboardEvent): void {
-		if (e.shiftKey && e.code === 'Tab') {
-			e.preventDefault();
-			elemLast.focus();
-		}
+	// Selector that matches any focusable (and thus tabbable) elements
+	const focusableElementSelector = 'a[href], button, input, textarea, select, details, [tabindex]:not([tabindex="-1"])';
+
+	// Get first and last focusable child from element
+	function getFirstAndLastFocusableChild(element: HTMLElement): { first: HTMLElement | undefined; last: HTMLElement | undefined } {
```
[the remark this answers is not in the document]
Hugos68: Where should this type be defined?

packages/skeleton/src/lib/actions/FocusTrap/focusTrap.ts
Hugos68: Where should this type be defined?

Hugos68:
Current bug/not desired behaviour: When 2 elements are created at the same time and one is a child of another so:
```html
<div use:outer>
  <div use:inner>
  
  </div>
</div>
```
The inner action will fire first, this means using the stack the outer div will be the one who takes the strictly focus even though we (and most consumers) will want/expect inner to be focused strictly. This is because upon dom creation the most inner childs are created before their parents thus the action of inner will be triggered first. I'm not sure how to tackle this problem, suggestions are absolutely welcome.

Mahmoud-zino:
@Hugos68 
cause of behavior:
> The order of execution of onMount and the initial afterUpdate is that children execute before their parent. This makes sense because conceptually each child needs to be mounted before the parent can complete its mount.

I agree with you, I think we should not handle this hell of edge cases, but display a clear and easy-to-see warning that the dev should only do this once in a visible area and that we are already using it in modals so no need to do it there.

Hugos68:
> @Hugos68 cause of behavior:
> 
> > The order of execution of onMount and the initial afterUpdate is that children execute before their parent. This makes sense because conceptually each child needs to be mounted before the parent can complete its mount.
> 
> I agree with you, I think we should not handle this hell of edge cases, but display a clear and easy-to-see warning that the dev should only do this once in a visible area and that we are already using it in modals so no need to do it there.

I was thinking to completely remove the data-strict-focus thing and just leave the focus trap as is, less headache for maintainers and seems like all the other websites have the same "bug" so it's just how things work honestly

Hugos68:
Yup sounds good, this PR doesn't have any value if the inert turns out anything like how we expect it to.

endigo9740:
Cool, well again we can reference it and use it going forward. But I'm going to close it for now to help prune our PR pile a bit. Thanks Hugos!

Issue #1484, opened 2023-05-11
Investigate programatic popups


endigo9740:
### Describe what feature you'd like. Pseudo-code, mockups, or screenshots of similar solutions are encouraged!

We've heard a lot of requests for this and I believe there is a good use case for something like this. I'm not convinced that extending our current popup feature would be possible in this regard given the nature of Svelte Actions - they are typically tied to Dom manipulations. Rather this could be a brand new or sibling feature.

Idea use cases:

- Quickly generate a tooltip
- Be able to reusing the same popup iterated loops (ex: tables, app rail labels, etc)
- Triggering the popups with not direct user interaction

This will likely continue to use Floating UI as as a dependency. Long term our goal is to split features like Popups to their own dedicated library and NPM package. When this occurs we'll be able to provide tighter integration for Floating UI, it's features, and it's types. 

### What type of pull request would this be?

New Feature

saturnonearth:
Did someone say popups?

![image](https://github.com/skeletonlabs/skeleton/assets/12503538/f61f9419-90b9-4529-8f5f-a3e7d5cc0143)

A few ideas...

1.) I think a top-level singleton component (that has the ability to pass settings) that catches all elements with the "title" (native tooltip) and will automatically display a simple tooltip would really fit in-line with Skeleton's overall no-nonsense, easy to start, approach.

It also aligns with Svelte's idea of "progressive enhancement" as it only applies to elements with the default title attribute which would be a native fallback for no JS users.

2.) I think that extending the existing popup feature for actions would not be a bad idea. 

Something as simple as

```html
<button
    class="btn"
    use:tooltip={{
    content: "Hello I am a tooltip!",
}}
>
    Hover me for 1 second
</button>
```
(with some pre-defined settings, or defaults)

Seems like a relatively simple way to allow people to provide a one-off tooltip for something they need, if they don't want to opt-in to all tooltips.

3.) Like the second option + the first,

```html
<button
    class="btn"
    use:tooltip
    title="Hello I am a tooltip!"
>
    Hover me for 1 second
</button>
```

An action that uses the title as a reference for the content.

endigo9740:
Heck yeah, you know I love singletons :D Not a bad idea at face value. I'm not sure my ETA for being able to revisit this, but I'll noodle on this a bit.

jwatts777:
+1
Ex popup.close(), popup.open()
Willing to do the legwork if you have a specific implementation that you want to follow.

[...]

PR #1616, opened 2023-05-31
Add Optional Open/Close Functions

### LINKED ISSUE #1615: Allow for Passing Custom Functions to Open/Close event listeners on popups

The title pretty much describes the request.
I want to have make a `hover` popup that does not disappear when pressing shift. Currently it's not possible as the element has a `mouseleave` event listener telling the popup to be closed.

Example of what I want to achieve (the "window" is done by pressing Shift):

https://github.com/skeletonlabs/skeleton/assets/32039708/507c002e-d448-40bf-831c-7005b5c67211

Allowing to the user to pass their own open and close functions would allow for such additional functionality.

Enhancement

There was once a conversation about attempting to create this sort of hover-to-window functionality, but it ended up going nowhere.

MrVauxs:
## Linked Issue

Closes #1615

## Description

A (definitely haphazard) implementation of the feature request. I have no idea of TypeScript.

## Changsets

khaki-olives-relax.md

## Checklist

- [ ] All test cases are passing - run `pnpm test` (errors with no test selected)

### DIFF
--- a/.changeset/khaki-olives-relax.md
+++ b/.changeset/khaki-olives-relax.md
@@ -1,0 +1,5 @@
+---
+"@skeletonlabs/skeleton": minor
+---
+
+Added optional open and close functions to popup event listeners
--- a/packages/skeleton/src/lib/utilities/Popup/popup.ts
+++ b/packages/skeleton/src/lib/utilities/Popup/popup.ts
@@ -186,11 +186,11 @@ export function popup(triggerNode: HTMLElement, args: PopupSettings) {
 			break;
 		case 'hover':
 			triggerNode.addEventListener('mouseover', args.open ? (event) => args.open?.(event, open) : open, true);
-			triggerNode.addEventListener('mouseleave', args.close ? (event) => args.close?.(event, close) : (() => close()), true);
+			triggerNode.addEventListener('mouseleave', args.close ? (event) => args.close?.(event, close) : () => close(), true);
 			break;
 		case 'focus-blur':
 			triggerNode.addEventListener('focus', args.open ? (event) => args.open?.(event, toggle) : toggle, true);
-			triggerNode.addEventListener('blur', args.close ? (event) => args.close?.(event, close) : (() => close()), true);
+			triggerNode.addEventListener('blur', args.close ? (event) => args.close?.(event, close) : () => close(), true);
 			break;
 		case 'focus-click':
 			triggerNode.addEventListener('focus', args.open ? (event) => args.open?.(event, open) : open, true);
--- a/packages/skeleton/src/lib/utilities/Popup/popup.ts
+++ b/packages/skeleton/src/lib/utilities/Popup/popup.ts
@@ -181,20 +181,20 @@ export function popup(triggerNode: HTMLElement, args: PopupSettings) {
 	// Event Listeners
 	switch (args.event) {
 		case 'click':
-			triggerNode.addEventListener('click', toggle, true);
-			window.addEventListener('click', onWindowClick, true);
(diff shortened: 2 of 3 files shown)


endigo9740:
@MrVauxs we have a policy for contributions on Skeleton that non-trivial updates require prior discussion with maintainers. The idea being that we can prevent folks spending a lot of time on features we cannot accept.

Items number 4 found here:
https://www.skeleton.dev/docs/contributing/requirements#pull-requests

Had we had prior discussion I would have been able to point out this functionality already exists in the popup feature - via the `state` param. It's a callback function that you can provide and will trigger any time an `open` or `close` event occurs for a particular popup and the current state will be reported in the response.

Given the amount of effort you've devoted here, I'll give you a chance to explain the benefit of this over what's already in place. Though typically we would recommend extending an extending features than the providing redundant options.

endigo9740:
@MrVauxs a bit more context as well if you check our roadmap:
https://github.com/skeletonlabs/skeleton/discussions/1367

We're in a transitionary phase where we've just migrated the project to a monorepo. As part of this we're looking to actually spin off the Popup feature to it's own standalone package. When we do this, we're going to likely rewrite or recreate the feature in a way that will more tightly integrate with Floating UI.

Floating UI will actually become a peer dependency for these package, which opens a lot of doors for us. A few brief examples of what that might include:

- Better type support through Floating UI's own provided types
- Simpler install process, given we can make this a peer dependency that NPM will auto-install
- No more need to pass the module import through a store to our action
- Plus we'll likely introduce new ways to use Floating UI - ex: a tap and hold context menu for mobile

Needless to say there's a lot of changes coming down the pipeline for this feature. But this is still a ways off. That said, we are field all input between now and then, so that's part of why I request your thoughts behind the request and change. Perhaps it's something we can consider for the future?

MrVauxs:
Thanks.

The feature I am suggesting directly ties into the `open` and `close` event listeners so users can decide **when** the popups are closed and opened. This is not something that can be done with `state`, as you cannot modify whether the popup has been actually opened and closed with it. You are simply notified that it had been.

endigo9740:
> The feature I am suggesting directly ties into the open and close event listeners so users can decide when the popups are closed and opened. This is not something that can be done with state, as you cannot modify whether the popup has been actually opened and closed with it. You are simply notified that it had been, but cannot influence it.

I follow, so it's programmatically triggering a open or close state, not listening for those. Hmm, that does seem useful.

I think this falls under a duplicate to:
- https://github.com/skeletonlabs/skeleton/issues/1484

This is an idea we're very much interested in exploring, but I'm leery to introduce any other additional major features to the popup in the near future given the proposed path  going forward. So let me think about this one for a bit. We'll keep it under consideration.

MrVauxs:
> I think this falls under a duplicate to:
> 
> * [Investigate programatic popups #1484](https://github.com/skeletonlabs/skeleton/issues/1484)

I understand. It was on my radar but this is less about being able to programmatically create popups as much as what amounts to creating custom `events` basically.

But sure, I understand. Not sure if I should devote my time looking for a workaround or just focus on other things in the meantime...

jwatts777:
@endigo9740 Funnily enough, this was exactly what I was looking for. I wanted custom control over the popup's state(open,close).
Nice work @MrVauxs. You just saved me 4+ hours. 
#1484 closing

endigo9740:
@MrVauxs I've decided this is something we're interested in exploring, but won't be part of the current implementation. We'll wait until we migrate popups to a standalone tool to revisit. I've provided a note to reference this PR when we get to that point:
- https://github.com/skeletonlabs/skeleton/issues/1484#issuecomment-1587857706

For now I'm going to close this, but again, we'll use it as reference in the future. Thanks for the contribution and sorry if it takes a bit longer to get in place.

PR #1629, opened 2023-06-03
replaced unicode arrows with template SVGs

### LINKED ISSUE #1557: Bug: unicode icons within button icons are not centered properly
### Current Behavior

The first/last icons are not centered in the button itself, as can be seen in the following picture:
Screenshot 2023-05-24 at 17 41 18

This is not a major problem, since the `buttonTextFirst` and `buttonTextLast` props can be set to override this.

However, it could maybe be nice to have a (close to) pixel perfect look out of the box!

Right now, the following HTML-encoded characters are used:
```typescript
/** Set the text label for the Previous button. */
export let buttonTextPrevious: CssClasses = '&larr;';
/** Set the text label for the Next button. */
export let buttonTextNext: CssClasses = '&rarr;';
/** Set the text label for the First button. */
export let buttonTextFirst: CssClasses = '&laquo;';
/** Set the text label for the Last button. */
export let buttonTextLast: CssClasses = '&raquo;';
```

I tried to find better alternatives but without luck. So maybe additional CSS could be needed? Or some SVG embeds? Not sure!

Just use the Paginator component with the default values and you should spot the problem.

Tested on Safari and Chrome

### Anything else?

_No response_

Mahmoud-zino:
## Linked Issue

Closes #1557

## Description

replaced Unicode arrows with templated SVGs.

## Note
[chevrons-left](https://fontawesome.com/icons/chevrons-left?f=classic&s=solid) & [chevrons-right](https://fontawesome.com/icons/chevrons-right?f=classic&s=solid) Icons are pro icons so I ended up using [angles-right](https://fontawesome.com/icons/angles-right?f=classic&s=solid) and [angles-left](https://fontawesome.com/icons/angles-left?f=classic&s=solid) Icons which looks the same.

## Changsets

bugfix: replaced unicode arrows with template SVGs in paginator

## Checklist


endigo9740:
@Mahmoud-zino those actually look really sharp - one addition note, we'll need to adjust for not just text color, but fill color as well when we implement the SVG so that these are still visible:

image

Mahmoud-zino:
@endigo9740 I added the `fill-current` to the `buttonClasses` so if the user wants to change it they can do it easily.
If this is not correct please tell me and I will change it.

packages/skeleton/src/lib/components/Paginator/Paginator.svelte
```diff
@@ -150,7 +163,7 @@
 		{#if showFirstLastButtons}
 			<button
 				type="button"
-				class={buttonClasses}
+				class="{buttonClasses} 0"
```
endigo9740, line 166: Was this a typo?

Mahmoud-zino: ops, yeah sorry this was a typo, I will fix it.

PR #1637, opened 2023-06-06
Feat: toast enhancement

### LINKED ISSUE #1311: Toast wishlist
### Toast enhancement wishlist

I wanted to keep a list of some things I would like to see for Toasts. I may implement some of these myself, but I thought it would be convenient to keep a list around in case anyone else wants to tackle some of these.

- [x] Animate on destroy
- [x] Smooth transitions when toasts are being shifted around
 (see  #1312) 
- [x] Ability to hide X (close) button
- [x] Remain visible if hovering
- [ ] Ability for rich content via targeting of html node or ID.
- [x] Ability to programmatically close specific toasts by reference
- [ ] Ability to pass promise to Toast to programmatically close when promise completes or fails.
   - This one is a bit more complex and may not need to be implemented.
   
chrome_0YIMRIqh9g

Enhancement

Mahmoud-zino:
## Linked Issue

Closes #1311

### Done
1. ability to hide the x button using the prop `hideDismiss`.
2. ability to remain visible if hovering using the prop `hoverVisible`.
3. ability to close toast programmatically using the toastId.
### Notes
1. I was worried that the user could activate `hideDismiss` and the toast will stay on screen infinitely. so I have set autohide to true automatically when `hideDismiss` is set and wrote a note in the docs about it.
2. Should I add some hover effect when `hoverVisible` is set so the user knows that hovering the toast will keep it open?
3. remaining visible when hovering is simpler when we just delete the timeout and recreate it after, but the time spent on the toast is not calculated, so should I track the time and kind of continue it after `mouseLeave` or is it ok to just reset the timer?
4. to close the toast programmatically I am returning the ID from the store `trigger` function, I don't think this change should be a breaking change, but I just wanted to make sure.

## Changsets

feat: Toast enhancement - ability to `hide the close button`, ability to `remain visible on hover`, ability to `close toasts programmatically` by Id.

## Checklist

### DIFF
--- a/.changeset/lazy-pants-yawn.md
+++ b/.changeset/lazy-pants-yawn.md
@@ -1,0 +1,8 @@
+---
+"@skeletonlabs/skeleton": minor
+---
+
+feat: Toast enhancement
+    1. Added the ability to hide the close button.
+    2. Added the ability to remain visible on hover.
+    3. Added the ability to close toasts programmatically by Id.
--- a/.changeset/lazy-pants-yawn.md
+++ b/.changeset/lazy-pants-yawn.md
@@ -2,7 +2,4 @@
 "@skeletonlabs/skeleton": minor
 ---
 
-feat: Toast enhancement
-    1. Added the ability to hide the close button.
-    2. Added the ability to remain visible on hover.
-    3. Added the ability to close toasts programmatically by Id.
+feat: Toast enhancement - ability to `hide the close button`, ability to `remain visible on hover`, ability to `close toasts programmatically` by Id.
--- a/.changeset/lazy-pants-yawn.md
+++ b/.changeset/lazy-pants-yawn.md
@@ -2,4 +2,4 @@
 "@skeletonlabs/skeleton": minor
 ---
 
-feat: Toast enhancement - ability to `hide the close button`, ability to `remain visible on hover`, ability to `close toasts programmatically` by Id.
+feat: Expanded Toast features to allow hiding the dismiss button, allow toast to remain visible on hover, and programatically close each instance.
--- a/cspell.json
+++ b/cspell.json
@@ -125,6 +125,7 @@
 		"Themer",
 		"tickmarks",
 		"Transpiles",
+		"undismissible",
 		"unsub",
 		"Unsubscriber",
 		"Vercel",
--- a/packages/skeleton/src/lib/utilities/Toast/Toast.svelte
+++ b/packages/skeleton/src/lib/utilities/Toast/Toast.svelte
@@ -76,6 +76,14 @@
 		toastStore.close($toastStore[index].id);
 	}
 
+	function onMouseEnter(index: number): void {
+		if ($toastStore[index]?.hoverVisible) toastStore.freeze(index);
+	}
+
+	function onMouseLeave(index: number): void {
+		if ($toastStore[index]?.hoverVisible) toastStore.unfreeze(index);
+	}
+
 	// Reactive
 	$: classesWrapper = `${cWrapper} ${cPosition} ${zIndex} ${$$props.class || ''}`;
 	$: classesSnackbar = `${cSnackbar} ${cAlign} ${padding}`;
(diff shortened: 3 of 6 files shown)


saturnonearth:
wow you literally went through every change in my list and knocked it out of the park!

> ability to close toast programmatically using the toastId.

If someone wanted to use this, Is there a way to grab the ID from somewhere other than the callback?

> Should I add some hover effect when hoverVisible is set so the user knows that hovering the toast will keep it open?

I don't think it's that needed but maybe a slight transform to make it slightly bigger on hover?

> remaining visible when hovering is simpler when we just delete the timeout and recreate it after, but the time spent on the toast is not calculated, so should I track the time and kind of continue it after mouseLeave or is it ok to just reset the timer?

I think resetting it is fine.

endigo9740:
@Mahmoud-zino I'll circle back to this soon - but make sure you adjust the Changeset to be a paragraph style. This is the data that'll appear on the generated changelog, so if it's an odd format it may not come through as desired. We should have our first look at the changelog format in the next hour or so, so keep an eye out for that.

Mahmoud-zino:
@saturnonearth 
> If someone wanted to use this, Is there a way to grab the ID from somewhere other than the callback?

Is there a specific use case you could use? I thought it only makes sense to close them when you trigger them -> you grab the Id from the callback, but if there is another use case I would be happy to implement it.

saturnonearth:
> @saturnonearth
> 
> > If someone wanted to use this, Is there a way to grab the ID from somewhere other than the callback?
> 
> Is there a specific use case you could use? I thought it only makes sense to close them when you trigger them -> you grab the Id from the callback, but if there is another use case I would be happy to implement it.

 I guess I just wondered if you could set a custom ID yourself instead of having to grab it from the callback.

endigo9740:
Hey @Mahmoud-zino, the logic for these new features looks sound. The only issues I have are in relation to param naming and the documentation presentation. If you can implement the following I'd be happy to merge this:

Param changes:

- Change `hideDismiss: true` -> `dismiss: false` (with `true` being the default)
- Change `hoverVisible` -> `hoverable: true` (with `false` being the default)

Doc Changes:

- Change the button text for the hover option to "Visible on Hover" (with caps as shown)
- For the hover example -> code tab, let's reword the warning as follows:
> When using the `dismiss` option the `autohide` feature will be enabled by default. This is required.
- Let's split the Programatic example to a head heading section directly after Trigger/Clear called "Programatic", add the text I supply below for the description, then show your exaple with buttons labeled "Show" and "Close".
> Create a reference ID for your toast so that you may programatically close the toast on demand.

Additionally you may want to modify the changeset to more accurate represent the changes above. Remember the changeset should only describe the change, not provide the implementation details. That's where the PR link and documention comes into play, for users that want to look at the specifics. I'd recommend something like:

> feat: Expanded Toast features to allow hiding the dismiss button, allow toast to remain visible on hover, and programatically close each instance

Mahmoud-zino:
@endigo9740 
> Change hideDismiss: true -> dismissable: false (with true being the default)

from what I know, interfaces can't have a default value, do you have another suggestion for the name (opposite of dismissable, I was thinking undismissable but it sounds a little bit odd)?

Mahmoud-zino:
@endigo9740 
I applied the changes you suggested and changed the hideDismiss to undismissable for now, but I will be happy to change it once you have a better idea (I searched a lot with no luck).
Also you might want to look at this comment regarding the node feature: https://github.com/skeletonlabs/skeleton/issues/1311#issuecomment-1579237845

endigo9740:
@Mahmoud-zino unfortunate `undismissable` is not a word, so we really can't use that. I'd suggest either flipping back to `hideDismiss: true` or `dismiss: false`. I did find find one reference using the latter:

https://github.com/mzohaibqc/svelte-toasts

(see the readme)

I agree flipping between positive/negative boolean values feels weird, but as long as it's documented it should work.

Mahmoud-zino:
@endigo9740 The problem is we can't change the default value of the prop (because we are using an interface) so I don't understand what you meant by flipping back to "hideDismiss: true" (also in PR it was false) if we could somehow have true as default value I would make sure to document it but I don't know if it is possible.

endigo9740:
@Mahmoud-zino again the default state is implicit rather than explicit within the code. If a value like `hideDismiss` is not 
provided then the toast will behave in it's "default" manner.

- Default: dismiss button is present
- `hideDismiss: true` provided - dismiss button is no longer provided

This will be our short term implementation for this feature. But when v2 rolls around we might want to swap the defaults to be like this:

- Default: dismiss button is not present
- `dismiss: true` - dismiss button is provided

However, this would be a breaking change, so we cannot implement this right now. But the semantics and state of `dismiss: true` is more inline with expectation for configurable options.

Passing `false` is typically not the advisable way to word and arrange this. But we're limited to what we can do within the constrains of how this operates right now, we cannot implement a breaking change.

Mahmoud-zino:
@endigo9740 aww now I get you, yeah sorry for misunderstanding that like 3 times (Monday effect 😶)

It makes sense, I will make the change 👍.

endigo9740:
@Mahmoud-zino I fixed a small style issue that was providing a gap on the right side of the toast when the dismiss option is hidden. Likewise the docs felt like a ton of examples without context, so I've adjusted this to provide more information about each type of setting. I think with this we're good to merge!

PR #1645, opened 2023-06-07
feat: added anchor tab

### LINKED ISSUE #1583: Using tabs for navigation

Currently to use navigation you have to either surround a tab with ``<a>`` tags or use ``goto()`` both comes with their own problems. The ``<a>`` solution will cause users to firstly tab onto the ``<a>`` and then the ``<div>`` with ``role="tab"``, another problem with this approach is that now the direct children of the tabgroup are not elements with the ``role="tab"`` attribute. The ``goto()`` method is a lot better but still not perfect since ``non-js`` users are now unable to visit certain parts of my website.

A suggested solution is the following, although it is far from fully thought out and sounds honestly a bit too complex for no reason:
Pass in a prop to ``TabGroup`` element called ``navigation`` either set to true or false (false by default) if this navigation property is set the tabs will be ``<a>'s`` instead of ``<div>'s``, the value of the tab's will now be passed into the ``href`` of that tab.

Enhancement

_No response_

Mahmoud-zino:
## Linked Issue

Closes #1583

## Description

Added Anchor tab to TabGroup

## Changsets

feat: Added anchor tab to TabGroup.

## Checklist

### DIFF
--- a/.changeset/itchy-hounds-unite.md
+++ b/.changeset/itchy-hounds-unite.md
@@ -1,0 +1,5 @@
+---
+"@skeletonlabs/skeleton": minor
+---
+
+feat: Added anchor tab to TabGroup.
--- a/packages/skeleton/src/lib/components/Tab/Tab.svelte
+++ b/packages/skeleton/src/lib/components/Tab/Tab.svelte
@@ -25,6 +25,8 @@
 	 * @type {any}
 	 * */
 	export let value: any;
+	/** Provide a hoverable title attribute for the tab. */
+	export let title = '';
 
 	// Props (a11y)
 	/** Set the ARIA controls value to define which panel this tab controls. */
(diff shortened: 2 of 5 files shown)


endigo9740:
@Mahmoud-zino I've rewritten a portion of the Tab documentation page to account for this new anchor component. Unlike App Rail, I think the idea of mixing and matching Tab/TabAnchor would actually be harmful to the UX. There's no difference in placement like the rail (lead/trail slots in the bar) which means there's no visual hierarchy. One tab would change the state and anchor would take you somewhere completely different!

Instead the docs now recommend the opposite, go all in on one type or the other. I've also updated the example to make use of local anchor links, not external. While users certainly can make tab links open externally, again, I think this will be bad practice as tabs don't really operate this way. They're meant to swap between a local set of content.

The ideal use case for TabAnchor is having a tab fixed as part of your layout, and swapping between several tabbed routes within your app. So I've updated the documentation and examples to indicate this.

Mahmoud-zino:
@endigo9740 You are absolutely correct, I don't know why I handled Tabs just like the App rail 😅.

Note: I fixed a mini typo in the docs.

and It looks good, I am for sure happy with the changes 👍

PR #1648, opened 2023-06-07
chore: update dependencies

AdrianGonz97:
## Description

We're a bit behind with a fair number of our dependencies, resulting in some of the security warnings that we get for the docs site. So I've gone ahead and incrementally updated everything that I can.

I also updated TS to `5.0.3` even though `5.1.3` has been [officially released](https://devblogs.microsoft.com/typescript/announcing-typescript-5-1/). It seems that `@typescript-eslint` needs a bit of time to catch up to the latest release, so once they do, I'll aim to update TS to the latest.

Upgrading to TS 5.0 breaks `vite-plugin-sveld` since they depend on `svelte-preprocess@4.X.X` which doesn't support TS 5.0 and up. I've gone ahead and added the plugin locally so that we can control that independently.

Edit: I've also gone ahead and included updating our dev deps for the `@skeletonlabs/skeleton` package as well.

## Changsets

Instructions: Changesets automate our changelog. If you modify files in `/package/skeleton`, run `pnpm changeset` in the root of the monorepo, follow the prompts, then commit the markdown file. Changes that add features should be `minor` while chores and bugfixes should be `patch`. Please prefix the changeset message with `feat:`, `bugfix:` or `chore:`.

## Checklist


endigo9740:
@AdrianGonz97 so now that we have the library and documentation split, it may be worth reviewing which project needs which dependency. I'm guessing you and Nik just cloned the same set for both projects. However, there's some that aren't needed in both - such as Floating UI. That's really only needed in the docs for now, we don't directly access any part of this from our popups.

endigo9740:
@AdrianGonz97 @niktek I've set Nik as a reviewer on this. If he gives the thumbs up I think it's good to merge. We can dogfood this as part of this current release cycle over the next couple weeks.

niktek (approved):
Looks good to me!

PR #1651, opened 2023-06-08
Docs: Scroll Containers + Carousel

### LINKED ISSUE #952: Carousel component

Hello,
Can we expect a carousel component to slide through multiple elements and images using custom controls, indicators, intervals, and options ?

New Feature

https://daisyui.com/components/carousel

https://flowbite-svelte.com/components/carousel/

Mahmoud-zino:
## Linked Issue

Closes #952

## Description

Added `Scroll Containers` TailwindCss Block.

## Changsets

feat: Added `Scroll Containers` TailwindCss Block.

## Checklist


Mahmoud-zino:
@endigo9740 I just found out, if we scroll using js, the CSS snapping feature is redundant...
If you look at the implementation and delete all snap-related tailwind classes the example will still work as it is.
I don't know if we should go out of our way to use the snapping feature, and if not, where should the carousel be implemented? > tailwind > blocks > carousel doesn't make sense when using js to scroll 😃.

endigo9740:
@Mahmoud-zino I know we could handle this with JS, but JS is always going to be less performant and more costly than CSS. The CSS snapping is also better for mobile touch as you can literally drag and snap each slide. The only reason we want to introduce JS is if we have to. If we can't control the thumbnails with the CSS snap, etc.

If this is giving you trouble let me know and I can circle back on it soon.

Mahmoud-zino:
@endigo9740 So I did change the scrolling to CSS using anchors.
Here are some notes I might use some help on:
1. when scrolling using the anchor the image is scrolled to the top of the page (which is normal behavior for other use cases but not carousel), can we avoid or change this behavior?
2. Can we somehow mark the active anchor, so the user knows it is active? I tried the `active:` class but it's not what we want.
3. I feel the info I included in the usage slot is somewhat not enough but I don't know what to write so If you have any Idea, it would be awesome.

endigo9740:
@Mahmoud-zino I'm going to set this back to the draft state until the revisions we discussed on Discord are in place. It'll help me identify which PRs need active review. No rush, just any time. Thanks!

Mahmoud-zino:
@endigo9740 I did the changes we discussed earlier.

## Notes
1. I have added empty containers in the main demo but you may have a better idea.
2. I used as little JS as possible to make the carousel fully functional... but maybe you have some improvements to the code.
3. I was thinking about adding a shopping / Product carousel example but in the end, it is still just a carousel with extra buttons, so I figured out to ask you first.

endigo9740:
@Mahmoud-zino so a couple things here:

1. The featured example needs a few adjustments:

- It should use the default gradient background to signify it's the featured demo
- It's not immediately obvious what it's showcasing. We either need some instruction like "Scroll left or right to see snapping effect".

2. I think we're starting to his some kind of rate limit from Unsplash. Maybe due to the number of images. I'm seeing this when the page loads and it persists for upwards of 30 seconds

Screenshot 2023-06-21 at 11 19 12 AM

Until I can get setup with proper API access - I'd say let limit the use of their images purely to the carousel example.

3. Per the shopping product carousel - I like the idea, but this might be overly complex for showcasing the concept. What do you think about replacing it with a movie poster style carousel instead. It should still contain three columns per section, but maybe we use a different source for the images. I'm thinking movies, with a simple poster image.

Screenshot 2023-06-21 at 11 28 17 AM

We can use a service like this to get the images. Something like Superhero genre movies should be safe and friendly:
https://www.themoviedb.org/keyword/9715-superhero/movie

Geoff Rich created something similar and he was able to embed the images directly:
https://github.com/geoffrich/star-wars-demo-svelte/blob/shared-element-transitions/src/movies.json

endigo9740:
@Mahmoud-zino make sure to save your shopping product markup for later too - I do think we'll be able to find some use for that. Perhaps as it's own dedicated Block! :)

Mahmoud-zino:
@endigo9740 
sorry I misunderstood what you meant earlier, I have implemented the movies carousel as you mentioned above.

Note: I couldn't use scale on hover as it would mess with the sizes and alignments so I changed the filled color on hover instead.

endigo9740:
@Mahmoud-zino I went down a bit of a rabbit hole here, but ended up making quite a few changes:

- Added a Tailwind Documentation reference table
- Featured Demo:
    - Adjusted the styling
    - Added padding between the items and scrollbar
    - Reverted to the default card styles
    - Used an Each loop for less repeated code
    - Implemented horizontal scroll padding
    - Add scroll alignment selection
- Carousel:
    - Refactored the styling, simplifying it a bit
    - Remove the scroll container for thumbnails - again let's keep it simple
    - Used loops to pair the featured and thumbnail images
    - Split the next/previous functions. This should be easier for users to digest.
    - Refactored the logic
    - Updated the code snippet to be smaller code blocks
- Multi-Column:
    - Refactored the styling to simplify
    - Refactored the logic into smaller functions
    - Updated the code snippet
    - Converted `movies.json` to a TS file with basic type safety
    
The overall goal is to teach the concept and then let users run with that. So in this in mind I've tried to boil down each example to it's simplest version. Including more bite size code snippets.

Let me know what you think!

Mahmoud-zino:
@endigo9740 
The bug was (like you said) a rounding problem, different browsers use different rounding methods (floor, ceiling, etc..) which causes this odd behavior.
```ts
// -1 is used because different browsers use different methods to round scrollWidth pixels.
if (elemMovies.scrollLeft < elemMovies.scrollWidth - elemMovies.clientWidth - 1) x = elemMovies.scrollLeft + elemMovies.clientWidth;
```
doing this will solve it independently from the browser method -_-

endigo9740:
Awesome @Mahmoud-zino, I think with that we're good. Merging this. It'll be part of the next release!

PR #1654, opened 2023-06-09
added Rating-Bar component

### LINKED ISSUE #1028: Rating Component

I could really use a decent rating component for a few of my projects. Would be nice to have option for color changes per "star", number labels. length option ( 5 stars or 10 for example) Icons like hearts or thumbs up as well as star. Tool tip descriptions. Half selections like 3.5 etc. 

Thank You.

screenshot-quasar dev-2023 02 23-08_48_14

New Feature

https://quasar.dev/vue-components/rating

more basic one:
https://primeng.org/rating

Mahmoud-zino:
## Linked Issue

Closes #1028

## Description

Added RatingBar component.

### Notes:
1. As Discussed on Discord, for now, the `RatingBar` is simply a `Presentation` component.
2. I need some Ideas for the `usage` slot in the docs 🙏 .

## Changsets

feat: Added svelte `RatingBar` component.

## Checklist


endigo9740:
@Mahmoud-zino I'll do a deeper review soon, but real quick:

1. Rename the component `<Ratings />` and the route `/ratings`
2. Change the default SVG fill color to `fill-on-primary-token` for a more neutral and accessible style
3. Numeric slot names make me nervous, so let's do `empty | half | full`

I'll review the code and such in a follow up!

jamess-ais (approved):

jamess-ais:
Whoops sorry, was viewing and saw there was a review button and was curious if I had the perms... turns out I do!

Ignore my approval!

endigo9740:
@Mahmoud-zino I've pushed a few changes and improvements to the doc and component itself. This includes:

- Flushed out the doc with more examples
- Added interactive binding for the current/max values - which is fun to play with!
- Updated the local imported icons file to be a bit more organized, include circles, moved attribution
- SVG image size is now responsive, so we can lead by example
- Update the component to include `fill` and `text` style props, with default values.
- Minor comment updates to match conventions

endigo9740:
@Chad_GPT#1355 couple follow up thoughts...

1. Should we maybe go ahead and wrap each icon with a span and add `regionIcon` classes so users can style these? Users could do something like `regionIcon="cursor-pointer"` if they want these to be interactive
2. Per interactivity - maybe when clicking one of these divs we dispatch an event that displays the Each loop index value? They'll already know how many max items there are, so if they get an index of:
    - 0 = first item
    - 4 = middle item (out of 10)
    - 9 = middle item (out of 10)
    
These two small changes would allow users to make these somewhat interactive on click. It wouldn't handle half-states, but that's probably ok for an initial version like this.

Mahmoud-zino:
@endigo9740 

### Some Notes
1. I wrapped the icons with buttons because adding `on:click` on span will enrage A11y, the down side is without interactivity the users will still get the cursor-pointer and the ring around the button.
2. I added a `User Interactivity` section to the docs with only the code on how to make the component user interactive. Should I add more details?

endigo9740:
@Mahmoud-zino this is perfect, one more thing with the buttons - make sure you always apply a `type`, in this case `type="button"`. If you don't the buttons will default to `type="submit"`, which can interfere with form functionality. If you can get that in by tomorrow I'll review and we'll likely go ahead and merge.

Super excited to include this, as it's a long awaited feature request that's sat idle for quite some time! Great job!

Mahmoud-zino:
@endigo9740 
Yeah, totally correct, I forgot about it.
I changed the type to button but didn't add a prop because I don't think you would ever consider adding the rating click as a submit for your form.

PR #1679, opened 2023-06-20
Bugfix: double drawer unexpected closing direction

### LINKED ISSUE #1671: Unexpected Drawer close animation with multiple Drawers
### Current Behavior

When using multiple Drawers, they slide out in unexpected directions when closing the last one

### Expected Behavior

When using multiple Drawers, they should slide out from where they came from

### Steps To Reproduce

Using the attached StackBlitz, press `l` -> `l` or `r` -> `r` to open and close a Drawer. Then it slides out and in as expected.

Using the attached StackBlitz, press for example `l` -> `r` -> `r` to first open the left Drawer, then close it and open the right Drawer and finally close the right Drawer as well. Notice how the right Drawer slides out in the wrong direction!

https://stackblitz.com/edit/skeletonlabs-repl-agdvvz

### More Information

https://github.com/skeletonlabs/skeleton/assets/31876997/b8813832-8f24-441c-ad48-cdecd1e1d56f

Mahmoud-zino:
## Linked Issue

Closes #1671

## Description

so by setting the transition like this:
```ts
transition:fly|local={{ x: anim.x, y: anim.y, duration }}
```
apparently svelte will save the params values and use them for `out` transition, so updating these values won't update the transition.

to solve this, I replaced the transition with in/out, so Svelte would check the value of the params before going out. which solves the problem.

Note: this is just my understanding of the issue, I didn't find any official docs that tackle this problem or indicate how Svelte handles the transitions internally, so my interpretation could be awfully wrong 😄.

## Changsets

bugfix: Fixed double drawers unexpected closing direction.

## Checklist

### DIFF
--- a/.changeset/fast-apricots-sing.md
+++ b/.changeset/fast-apricots-sing.md
@@ -1,0 +1,5 @@
+---
+"@skeletonlabs/skeleton": patch
+---
+
+bugfix: Fixed double drawers unexpected closing direction.
--- a/packages/skeleton/src/lib/utilities/Drawer/Drawer.svelte
+++ b/packages/skeleton/src/lib/utilities/Drawer/Drawer.svelte
@@ -177,6 +177,7 @@
 		use:focusTrap={true}
 	>
 		<!-- Drawer -->
+		<!-- separate In/Out so anim values update -->
 		<div
 			bind:this={elemDrawer}
 			class="drawer {classesDrawer}"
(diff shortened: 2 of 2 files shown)


endigo9740:
@Mahmoud-zino it's a fairly non-standard edge case, so we'll give this a go. Just a reminder that this will likely cause a merge conflict with the transitions PR - so make sure to update when you can.

PR #1686, opened 2023-06-20
bugfix: interactive ratings icons on demand

### LINKED ISSUE #1685: Ratings issue buttons and interactive type
Currently we've opted for `<button>` elements to support interactive mode. However this has an issue when not using interaction in that a pointer cursor will still appear - as if the icons are clickable.

To resolve this we should modify either the template, or the styles, to remove the pointer styling.

We should add a new `interactive` prop so that we know when interactive mode is activated. Then:

- Option 1: use `<svelte:element>` to modify the element - but then we have to juggle the `type="button"`
- Option 2: use CSS to disable the pointer events

Option 1 would use more semantically correct HTML, while Option 2 is simpler.

Mahmoud-zino:
## Linked Issue

Closes #1685

## Description

Added iconType property to make the icons interactive on demand.

## Changsets

bugfix: `Ratings` icons are only interactive when `iconType` is set to `button`.

## Checklist

### DIFF
--- a/.changeset/young-cameras-sparkle.md
+++ b/.changeset/young-cameras-sparkle.md
@@ -1,0 +1,5 @@
+---
+"skeleton.dev": patch
+---
+
+bugfix: `Ratings` icons are only interactive when `iconType` is set to `button`.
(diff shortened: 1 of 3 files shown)


endigo9740:
@Mahmoud-zino I went ahead and transitions this over to using the `interactive` prop to be a bit more of a declarative mode. This might have multiple purposes in the future, but is more explicit to what is being done rather than setting the element type. You might look at how I implemented the dynamic attribute type - this is a bit of workaround for adding `type` to `<svelte:element>`. It's kind of handy!

PR #1706, opened 2023-06-29
Feat/table of contents refactor

### LINKED ISSUE #1030: Table of Contents refactor
As part of the recent Doc Shell update we've identified several issues with the [Table of Contents](https://www.skeleton.dev/components/table-of-contents) component:

- [ ] We should abstract the permalink/header id generator away from this component
- [ ] This component should become a "dumb" interface that just reacts to either manually or auto-assigned values
- [ ] https://github.com/skeletonlabs/skeleton/issues/1589
- [x] https://github.com/skeletonlabs/skeleton/issues/1143
- [x] https://github.com/skeletonlabs/skeleton/issues/1156

We should look to refactor this component to work in a more consistent manner.

Mahmoud-zino:
## Linked Issue

Closes #1030

## Description

TODO

## Changsets

TODO

## Checklist


endigo9740:
@Mahmoud-zino I'll have more time (and mental bandwidth) to review this in depth next week. We'll keep this pending until then. Thanks!

Mahmoud-zino:
## Summary 1
This is the initial refactor of the component, Please feel free to suggest any changes (including deleting everything and rebuilding it 😄).

I know you wanted to make the component "dumb" and extract the logic outside, but I think what is now implemented is solid and should satisfy our needs.

### Here are the changes I made:
#### Props
1. Introduced a `QueryIndent` type (better name suggestions would be appreciated) where the user can add a query (type, class, or id) and an indentClass representing the associated CSS class. This enables users to include items other than headings.
2. To make the TOC fully functional, we needed some mechanism to exclude items, so I added excludeQuery, which will be added to the `:not()` selector.
3. scrollParent and target are still the same as the last version.
4. Added `triggerUpdate` prop of type `unknown`, when it is changed it will trigger the TOC update. This helps for example to update the TOC depending on the route path.
5. For the styles I just used your stuff, which are like always, really good 👍.

#### Functionality
1. refreshList: very similar to what it was before, just adding the exclude queries. This also solves the bug where ignored elements will still be activated, so the table of content will have no active element visible.
2. changed the activated element logic, so it doesn't disappear when no elements are present on the screen (scroll to the end of the page to see it in the old version)
3. Since I wanted to be able to add multiple TOCs on the same page (especially nice for our TOC docs page), I ended up using `scrollBy` instead of `scrollIntoView`, but It will still work the same.
4. I didn't remove the permalink/header id generator away from the component, because the logic I am using is coupled to the TOC, so it didn't make sense to extract it.

#### HTML
1. Since the TOC is normally implemented on the side, I replaced the top div with aside, but I am not sure this is a good idea, because a lot of blogs place the TOC at the top of the page.
2. everything else is still the same.

### Suggestions (I need your confirmation on those, I am skeptical to add them)
1. we could add a flash animation, It will be activated when the user clicks on one of the links in the TOC.
This can be done using the dynamic-transitions feature and our future custom transition "Flash" 😄 (No, simpler is better)
2. we could add a different style to all links above the active link, (like gray color for example)
This will introduce a style prop that is empty by default and the user can use it to change the style. (No, simpler is better)
3. Add a new section in the docs to show how to add a hidden link (like our demo link in docs). 
4. I haven't included the permalink icon on headers because the new TOC won't only operate with headers, however, we could add a prop to show/hide them which gives users more control.

Mahmoud-zino:
As discussed in Discord, it was decided to:
1. take a simpler approach to only query links with class `.permalink` and display spacing depending on the wrapping element (h2-h6)
2. provide our styling for the `.permalink` class (users can always modify/override it)

Mahmoud-zino:
@endigo9740 
These are some concerns about the new implementation:
## TOC
1. we still have to pass `scrollParent` & `target` to scan and activate the elements in TOC (I wish there is a better way of handling this but I can't think of any).
2. when querying for the permalinks in the TOC to activate the top visible element, I am looking for a permalink that has an Id matching the href of the link passed to the TOC:
```ts
const element = parentElement?.querySelector(`${link.href}`) as HTMLAnchorElement | null;
```
this means the user MUST use id for the permalink and not href... is this something you would encourage? if so I will make sure to include it in the docs, but I wanted to ask you first.
## tocStore
We agreed to make the store hidden for users, however, to support both manual and auto mode, the user will have to use the store and pass it to the TOC. so the store is visible.
## TOCCrawler
I know, I know, the name is bad, but it is just a placeholder, and I haven't implemented it fully, I still have to implement a way for users to update/rescan elements.
1. When auto-generating the permalinks, I am using the id of the headings as the Id of the permalink, which is bad because IDs should be unique. the reason I have opted for this wrong method is the fact that we can't use the text of the heading as the permalink ID because of issues like this: #1741 , do you have any suggestions in this regard?

----

Currently, I have set up the docs to display these usages but didn't invest much time in formatting the text so users can understand it easily, I will get back to it once we solve those issues.

endigo9740:
Hey @Mahmoud-zino, so I've had a chance to review over your implementation here. I'm starting to worry this may be a situation where I have a particular visions of how I want this implemented, and I'm not adequately able to express it in such a way that you or anyone else could possibly hope to implement it in a reasonable fashion.

Rather than trying to adjust what you're implement or trying to spend another afternoon talking through all this, how about I go ahead and setup a second PR, or maybe even a standalone prototype, that implements the version of this I have in my head?

I'd still like your help with certain portions, such as implementing the dynamic transition features and some of the unique features you were aiming to implement in your drafts. Would you be ok with that?

If so I'll likely aim to begin this on Monday next week.

Mahmoud-zino:
@endigo9740 sure, I will be really happy to see how you would solve it and I will for sure help wherever I can 👍

endigo9740:
I've worked with @Mahmoud-zino to create another PR that approaches this from another direction. Please see this PR as the successor to this change:

- https://github.com/skeletonlabs/skeleton/pull/1832

PR #1730, opened 2023-07-05
breaking: Skeleton Version 2

AdrianGonz97:
Before merging this into `dev` we must:
- [x] Remove the `private` field from `packages/plugin/package.json`


endigo9740:
I'll create a running list of issues to resolve here. This may expand over time:

- [ ] The onboarding instructions for the Tailwind config to favor `.ts | .js | .cjs` in that tab order
- [ ] The onboarding instructions for the Tailwind config need recommend `.ts` for Typescript and `.js` for JS
- [ ] When using Typescript, users will need to install default node types via: `npm add --save-dev @types/node`

Issue #1708, opened 2023-06-30
Fix Svelte 4 a11y warnings


AdrianGonz97:
### Current Behavior

These are denoted with a `TODO: Remove for V2` tag so that they may be easily searchable.

Mahmoud-zino:
Since the AppRailTile has also an issue with a11y, It could be a good opportunity to refactor it and implement keyboard support.

ollema:
Just wanted to mention that I believe that one of the warnings was not silenced:
```
7:06:18 PM [vite-plugin-svelte] /node_modules/.pnpm/@skeletonlabs+skeleton@1.9.0_svelte@4.0.3/node_modules/@skeletonlabs/skeleton/dist/components/Ratings/Ratings.svelte:29:3 A11y: <svelte:element> with click handler must have an ARIA role
```
Maybe this was known?

Mahmoud-zino:
I would like to work on this issue 🙏.

endigo9740:
@Mahmoud-zino I might have you help me with one, but I really need to be hands on for anything a11y related. There's a lot of rules and we've gotta make sure changes are implemented exactly as need be. In this case I know we'll be refactoring a number of components, but I don't know the specifics needed. Each will be determined on a case-by-case basis - even though we might end up making similar changes in multiple places.

I'll try to ping you when I circle back to this though. Thanks!

endigo9740:
Making a quick note here regarding these changes in regards to the Tab component specifically:

- https://github.com/skeletonlabs/skeleton/issues/1752#issuecomment-1712090347

endigo9740:
I'm closing this as we will likely not have time to fix this in v2, and v3 will contain rewrites as we port to Svelte 5. Most, if not all, of these issues should be handled along the way. This means this issue is no longer relevant to active tasks.

PR #1756, opened 2023-07-17
Add passive event listener Modal

### LINKED ISSUE #1755: Passive event listener not added to 'touchstart' and 'touchend'
This error is showing when using [Modals](https://www.skeleton.dev/utilities/modals) :
`[Violation] Added non-passive event listener to a scroll-blocking 'touchstart' event. Consider marking event handler as 'passive' to make the page more responsive. See https://www.chromestatus.com/feature/5745543795965952`
I think it's related to this :
https://github.com/skeletonlabs/skeleton/blob/fa2af91a23e37a92ddf820c578ffaab135766a99/packages/skeleton/src/lib/utilities/Modal/Modal.svelte#L204-L205
It should be:
```js
on:touchstart|passive
on:touchend|passive
```

ibilux:
## Linked Issue

Closes #1755

## Description

This error is showing when using [Modals](https://www.skeleton.dev/utilities/modals) :
`[Violation] Added non-passive event listener to a scroll-blocking 'touchstart' event. Consider marking event handler as 'passive' to make the page more responsive. See https://www.chromestatus.com/feature/5745543795965952`

### DIFF
--- a/packages/skeleton/src/lib/utilities/Modal/Modal.svelte
+++ b/packages/skeleton/src/lib/utilities/Modal/Modal.svelte
@@ -201,8 +201,8 @@
 			data-testid="modal-backdrop"
 			on:mousedown={onBackdropInteractionBegin}
 			on:mouseup={onBackdropInteractionEnd}
-			on:touchstart
-			on:touchend
+			on:touchstart|passive
+			on:touchend|passive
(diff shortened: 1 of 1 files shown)


endigo9740:
@AdrianGonz97 looping you in on this as I figure it may overlap with #1708

PR #1768, opened 2023-07-18
bugfix/inputchips-updates-twice

### LINKED ISSUE #1759: InputChips subscribe twice to passed in store
### Current Behavior

If a store is used as a component value, input chips currently subscribes twice to it.
Once, to set the value to the internal hidden Select, and once for it's other logi.

If people use a fancy store that might do API requests or similar, this could lead to problems where the API is called twice or logic is fired twice that shouldn't be called twice.

This might also affect other components with a hidden select like the ListBoxes.

Relevant Discord link:
https://discord.com/channels/1003691521280856084/1130244031859925152

### Expected Behavior

Components should subscribe to passed in store values only once.

### Steps To Reproduce

1) Using input chips
```html
<InputChip bind:value={$inputChipWritable}/>
```

2) Pass in a store with this setup:
```ts
const inputChipWritable: Writable<string[]> = writable([]);
    inputChipWritable.subscribe((value) => {
        console.log("sub input");
    });
```
The console log will be triggered twice

_No response_

### More Information

_No response_

Mahmoud-zino:
## Linked Issue

Closes #1759

## Description

after deleting 
```ts
$: chipValues =
		value?.map((val, i) => {
			if (chipValues[i]?.val === val) return chipValues[i];
			return { id: Math.random(), val: val };
		}) || [];
```
every thing works as expected and the value is updated only once.

I have tested it with all use cases from the docs.

## Changsets

bugfix: InputChips updates bound value only once.

## Checklist

### DIFF
--- a/.changeset/twenty-spoons-hope.md
+++ b/.changeset/twenty-spoons-hope.md
@@ -1,0 +1,5 @@
+---
+"@skeletonlabs/skeleton": patch
+---
+
+bugfix: InputChips updates bound value only once.
--- a/packages/skeleton/src/lib/components/InputChip/InputChip.svelte
+++ b/packages/skeleton/src/lib/components/InputChip/InputChip.svelte
@@ -158,11 +158,6 @@
 	$: classesInterface = `${cInterface}`;
 	$: classesChipList = `${cChipList}`;
 	$: classesInputField = `${cInputField}`;
-	$: chipValues =
-		value?.map((val, i) => {
-			if (chipValues[i]?.val === val) return chipValues[i];
-			return { id: Math.random(), val: val };
-		}) || [];
 </script>
 
 <div class="input-chip {classesBase}" class:opacity-50={$$restProps.disabled}>


endigo9740:
@Mahmoud-zino checking the Git Blame I see @Sarenor introduced this portion of the code per:

- https://github.com/skeletonlabs/skeleton/commit/cbef3c3414d2cee2a20f9352f0812ef9bdc3c6b2
- https://github.com/skeletonlabs/skeleton/pull/1678

It looks like this was added to allow for null values to be handled by the autocomplete. Can the two of you sync on this and determine the best course of action here? My gut is telling me that just flat out removing this would introduce a regression.

Sarenor:
I just added the null-check, the initial code was added to handle duplicate values:
https://github.com/skeletonlabs/skeleton/commit/2ba426b712fff76a47e7c949fcbbfadfcea92142
(by Chris, at that ;) ) which seems to work though, so the ID-Mapping is no longer necessary? Seems kind of strange but I'll take it!

Passing in an undefined list breaks the component though, and I'm not quite sure if we want that as intentional behaviour?

endigo9740:
Hmm, I wonder why the blame showed you. That's odd.

Sarenor:
Cause it was me that changed those lines the last time when I added the ``|| [];`` and prettier reformatted the lines :)

Mahmoud-zino:
I will get back to this issue once I am done with the tree-view, I want to handle undefined without a reactive assignment, once this is done I think we can safely delete the reactive assignment.

Mahmoud-zino:
@Sarenor 
> Passing in an undefined list breaks the component though, and I'm not quite sure if we want that as intentional behaviour?

can you please provide your setup, I tried passing undefined and got no errors, It is working as expected, maybe I am missing something here...

endigo9740:
@Sarenor just pinging you to remind you about @Mahmoud-zino request above.

Sarenor:
@Mahmoud-zino Sorry, totally missed this while I was camping.
Got the reproduction here:
https://github.com/skeletonlabs/skeleton/compare/dev...Sarenor:skeleton:undefined-input-chip?expand=1
If I set the anything-list in the docs to undefined the docs page for InputChip doesn't load anymore

Mahmoud-zino:
@Sarenor this happens because we are accessing the lists to display what is selected/allowed/not-allowed etc... on the docs page. but the error is not coming from the component.

so commenting this:
```html
<code class="code">{anythingList.length ? anythingList.join(', ') : 'No chips available.'}</code>
```
the component will work as expected.
If you can confirm, I will mark the PR as ready for review

Mahmoud-zino:
@Sarenor no no, it is just easy to miss 👍

endigo9740:
Thanks for your cooperation @Mahmoud-zino @Sarenor 

I think with that we're good to move forward with this. I'll merge and we'll monitor the reaction from end users going forward.

PR #1793, opened 2023-07-25
Check for slot content in `AppRail`

ibilux:
## Description

Check for `lead` and `trail` slots content in `AppRail` before inserting html code.

### DIFF
--- a/.changeset/young-bags-add.md
+++ b/.changeset/young-bags-add.md
@@ -1,0 +1,5 @@
+---
+"@skeletonlabs/skeleton": patch
+---
+
+chore: set conditional check for `lead` and `trail` slot content in AppRail
--- a/packages/skeleton/src/lib/components/AppRail/AppRail.svelte
+++ b/packages/skeleton/src/lib/components/AppRail/AppRail.svelte
@@ -63,9 +63,13 @@
 
 <div class="app-rail {classesBase}" data-testid="app-rail">
 	<!-- Slot: lead -->
-	<div class="app-bar-lead {classesRegionLead}"><slot name="lead" /></div>
+	{#if $$slots.lead}
+		<div class="app-bar-lead {classesRegionLead}"><slot name="lead" /></div>
+	{/if}
 	<!-- Slot: Default -->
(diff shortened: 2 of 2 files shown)


Sarenor:
This doesn't look bad, but it should definitely include a change set.

Mahmoud-zino:
The PR is targeting the master branch and the branch name doesn't follow Skeleton's style guide.
please reffer to https://www.skeleton.dev/docs/contributing

endigo9740:
@ibilux if you need us to change the target branch we can, you should always target `dev` please. But a changeset will be required here. Our changelog is programmatic, generated from that. Make sure to read the PR message rather than replace it, it covers this and other requirements. See another PR if you need instruction for this.

ibilux:
> The PR is targeting the master branch and the branch name doesn't follow Skeleton's style guide. please refer to https://www.skeleton.dev/docs/contributing

I didn't notice i was on the `master` branch, should I close it and make a new one in the `dev` branch.

ibilux:
> @ibilux if you need us to change the target branch we can, you should always target `dev` please. But a changeset will be required here. Our changelog is programmatic, generated from that. Make sure to read the PR message rather than replace it, it covers this and other requirements. See another PR if you need instruction for this.

I think the `cBase` class should be updated too ?
```ts
const cBase = `grid grid-rows-[${[$$slots.lead?'auto':undefined,'1fr',$$slots.trail?'auto':undefined].filter(Boolean).join('_')}] overflow-y-auto`;
```

endigo9740:
@ibilux I'd recommend reading through out style guide in the contributor section here:
https://www.skeleton.dev/docs/contributing/style-guide

We don't set dynamic values in the `cBase`. Create a reactive property to handle this.

endigo9740:
@ibilux @Sarenor @Mahmoud-zino I was in process on switching the target branch and appending a changeset. However, after reviewing the change to the grid styling, I'm thinking this may not be a useful change. Having empty slots causes no harm, and would make adjusting the `grid-rows-x` styling VERY difficult to adjust per all possible configurations. The simplest solution is sometimes the best, so I think we're close this out and keep things as is.

PR #1831, opened 2023-08-06
breaking: Refactor global stores for `Modal`, `Toast`, and `Drawer`

### LINKED ISSUE #522: Unsafe store behaviour of utility components in SvelteKit

AdrianGonz97:
## Linked Issue

Closes #522 

## Description

This PR aims to fix the SSR related issues that comes with stores that are defined in the global scope. To fix this, we'll be using Svelte's context API as a means of attaching these stores to the component tree for safe keeping.

### Store Initialization
The first change is how these utility stores are initialized. As we are using contexts now, we need to initialize and set the contexts for these stores in the root layout file. 

We provide a utility function, `initializeStores()`, that will handle the initialization for the users:
```svelte
<!-- src/routes/+layout.svelte -->
<script lang="ts">
	import { Toast, Modal, Drawer, initializeStores } from "@skeletonlabs/skeleton";
	
	initializeStores();
</script>

<Toast />
<Modal />
<Drawer />
```

### Retrieving Stores
Stores for `Modal`, `Toast`, and `Drawer` are now accessed through a function that will retrieve their respective stores.

```ts
import { getToastStore, getDrawerStore, getModalStore } from "@skeletonlabs/skeleton";

const toastStore = getToastStore();
const drawerStore = getDrawerStore();
const modalStore = getModalStore();
```

...and that's it! It's a pretty minor, yet effective change that clears up all of the nasty SSR issues that these particular stores have been troubling us with.

As a bonus, these stores can now be safely used at the top level of components without any fear of the server's store persisting its state:
```svelte
<script lang="ts">
	import { toastStore } from '@skeletonlabs/skeleton';

	toastStore.trigger({
		message: 'Howdy!',
		autohide: false
	});
</script>

<!-- ... -->
```

## Changsets

## Checklist

### DIFF
--- a/packages/skeleton/src/lib/utilities/index.ts
+++ b/packages/skeleton/src/lib/utilities/index.ts
@@ -2,6 +2,23 @@ import { initializeModalStore } from './Modal/stores.js';
 import { initializeToastStore } from './Toast/stores.js';
 import { initializeDrawerStore } from './Drawer/stores.js';
 
+/**
+ * Used to initialize the stores for the `Modal`, `Toast`, and `Drawer` utilities.
+ *
+ * @example
+ * ```svelte
+ * <!-- App's root +layout.svelte -->
+ * <script>
+ * 	import { initializeStores, Toast, Modal, Drawer } from "@skeletonlabs/skeleton";
+ *
+ * 	initializeStores();
+ * </script>
+ *
+ * <Toast />
+ * <Modal />
+ * <Drawer />
+ * ```
+ */
 export function initializeStores() {
 	const modalStore = initializeModalStore();
 	const toastStore = initializeToastStore();
--- a/packages/skeleton/src/lib/utilities/index.ts
+++ b/packages/skeleton/src/lib/utilities/index.ts
@@ -2,14 +2,7 @@ import { initializeModalStore } from './Modal/stores.js';
 import { initializeToastStore } from './Toast/stores.js';
 import { initializeDrawerStore } from './Drawer/stores.js';
 
-type Opts = {
-	modal: boolean;
-	toast: boolean;
-	drawer: boolean;
-	popup: boolean;
-	lightswitch: boolean;
-};
-export function initializeStores(opts: Opts = { modal: true, toast: true, drawer: true, popup: true, lightswitch: true }) {
+export function initializeStores() {
 	const modalStore = initializeModalStore();
 	const toastStore = initializeToastStore();
 	const drawerStore = initializeDrawerStore();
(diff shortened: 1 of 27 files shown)


endigo9740:
Seems good at at glance @AdrianGonz97. I'll test it out this week and we'll aim to get this in for the v2 release. Thank you for updating the docs already as well!

packages/skeleton/src/lib/utilities/index.ts
```diff
@@ -0,0 +1,28 @@
...
+ * <Modal />
+ * <Drawer />
+ * ```
+ */
+export function initializeStores() {
+	const modalStore = initializeModalStore();
+	const toastStore = initializeToastStore();
+	const drawerStore = initializeDrawerStore();
+
+	return { drawerStore, modalStore, toastStore };
```
Sarenor, line 27: Should we return these here or is that just going to lead to people using it for anti patterns like prop drill down?

AdrianGonz97: @Sarenor Fair point! I figured that there was some use case where someone would want to immediately use a store after initialization, but I'd be down to just making this return void instead.

Sarenor: I'm open to discussion, there might be a use for it as you said. Just want to avoid people thinking they have to pass it down

AdrianGonz97: Nah, I'm convinced. I was debating this internally anyway when I initially wrote it. Having a single interface for retrieving the stores is better and more clear to the users!

endigo9740:
@AdrianGonz97 I always like to add context for WHY you're implementing something in the instruction, so I've modified this bit in the featured example for drawers, modals, and toasts:

Screenshot 2023-08-09 at 12 52 02 PM

I think with this I'm good with moving forward and merging this.

endigo9740 (approved):

PR #1836, opened 2023-08-08
Feat/radio group region label prop

### LINKED ISSUE #1786: "Incorrect" suggestion in Radio Groups
### Link to the Page

https://www.skeleton.dev/components/radio-groups#full-width

Documentation recommends using

`<RadioGroup display="flex">...</RadioGroup>`

"to stretch and fill the full width.", which is incorrect as it results in a not-streched group:

vivaldi_EzNkqnD7mZ

To achieve the desired result (see next pic) we need to use an arbitrary variant as there is no prop to add classes to the child labels. `class="flex [&>label]:flex-auto"` looks to work.

vivaldi_PXcIeTq2O1

I guess easiest fix is to add that class to the doc example, although maybe it can be useful to add a new prop.

Let me know which solution do you think it's best and I can make a pr.

SebasF1349:
## Linked Issue

Closes #1786 

## Description

Added `regionLabel` prop to the Radio Group

## Changsets

## Checklist

### DIFF
--- a/.changeset/shy-suits-provide.md
+++ b/.changeset/shy-suits-provide.md
@@ -1,0 +1,6 @@
+---
+"@skeletonlabs/skeleton": minor
+"skeleton.dev": patch
+---
+
+Added `regionLabel` prop to Radio Groups
--- a/packages/skeleton/src/lib/components/Radio/RadioGroup.svelte
+++ b/packages/skeleton/src/lib/components/Radio/RadioGroup.svelte
@@ -27,6 +27,8 @@
 	export let color: CssClasses = '';
 	/** Provide classes to set the highlighted SVG fill color. */
 	export let fill: CssClasses = '';
+	/** Provide classes for the label region. */
+	export let regionLabel: CssClasses = '';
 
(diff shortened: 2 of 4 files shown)


Mahmoud-zino:
suggestion: we might want to add `regionItem` and `regionInput` in the same PR to cover all possible regions.

endigo9740:
@SebasF1349 FYI you had the right idea, but the fix isn't quite right.

This was actually a regression introduced when we added the `label > div` structure. All the "base" styles were being applied to the `div`, even though base styles should ALWAYS be applied to the outter-most parent element. The `label` in this case.

You could see `cBase` set already setting `flex-auto` previously. But once that was removed from `label`, this brought about the regression.

I've committed a small change that resolves all this, and kept your `regionLabel` prop, as I'm sure it'll be useful! Let me know if you approve and I'll aim to merge this right away.

SebasF1349:
Now the docs comment makes sense, it was correct before. Thanks for the explanation. Didn't notice the `flex-auto`, should have checked the other classes. Let's merge it!

PR #1910, opened 2023-08-20
bugfix/tree-view-relational-updates

### LINKED ISSUE #1885: TreeView doesn't uncheck to root node only parent.
### Current Behavior

image

### Expected Behavior

_No response_

### Steps To Reproduce

If i check "Auto & Tires" and then "Baby Oils & Lotions", "Auto & Tires" stay chcked.

_No response_

### More Information

_No response_

Mahmoud-zino:
## Linked Issue

Closes #1885

## Needs to be done
we are now using export functions to handle part of the relations, but it is displayed in the props table. we need a way to hide it.
image

## Description

fixed relational updates for both recursive and non-recursive modes.

This is a basic diagram of how the tree-view relations behave:
Untitled Diagram drawio

### Extra changes
1. Forward all events triggered in recursive mode to tree-view.
2. deleted the toggleAll functions (to handle all possible setups `selectAll`, `deselectAll` will need extreme effort for such a tiny feature -_-)

## Changsets

bugfix: Tree-view relational updates are now working as expected.

## Checklist

### DIFF
--- a/.changeset/cool-glasses-grow.md
+++ b/.changeset/cool-glasses-grow.md
@@ -1,0 +1,5 @@
+---
+"@skeletonlabs/skeleton": patch
+---
+
+bugfix: Tree-view relational updates are now working as expected.
--- a/packages/create-skeleton-app/CHANGELOG.md
+++ b/packages/create-skeleton-app/CHANGELOG.md
@@ -4,15 +4,15 @@
 
 ### Patch Changes
 
-- Fixed @rc pinning ([#1882](https://github.com/skeletonlabs/skeleton/pull/1882))
+-   Fixed @rc pinning ([#1882](https://github.com/skeletonlabs/skeleton/pull/1882))
 
 ## 0.0.46-rc.2
 
 ### Patch Changes
 
-- CSA updates for V2 ([#1859](https://github.com/skeletonlabs/skeleton/pull/1859))
-  - Install vite-plugin-tailwind-purgecss to enable purging of unused component CSS
-  - mdsvex is an optional install
-  - No longer installs packages by default
-  - Templates can now specify additional packages to be installed
-  - Bug fixes and improvements all over
+-   CSA updates for V2 ([#1859](https://github.com/skeletonlabs/skeleton/pull/1859))
+    -   Install vite-plugin-tailwind-purgecss to enable purging of unused component CSS
+    -   mdsvex is an optional install
+    -   No longer installs packages by default
+    -   Templates can now specify additional packages to be installed
+    -   Bug fixes and improvements all over
--- a/packages/skeleton/src/lib/components/TreeView/TreeView.svelte
+++ b/packages/skeleton/src/lib/components/TreeView/TreeView.svelte
@@ -81,40 +81,6 @@
 			}
 		});
 	}
-	/**
-	 * select all tree view items. Only available in Multiple selection mode.
-	 * @type {() => void}
-	 */
-	export function selectAll(): void {
-		const detailsElements = [...tree.querySelectorAll('details.tree-item')] as HTMLDetailsElement[];
-		detailsElements.forEach((details) => {
-			const input: HTMLInputElement | null = details.querySelector('input[type="checkbox"].tree-item-checkbox');
-			if (!input) return;
-			if (!input.checked) {
-				// needs delay
-				setTimeout(() => {
-					input.click();
-				}, 5);
-			}
-		});
-	}
-	/**
-	 * deselect all tree view items. Only available in Multiple selection mode.
-	 * @type {() => void}
-	 */
-	export function deselectAll(): void {
-		const detailsElements = [...tree.querySelectorAll('details.tree-item')] as HTMLDetailsElement[];
-		detailsElements.forEach((details) => {
-			const input: HTMLInputElement | null = details.querySelector('input[type="checkbox"].tree-item-checkbox');
-			if (!input) return;
-			if (input.checked) {
-				// needs delay
-				setTimeout(() => {
-					input.click();
-				}, 5);
-			}
-		});
-	}
 
 	// Context API
 	setContext('open', open);
@@ -149,7 +115,7 @@
 	aria-disabled={disabled}
 >
 	{#if nodes && nodes.length > 0}
-		<TreeViewDataDrivenItem bind:nodes />
+		<TreeViewDataDrivenItem bind:nodes on:change on:click on:toggle on:keydown on:keyup />
 	{:else}
 		<slot />
 	{/if}
--- a/packages/skeleton/src/lib/components/TreeView/TreeViewDataDrivenItem.svelte
+++ b/packages/skeleton/src/lib/components/TreeView/TreeViewDataDrivenItem.svelte
@@ -2,13 +2,10 @@
 	/**
 	 * This component is only in Data-driven tree-view to add children recursively.
 	 */
-	import { createEventDispatcher, getContext, onMount } from 'svelte';
+	import { getContext, onMount } from 'svelte';
 	import TreeViewDataDrivenItem from './TreeViewDataDrivenItem.svelte';
 	import TreeViewItem from './TreeViewItem.svelte';
 	import type { TreeViewNode } from './types.js';
-
-	// events
-	const dispatch = createEventDispatcher();
 
 	// this can't be passed using context, since we have to pass it to recursive children.
 	/** Provide data-driven nodes. */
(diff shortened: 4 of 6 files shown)


Sarenor:
Didn't break while I was testing it :)
Have you tried providing  an ``@type`` comment to the exported functions?
```js
/**
 * called when X happens
 * @type {() => A|B|C}
 */
 ```

**Comment by Mahmoud-zino** [comment]
> Didn't break while I was testing it :) Have you tried providing an `@type` comment to the exported functions?
> 
> ```js
> /**
>  * called when X happens
>  * @type {() => A|B|C}
>  */
> ```

@type will only change the type, but I wanted to hide it entirely because users will not need this information. and it is displaying the function body in the value which is hella ugly 😄

**Comment by Sarenor** [comment]
Damn. I hoped that'd overwrite a bit more. Ah well... :|

**Comment by ak4zh** [comment]
I am trying to render a pre-checked parent but the parent remains unchecked.

```js
let multipleDD: TreeViewNode[] = [
	{
		content: 'Books',
		lead: '<i class="fa-solid fa-book-skull"></i>',
		open: true,
		indeterminate: false,
		checked: true,
		children: [
			{ content: 'Clean Code', value: 'Clean Code', checked: true },
			{ content: 'The Clean Coder', value: 'The Clean Coder', checked: true },
			{ content: 'The Art of Unix Programming', value: 'The Art of Unix Programming', checked: true }
		],
		value: 'books'
	},
	{
		content: 'Movies',
		lead: '<i class="fa-solid fa-film"></i>',
		children: [
			{ content: 'The Flash', value: 'The Flash' },
			{ content: 'Guardians of the Galaxy', value: 'Guardians of the Galaxy' },
			{ content: 'Black Panther', value: 'Black Panther' }
		],
		value: 'movies'
	},
	{
		content: 'TV',
		lead: '<i class="fa-solid fa-tv"></i>',
		value: 'tv'
	}
];
```
Screenshot 2023-08-21 at 7 50 15 PM

ak4zh:
I've encountered another issue.
I added a log at https://github.com/Mahmoud-zino/skeleton/blob/411b3b660a7ae159d03119abcecf1a350a01b3f9/packages/skeleton/src/lib/components/TreeView/TreeViewDataDrivenItem.svelte#L72, and I can see an infinite log loop. It appears that the code is stuck in some sort of recursive loop, leading to continuous execution of reactivity.

Mahmoud-zino:
@ak4zh can you please provide your setup, the tree-view is a very tricky component so I really appreciate you digging in it and finding these juicy bugs 😉

ak4zh:
@Mahmoud-zino 
1) I just checked, and the parent item check is now working correctly.
2) It turns out the second issue was a mistake. I put in a log, and I saw more than 10000 logs in the console. But this seems to only happen when I add the log to the skeleton code while my app is already using the skeleton setup on my computer. 

I fixed it by putting in the log, closing my app, and then starting the server again. Now everything seems fine – the count only goes up as I add or remove items in the tree. So, when I test checking and unchecking things, it's all good.

Screenshot 2023-08-21 at 10 33 32 PM

ak4zh:
Additionally I notice you are using the spread operator to add the `node.value`. Can we use push instead?
I am dealing with a huge list of items in the `TreeView`, so I assume push would be better.

ak4zh:
Verified the functionality works fine after [012464c](https://github.com/skeletonlabs/skeleton/pull/1910/commits/012464ca50a2456bfaed42cdb1b0e41ed9bc5180).
Triggers only when something changes.

Mahmoud-zino:
> Additionally I notice you are using the spread operator to add the `node.value`. Can we use push instead? I am dealing with a huge list of items in the `TreeView`, so I assume push would be better.

This is used to trigger reactivity in svelte, so if we want to use .push instead we will still have to do
```ts
array.push(value);
array = array;
```
which I think will have the same effect on speed + the spread method is the recommended way of doing this from the svelte docs https://svelte.dev/tutorial/updating-arrays-and-objects

ak4zh:
Got it. Thanks for the explanation.
Waiting for this to be merged so I can make use of it.
The new feature of my app heavily depends on the TreeView, with the new changes that dropped in v2 and the above bug fixes it is now super easy to implement really powerful TreeViews.

Before this I was generating the tree view myself and it was messy code.
Excited to push it to production as soon as this is merged.

ak4zh:
@Mahmoud-zino On another thought I realised since the recent changes, we've moved away from using reactivity. As a result, I proceeded to test with `array.push`, and everything appears to be functioning accurately.

Just adding a note.

endigo9740:
FYI my goal will be to review this again tomorrow. If we're confident all is well I'll aim to merge then. This will make it part of the latest v2 RC release.

ak4zh:
I recently noticed that the selectAll and deselectAll functions are being removed in this pull request. I've been brainstorming ways to streamline this, and I came up with a somewhat unconventional solution. What if we try this approach?

```js
let refreshKey = String(Math.random());

export const actions = {
	updateAll(subNodes: TreeViewNode[], updateField: 'checked'|'open', status: boolean) {
		subNodes.forEach(n => {
			n[updateField] = status;
			if (updateField === 'checked') n.indeterminate = false;
			if (n.children && n.children.length) actions.updateAll(n.children, updateField, status)
		})
		refreshKey = String(Math.random());
	},

	selectAll() {
		actions.updateAll(nodes, 'checked', true)
	},

	deselectAll() {
		actions.updateAll(nodes, 'checked', false)
	},

	expandAll() {
		actions.updateAll(nodes, 'open', true)
	},

	collapseAll() {
		actions.updateAll(nodes, 'open', false)
	}
}
```

```html
	{#key refreshKey}
		{#if nodes && nodes.length > 0}
			<TreeViewDataDrivenItem bind:nodes on:change on:click on:toggle on:keydown on:keyup />
		{:else}
			<slot />
		{/if}
	{/key}
```


```html
// +page.svelte
<svelte:fragment slot="preview">
	<TreeView bind:actions={multiDDAction} bind:nodes={multipleDD} selection multiple />
</svelte:fragment>
<svelte:fragment slot="footer">
	<button class="btn variant-filled" on:click={() => multiDDAction.selectAll()}>Select All</button>
	<button class="btn variant-filled" on:click={() => multiDDAction.deselectAll()}>Unselect All</button>
	<button class="btn variant-filled" on:click={() => multiDDAction.expandAll()}>Expand All</button>
	<button class="btn variant-filled" on:click={() => multiDDAction.collapseAll()}>Collapse All</button>
</svelte:fragment>
```

ak4zh:
Here's my repo with above suggested changes for quick test, everything seems to work correctly, not sure if this hacky approach is good enough for chris:
https://github.com/ak4zh/skeleton/tree/bugfix/tree-view-relational-updates

Mahmoud-zino:
> @Mahmoud-zino On another thought I realised since the recent changes, we've moved away from using reactivity. As a result, I proceeded to test with `array.push`, and everything appears to be functioning accurately.
> 
> Just adding a note.

just tested both variants there is no difference whatsoever (also on performance).
but please note this is not because we are not using reactivity, it is because the code is executed in `onMount`, and reactivity is still triggered in `TreeViewItem`.
However, I would like to keep it as is just to keep the code clear and the style constant, if we notice any issue in performance we can change it any time.

Mahmoud-zino:
> I recently noticed that the selectAll and deselectAll functions are being removed in this pull request. I've been brainstorming ways to streamline this, and I came up with a somewhat unconventional solution. What if we try this approach?
> 
> ```js
> let refreshKey = String(Math.random());
> 
> export const actions = {
> 	updateAll(subNodes: TreeViewNode[], updateField: 'checked'|'open', status: boolean) {
> 		subNodes.forEach(n => {
> 			n[updateField] = status;
> 			if (updateField === 'checked') n.indeterminate = false;
> 			if (n.children && n.children.length) actions.updateAll(n.children, updateField, status)
> 		})
> 		refreshKey = String(Math.random());
> 	},
> 
> 	selectAll() {
> 		actions.updateAll(nodes, 'checked', true)
> 	},
> 
> 	deselectAll() {
> 		actions.updateAll(nodes, 'checked', false)
> 	},
> 
> 	expandAll() {
> 		actions.updateAll(nodes, 'open', true)
> 	},
> 
> 	collapseAll() {
> 		actions.updateAll(nodes, 'open', false)
> 	}
> }
> ```
> 
> ```
> 	{#key refreshKey}
> 		{#if nodes && nodes.length > 0}
> 			<TreeViewDataDrivenItem bind:nodes on:change on:click on:toggle on:keydown on:keyup />
> 		{:else}
> 			<slot />
> 		{/if}
> 	{/key}
> ```
> 
> ```
> // +page.svelte
> <svelte:fragment slot="preview">
> 	<TreeView bind:actions={multiDDAction} bind:nodes={multipleDD} selection multiple />
> </svelte:fragment>
> <svelte:fragment slot="footer">
> 	<button class="btn variant-filled" on:click={() => multiDDAction.selectAll()}>Select All</button>
> 	<button class="btn variant-filled" on:click={() => multiDDAction.deselectAll()}>Unselect All</button>
> 	<button class="btn variant-filled" on:click={() => multiDDAction.expandAll()}>Expand All</button>
> 	<button class="btn variant-filled" on:click={() => multiDDAction.collapseAll()}>Collapse All</button>
> </svelte:fragment>
> ```

`key` will destroy the component and recreate it, I would avoid it when possible.
also and this is why I deleted this feature, we have to provide a solution for non-recursive mode also, which is not a trivial task since we don't have direct reference to children. so for the time being I would say we let the user handle this and maybe in the future once we have a better solution we can add it to the tree view.

AdrianGonz97 (approved):
I've made a couple of changes that get rid of some of the type assertions that weren't really necessary, as well as replacing the `group = [...group, node.value]` in favor of mutating the array. Svelte's tutorial may mention it as a convenient way to trigger reactivity, but I have strong opinions against it 😄.

Besides that, everything seems to be working perfectly.  Awesome work!

Mahmoud-zino:
>  but I have strong opinions against it 😄.

I would be interested in hearing your opinions 😄, we have a lot of places where we are using the spread operator to trigger interactivity, so if there is a problem other than the style it might be worth it to change them.

AdrianGonz97:
@Mahmoud-zino 
The spread operator is _horrendously_ slow. Like, several orders of magnitude slower than it's mutating counterpart. Mainly because it's allocating new memory **every single time** we use it. It certainly has its place for when you want to clone the values of an array into another, but when it comes to using it as a means to trigger reactivity, such as appending a new value to the array, it's horrific!

I like to reference this example when talking about the consequences of this approach:
slow as hell
And it's not like this is a large dataset either - it's only 1K elements!

As library authors, we should be, generally, cautious when it comes to performance, striving to be as performant as _reasonably_ possible. Using `arr.push(value)` instead of `arr = [...arr, value]` would be one of the small steps towards that goal.

This method of reassignment is one of those React idioms that has weaseled it's way into Svelte. It's popular among the React folks because of how their reactivity system works. Over there, state is based on _referential_ equality. Meaning, they _have_ create a whole new array in order to update their state. Over here in Svelte land, we have the luxury of simply mutating the state and being done with it. No excess memory allocation required. We might as well take advantage of it!

*/end-of-rant*

Mahmoud-zino:
oof, I though the different was trivial...
@AdrianGonz97 Thank you very much for this valuable infos, I will keep that in mind for the future 👍

PR #1918, opened 2023-08-22
bugfix: make inputchip values reactive

ak4zh:
## Description

InputChips was not updating when the selection changes.

## Checklist

### DIFF
--- a/.changeset/big-horses-wink.md
+++ b/.changeset/big-horses-wink.md
@@ -1,0 +1,5 @@
+---
+"@skeletonlabs/skeleton": patch
+---
+
+bugfix: make inputchip values reactive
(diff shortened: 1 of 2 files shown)


Sarenor:
Fixed #1922 now that #1922 exists

endigo9740:
@ak4zh two things here:

1. Make sure not to modify the PR template. We've implemented all fields that we wish to be populated. This is why @Sarenor had to post the linked issues separately.
2. I'm not sure I agree with the change proposed here. While it may solve the issue at hand, changing a huge swath of code to be reactive is like using a jackhammer to hammer in a nail - it's overkill. The more code the you make reactive, the more performance is impacted. I don't think we can move forward with this approach.

I'll try to sit down and review this issue in depth this week, as we've had a lot of comments regarding the Autocomplete not working in certain scenarios. I'd like to try and find the root of the issue and review more optimal ways to solve for this. So I appreciate the PR, but for now this will be on pause until I can revisit.

LarsHadidi:
Thank you 😃

PR #1919, opened 2023-08-22
bugfix: allow unstyled file buttons

### LINKED ISSUE #1876: Button Groups do not support File Buttons
### Current Behavior

Users on Discord have pointed out that Button Groups do not currently support the use of a embedded File Button component.

image

### Steps To Reproduce

The reason for this is Button Groups expect the child button elements to be native `<button>` elements without styles applied:
https://www.skeleton.dev/elements/buttons#button-groups

```html
<div class="btn-group-vertical variant-filled">
	<button>Months</button>
	<button>Days</button>
	<button>Years</button>
</div>
```

However, the File Button has a hardcoded `.btn` class implemented, meaning it has default styles which breaks this use case:
https://github.com/skeletonlabs/skeleton/blob/master/packages/skeleton/src/lib/components/FileButton/FileButton.svelte#L22

### More Information

We should move to make the `.btn` style a customizable prop. Perhaps included in the existing `button` prop styles along with the variant definition. As there may be other scenarios where an unstyled button is preferred.

ak4zh:
## Linked Issue

Closes #1876 

## Description
 
- Moved `.btn` class to `button` prop so it can be customised to render a buton without hardcoded styles.

## Changsets

## Checklist

### DIFF
--- a/.changeset/soft-tables-wave.md
+++ b/.changeset/soft-tables-wave.md
@@ -1,6 +1,5 @@
 ---
-"skeleton.dev": patch
 "@skeletonlabs/skeleton": patch
 ---
 
-Allow unstyled File Buttons
+Chore: Update file button `button` prop to allow for unstyled buttons
--- a/.changeset/soft-tables-wave.md
+++ b/.changeset/soft-tables-wave.md
@@ -1,0 +1,6 @@
+---
+"skeleton.dev": patch
+"@skeletonlabs/skeleton": patch
+---
+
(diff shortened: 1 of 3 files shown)


ak4zh:
I adjusted the default value of `width` to `w-full` so that the buttons within button groups would display properly. However, it seems appropriate to restore the original setting and advise that when utilized with button groups, the width should be explicitly set to `w-full`.

Issue #1142, opened 2023-03-12
NEXT: New Theme Generator


endigo9740:
> NOTE: This issue will need to be updated or regenerated to specify the requirements for the v4 release. We will not be implementing many of the ideas presented below.

This will act as a hub to centralize this information.

## UX Flow

![Screenshot 2024-09-16 at 1 37 37 PM](https://github.com/user-attachments/assets/49a1ff57-d797-436c-b5e7-82e629ec811c)

## Maintainer Requests

The following requests are coming straight from the Skeleton team. These are highly likely be implemented:

Generator:

- [x] Migrate the theme generator new a new standalone full-stack SvelteKit application (including database)
- [x] Consider making use of [chroma.js](https://gka.github.io/chroma.js/) for color management ([reference](https://github.com/skeletonlabs/skeleton/pull/1177))
- [x] Implement a wizard-like UX flow
- [x] Enable the ability to edit each individual color swatch on demand
- [x] Provide a feature to pin/lock certain colors on randomize ([reference](https://github.com/skeletonlabs/skeleton/issues/1153))
- [x] Make considerations for accessing theme color values via Typescript/Javascript
- [x] Allow users to import existing themes for additional edits
- [x] Unify theme import syntax and allow for direct import to support HTML Canvas usage ([reference](https://github.com/skeletonlabs/skeleton/issues/2360#issuecomment-1894410516))
- [x] A11y: provide better support for high/low contrast for light sensitive users  ([reference](https://github.com/skeletonlabs/skeleton/pull/1178))
- [x] https://github.com/skeletonlabs/skeleton/issues/2559
- [x] A11y: explicitly state support for [WCAG/APCA compatibility](https://typefully.com/u/DanHollick/t/sle13GMW2Brp)
- [ ] Post-Launch: Make considerations for social features: share, upvote, highest rated
- [ ] Post-Launch:  Consider the use of [Superforms](https://superforms.rocks/) and [Zod](https://zod.dev/) for form handling.
- [ ] Post-LauncH: Review integration with font service (ex: [Google Fonts](https://fonts.google.com/)) to preview and generate font imports
- [ ] Post-Launch: Background image + mesh gradient previews
- [ ] Proposal: [generate custom presets again user-defined colors](https://github.com/skeletonlabs/skeleton/issues/2964#issuecomment-2707688441)

Theme:

[...]

PR #1925, opened 2023-08-23
feat: color theory based theme generator

ak4zh:
## Linked Issue

Closes: None currently

## Description

This pull request introduces a new feature that allows users to generate theme colors based on various color theories. The feature adds a dropdown selector that lets users choose a color theory, and then generates a color palette following that theory.

## Changes

- Added a dropdown selector to the color generation section.
- Integrated color theory functions to generate theme colors based on the selected theory.
- Created functions for generating colors using various color theories, including Random, Complementary, Analogous, Triadic, Split-Complementary, Tetradic, Monochromatic, Neutral with Accents, Pastel, and High Contrast.

## Purpose

The purpose of this pull request is to discuss the proposed feature and gather feedback from the community. Users will be able to experiment with different color theories and generate theme colors that match their preferences and needs.

## Next Steps

This is a draft pull request, and I'm looking forward to receiving feedback on the feature implementation and user experience. Once the feedback is collected and addressed, we can move forward with refining the feature and ensuring it aligns well with the goals of the project.

Please feel free to share your thoughts and suggestions on this proposed feature. Your feedback is highly appreciated!

This PR is a work-in-progress and is being submitted as a draft for discussion and feedback.

## Changsets

## Checklist

- [ ] This PR targets the `dev` branch (NEVER `master`)

### DIFF
--- a/sites/skeleton.dev/src/lib/layouts/DocsThemer/DocsThemer.svelte
+++ b/sites/skeleton.dev/src/lib/layouts/DocsThemer/DocsThemer.svelte
@@ -12,11 +12,11 @@
 
 	// Local Utils
 	import { storePreview } from './stores';
-	import { storeTheme } from '$lib/stores/stores';
 	import type { ColorSettings, FormTheme, ContrastReport } from './types';
 	import { inputSettings, fontSettings } from './settings';
 	import { type Palette, generatePalette, generateA11yOnColor, hexValueIsValid, getPassReport } from './colors';
 	import type { PopupSettings } from '@skeletonlabs/skeleton';
+	import { generateAnalogousColors, generateComplementaryColors, generateHighContrastColors, generateMonochromaticColors, generateNeutralWithAccentColors, generatePastelColors, generateRandomHexColors, generateSplitComplementaryColors, generateTetradicColors, generateTriadicColors, randomHexCode } from './colorTheories';
 
 	// Stores
 	const storeThemGenForm: Writable<FormTheme> = localStorageStore('storeThemGenForm', {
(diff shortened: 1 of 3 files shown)


niktek:
Hey @ak4zh ,

I'd be curious to know how the different color theories would work with the base design tokens (primary, secondary, tertiary, etc) and hence how they get consumed into the components themselves.  For instance monochromatic is already kind of implicit in the shades of each color token, but clearly would need to be consumed differently in the components or force end devs to override defaults of secondary and tertiary uses to use the shades from Primary to achieve a monochromatic theory.

Also, I know it's a draft, but the v2 branch is disappearing in less than a week, when it gets merged to dev+master.  I think it might be best to raise an actual discussion first because there are a lot of implications and approaches to this that need to be fleshed out before actual code implementation.  For that discussion it would be great to see something like example palettes of each theory, what tokens they would map to, how they would be consumed/reflected by components. Then we can look at whether it would have any effect on color checking, theme data formats etc.

Cheers,
Nik

ak4zh:
@niktek I will go ahead and create a discussion, thanks for pointing that out. Also I understand the v2 branch will vanish in a few days, I wouldn’t mind opening a new pull request with dev as the base branch.

I just created this draft pull request to have a live version so we can play around with it for discussion. I am not very good with colours selections and the current random generator does not generate any usable set of colours as it’s truly random.

I was just playing around with these settings and it seemed to generator some sensible colour combinations compared to truly random ones which can be used right away or can be further customised. So wanted to have something up for others to play around.

endigo9740:
Discussion should go here please. We have a LOT of changes planned and this will need to come as a focused up:
- https://github.com/skeletonlabs/skeleton/issues/1142

ak4zh:
@niktek To answer the other part about design token, I was thinking of ways to specify a near colour for each design token color. For example we can mention that the error color should be near red color but should one that syncs well with the selected color theory generated colors.

I am still thinking of how to achieve that but just shared because this does generates something better than random, and others who actually made this theme generator can help me with their inputs.

niktek:
Yep I'm seeing some situations where the randomisation appears better on the palette on screen case, but don't really flow through to how they would show up when consumed by the components themselves and then of course there are just some examples where they are all competing for darkness!

image

I've added your PR to the list already going in the issue @endigo9740 linked above so we can weigh all the changes together.

ak4zh:
@niktek Yes I see Monochromatic, Neutral and Pastel does not generate good options as it follows a light to dark variation, which we already have in design system.
The rest of the theories does generates somewhat usable colors.

endigo9740:
The issue with this PR is that it's working being completed in a place that no longer be maintained soon. Essentially whatever is happening between now and the v2 release next week, and then we'll hit a feature freeze until the standalone version is created.

Given that, we have this PR as reference, but will not be moving forward with it.

PR #1969, opened 2023-08-29
docs(Autocomplete): update Autocomplete event typing

leranjun:
## Description

This PR updates the event type for the Autocomplete component from `any` to `CustomEvent<AutocompleteOption>`.

The problem, though, is that the user needs to manually cast the values as e.g. string, since value is `unknown` in `AutocompleteOption`. Should there be a generic type `AutocompleteOption<T>` to let the user define what values look like?

## Changsets

## Checklist

### DIFF
--- a/.changeset/beige-boats-heal.md
+++ b/.changeset/beige-boats-heal.md
@@ -1,0 +1,5 @@
+---
+"skeleton.dev": patch
+---
+
+chore: update event typing for Autocomplete component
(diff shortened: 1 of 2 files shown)


endigo9740:
I'll let @AdrianGonz97 review this as he's a bit more Typescript savvy than myself. But looks good at a glance!

AdrianGonz97 (approved):
Looks great! Just needed to delete the changeset as we don't use them when we add changes to the docs site. Thanks!

PR #1990, opened 2023-09-01
bugfix: added missing source code in input group examples

evccyr:
## Linked Issue

None

## Description

One of the forms' examples was missing the some stub code. I added that by copying it straight from the page source code.

## Changsets

---
"skeleton.dev": major
---

bugfix: added missing source code to tailwind forms example

## Checklist

### DIFF
--- a/.changeset/tall-mice-exercise.md
+++ b/.changeset/tall-mice-exercise.md
@@ -1,0 +1,5 @@
+---
+"skeleton.dev": major
+---
+
+bugfix: added missing source code to tailwind forms example
--- a/.changeset/tall-mice-exercise.md
+++ b/.changeset/tall-mice-exercise.md
@@ -1,5 +1,0 @@
----
-"skeleton.dev": major
----
-
-bugfix: added missing source code to tailwind forms example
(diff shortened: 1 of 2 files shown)


endigo9740:
@evccyr seems like a useful add, but there's a couple errors with providing the code snippets verbatim from the page. Mainly, you've got documentation specific details in place that won't be relevant to end users:

- Not everyone will have Font Awesome installed, so you need to replace icon references using`<i>` tags
- Replace anchor `href` links like `href="/elements/forms"` with `href="/"`

I might also suggest labeling which snippet is which with a `<p>` tag above each snippet.

I see a few instances of these issues. If you can patch these up we'll be glad to merge!

evccyr:
question: When is an `<i>` tag supposed to be replaced by `(segment)`?
I noticed that the [only example code](https://github.com/skeletonlabs/skeleton/blob/72bd820dc65fefa61523a7d71eef6feed354dfed/sites/skeleton.dev/src/routes/(inner)/elements/forms/%2Bpage.svelte#L511C3-L511C3) had the `(segment)` placeholder for an `<i>` tag. I want to know what it means and when is someone supposed to use that instead of an `(icon)` placeholder.

endigo9740:
Not sure where `(segment)` comes into play, just replace any/all reference to `<i>` with `(icon)` if you're trying to represent the featured example.

I meant, just literally put a `<p>` tag above each code block. I was using the term "label" loosely here.

PR #2021, opened 2023-09-08
feat/Autocomplete Generics

### LINKED ISSUE #1188: Use generics for types where possible

Initially thought of this for the `Autocomplete` component, but after [discussion on Discord](https://discord.com/channels/1003691521280856084/1003699523522142399/1087811314526720091), it seems this may be a more sweeping change.

[Svelte has support for generic types](https://github.com/sveltejs/rfcs/pull/38). We can constrain them similar to `<T extends S>` in a generic:
```ts
type GenericType = $$Generic<{value: string}>;
export let selected: GenericType | undefined;
```
Now if we had a list of items of type `B = { value: string;  foo: Bar }`, and the user sets `bind:selected` on the parent, the type of `selected` would be `B`, instead of just `{value: string}`

Enhancement

_No response_

HugeLetters:
## Linked Issue

Closes #1188 

## Description

Fixes generics on values and other props in Autocomplete component

## Changsets

feat: autocomplete generics

## Checklist

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,5 @@
 {
-    "cSpell.language": "en"
+  "cSpell.language": "en",
+  "prettier.singleQuote": true,
+  "prettier.semi": true
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,3 @@
 {
-  "cSpell.language": "en",
-  "prettier.singleQuote": true,
-  "prettier.semi": true
+  "cSpell.language": "en"
 }
--- a/packages/skeleton/src/lib/components/Autocomplete/Autocomplete.svelte
+++ b/packages/skeleton/src/lib/components/Autocomplete/Autocomplete.svelte
@@ -1,15 +1,19 @@
 <script lang="ts" context="module">
 	import { slide } from 'svelte/transition';
-	import { type Transition, type TransitionParams, prefersReducedMotionStore } from '../../index.js';
+	import { prefersReducedMotionStore, type Transition, type TransitionParams } from '../../index.js';
 	import { dynamicTransition } from '../../internal/transitions.js';
 
 	// eslint-disable-next-line @typescript-eslint/no-unused-vars
 	type SlideTransition = typeof slide;
 	type TransitionIn = Transition;
 	type TransitionOut = Transition;
+	type Value = unknown;
 </script>
 
-<script lang="ts" generics="TransitionIn extends Transition = SlideTransition, TransitionOut extends Transition = SlideTransition">
+<script
+	lang="ts"
+	generics="Value = unknown, TransitionIn extends Transition = SlideTransition, TransitionOut extends Transition = SlideTransition"
+>
 	import { createEventDispatcher } from 'svelte';
 	// import { flip } from 'svelte/animate';
 	// import {slide} from 'svelte/transition';
@@ -19,33 +23,33 @@
 
 	// Event Dispatcher
 	type AutocompleteEvent = {
-		selection: AutocompleteOption;
+		selection: AutocompleteOption<Value>;
 	};
 	const dispatch = createEventDispatcher<AutocompleteEvent>();
 
 	// Props
 	/**
 	 * Bind the input value.
-	 * @type {unknown}
+	 * @type {Value | undefined}
 	 */
-	export let input: unknown = undefined;
+	export let input: Value | undefined = undefined;
 	/**
 	 * Define values for the list.
-	 * @type {AutocompleteOption[]}
+	 * @type {AutocompleteOption<Value>[]}
 	 */
-	export let options: AutocompleteOption[] = [];
+	export let options: AutocompleteOption<Value>[] = [];
 	/** Limit the total number of suggestions. */
 	export let limit: number | undefined = undefined;
 	/**
 	 * Provide allowlist values.
-	 * @type {unknown[]}
+	 * @type {Value[]}
 	 */
-	export let allowlist: unknown[] = [];
+	export let allowlist: Value[] = [];
 	/**
 	 * Provide denylist values.
-	 * @type {unknown[]}
+	 * @type {Value[]}
 	 */
-	export let denylist: unknown[] = [];
+	export let denylist: Value[] = [];
 	/** Provide a HTML markup to display when no match is found. */
 	export let emptyState = 'No Results Found.';
 	// Props (region)
@@ -90,20 +94,20 @@
 	// Local
 	$: listedOptions = options;
 
-	function filterByAllowDeny(allowlist: unknown[], denylist: unknown[]) {
+	function filterByAllowDeny(_allowlist: typeof allowlist, _denylist: typeof denylist) {
 		let _options = [...options];
 		// Allowed Options
-		if (allowlist.length) {
-			_options = _options.filter((option) => allowlist.includes(option.value));
+		if (_allowlist.length) {
+			_options = _options.filter((option) => _allowlist.includes(option.value));
 		}
 
 		// Denied Options
-		if (denylist.length) {
-			_options = _options.filter((option) => !denylist.includes(option.value));
+		if (_denylist.length) {
+			_options = _options.filter((option) => !_denylist.includes(option.value));
 		}
 
 		// Reset options
-		if (!allowlist.length && !denylist.length) {
+		if (!_allowlist.length && !_denylist.length) {
 			_options = options;
 		}
 
@@ -125,7 +129,7 @@
 		return _options;
 	}
 
-	function onSelection(option: AutocompleteOption) {
+	function onSelection(option: AutocompleteOption<Value>) {
 		/** @event {AutocompleteOption} selection - Fire on option select. */
 		dispatch('selection', option);
 	}
--- a/packages/skeleton/src/lib/components/Autocomplete/Autocomplete.svelte
+++ b/packages/skeleton/src/lib/components/Autocomplete/Autocomplete.svelte
@@ -8,11 +8,13 @@
 	type TransitionIn = Transition;
 	type TransitionOut = Transition;
 	type Value = unknown;
+	type Meta = unknown;
 </script>
 
 <script
 	lang="ts"
-	generics="Value = unknown, TransitionIn extends Transition = SlideTransition, TransitionOut extends Transition = SlideTransition"
+	generics="Value = unknown, Meta = unknown, 
+	TransitionIn extends Transition = SlideTransition, TransitionOut extends Transition = SlideTransition"
 >
 	import { createEventDispatcher } from 'svelte';
 	// import { flip } from 'svelte/animate';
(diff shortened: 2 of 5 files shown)


.vscode/settings.json
HugeLetters, line 1: I had to add this - otherwise my default Prettier settings would change formatting

endigo9740:
Thanks for submitting the draft @HugeLetters. @AdrianGonz97 and @ryceg are our resident Typescript experts, so I'm going to ping them here to provide feedback asap.

To make sure we're on the same page, I [mentioned this in the main thread](https://github.com/skeletonlabs/skeleton/issues/1188#issuecomment-1712139388) this might be something we want to extend to Modals and other Skeleton features too. I'm fine if we want to take a piecemeal approach to that, along as these aren't lost in the shuffle.

.vscode/settings.json
endigo9740: @niktek I'll let you review this specifically

niktek: How did you format ? The formatting should be done via package script i.e. `pnpm format` which uses the project settings and not VSCode.  Put another way, 2.8k commits in we haven't needed those settings before.

HugeLetters: > How did you format ? The formatting should be done via package script i.e. `pnpm format` which uses the project settings and not VSCode. Put another way, 2.8k commits in we haven't needed those settings before.

I have a format on save turned on - so either I need to disable it in workspace settings which would also go to that commit or I need to change my default Prettier settings. 
So for example I can't check-out your script files w/o formatting them.

HugeLetters:
> Thanks for submitting the draft @HugeLetters.
> 
> @AdrianGonz97 and @ryceg are our resident Typescript experts, so I'm going to ping them here to provide feedback asap.
> 
> To make sure we're on the same page, I [mentioned in the main thread](https://github.com/skeletonlabs/skeleton/issues/1188#issuecomment-1712139388) this approach might be something we want to extend to Modals and other Skeleton features too. I'm fine if we want to take a piecemeal approach to that, as long as these aren't lost in the shuffle.

Sure, once this gets approved and I get a better understanding of kind of code style is appreciated here I'll try to work on other components too. 
Btw, what should the procedure be on minor improvements? Like [here](https://github.com/HugeLetters/skeleton/blob/620aaafa697043fd099c7abc3585e1b0b12b546c/packages/skeleton/src/lib/components/Autocomplete/Autocomplete.svelte#L98) there's an unnecessary array copying or [here](https://github.com/HugeLetters/skeleton/blob/620aaafa697043fd099c7abc3585e1b0b12b546c/packages/skeleton/src/lib/components/Autocomplete/Autocomplete.svelte#L155) when there's no limit set.

endigo9740:
> Btw, what should the procedure be on minor improvements? ...

We keep scope small and limited in PRs. This helps keep changes localized and not grow too large. For additional issues it's worth creating new issues and dedicated PRs for each change.

I will also advise you to browser our issue backlog. There's a number of issues folks have with the autocomplete component specifically. It's a feature implemented by me, but not one I've had a lot of prior experience with, so I think there's definitely some room for improvement. At this point we're reaching a point a more consorted effort to improve the feature top to bottom might be in order.

If you're not already on Discord, I might recommend joining and jumping into the `#contributor` channel. This is a great place to discuss and plan bigger ideas. That or creating a Discussions section. Once a set plan is established we can generate one or more issues against that.

AdrianGonz97 (changes requested):
I love the proposed changes, great job! I pushed a couple of changes when it came to annotating types and some variable names.

One thing I wanted to mention is that since we're already modifying the type of `value` into a generic, we may as well do the same for `meta`. 

Additionally, `keywords` looks to accept a `string` type, despite being labeled as `any`. Note the comment for it: `/** Provide a comma separated list of keywords. */`. What are your thoughts on this @endigo9740? I looked back in the original PR that introduced the component and didn't see any conversations about keeping `keywords` as `any`, so I'm assuming this is an `any` type that slipped between the cracks.

endigo9740:
@AdrianGonz97 Per `keywords`, yes those should be a `string` ideally. Offhand I can't recall if we're parsing those, but they will be read as a string in the end.

AdrianGonz97:
@endigo9740 Thanks! I thought as much. Looking at the code, we're not parsing it at all and the only time it's used is when everything is being strinigified for search - see [here](https://github.com/skeletonlabs/skeleton/blob/f7a4d2331815b9bb59b0c0adf1d6edd7ef09eeee/packages/skeleton/src/lib/components/Autocomplete/Autocomplete.svelte#L121).

In that case, I think it's safe to change that `any` to a `string` then.

.vscode/settings.json
HugeLetters: > Can't say that I'm a fan of this particular change. I think there's a better way to go about this rather than having 2 sources of truth for prettier. I have an idea for this, but for now I'll suggest reverting this change from the PR and we can address this in a separate PR.

Sure, I can just not include these settings changes in my PR - wasn't sure about the best way to approach this but a formatting config enforced directly by your repo would be nice I think.

HugeLetters:
> I love the proposed changes, great job! I pushed a couple of changes when it came to annotating types and some variable names.
> 
> One thing I wanted to mention is that since we're already modifying the type of `value` into a generic, we may as well do the same for `meta`.
> 
> Additionally, `keywords` looks to accept a `string` type, despite being labeled as `any`. Note the comment for it: `/** Provide a comma separated list of keywords. */`. What are your thoughts on this @endigo9740? I looked back in the original PR that introduced the component and didn't see any conversations about keeping `keywords` as `any`, so I'm assuming this is an `any` type that slipped between the cracks.

Working on making Meta generic.

Also a bit of a problem with Value generic are scenarios like this
image
Clearly here I would input to be the source of truth for the generic but TS thinks options takes precedence - I honestly have no idea if it's possible to somehow enforce this behavior. But I will say I don't think it's anything more than annoyance since you can avoid this with explicit typing.
Specifically this error goes away if you just move input to a variable instead of inline(as you should).

And also I have to update your code in the docs like this after adding a generic for Meta
```ts
type FlavorOption = AutocompleteOption<string, { healthy: boolean }>;
const flavorOptions: FlavorOption[] = [
	{ label: 'Vanilla', value: 'vanilla', keywords: 'plain, basic', meta: { healthy: false } },
	{ label: 'Chocolate', value: 'chocolate', keywords: 'dark, white', meta: { healthy: false } },
	{ label: 'Strawberry', value: 'strawberry', keywords: 'fruit', meta: { healthy: true } },
	{ label: 'Neapolitan', value: 'neapolitan', keywords: 'mix, strawberry, chocolate, vanilla', meta: { healthy: false } },
	{ label: 'Pineapple', value: 'pineapple', keywords: 'fruit', meta: { healthy: true } },
	{ label: 'Peach', value: 'peach', keywords: 'fruit', meta: { healthy: true } }
];
```
Because otherwise TS infers generic Meta type as `{healthy: false}` and then gives error on every `{health: true}` instead of inferring it as a union.

packages/skeleton/src/lib/components/Autocomplete/Autocomplete.svelte
```diff
@@ -8,22 +8,25 @@
...
+	generics="Value = unknown, Meta = unknown, 
+	TransitionIn extends Transition = SlideTransition, TransitionOut extends Transition = SlideTransition"
 >
 	import { createEventDispatcher } from 'svelte';
 	// import { flip } from 'svelte/animate';
 	// import {slide} from 'svelte/transition';
 
 	// Types
 	import type { AutocompleteOption } from './types.js';
+	type Option = AutocompleteOption<Value, Meta>;
```
HugeLetters, line 25: Would love a suggestion if there's a better name for this, haha. 
Since this is reused so much throughout the component I've extracted it to a separate type.

packages/skeleton/src/lib/components/Autocomplete/types.ts
```diff
@@ -1,12 +1,12 @@
...
 	/** Provide a unique display label per option. Supports HTML. */
 	label: string;
 	/** Provide a unique option value. */
 	value: Value;
 	/** Provide a comma separated list of keywords. */
-	keywords?: any;
+	keywords?: string;
 	/** Pass arbitrary data per option. */
-	meta?: any;
+	meta?: Meta;
```
HugeLetters, line 11: A DX improvement could be made that if meta value is provided then it can't be undefined like so
```ts
export type AutocompleteOption<Value = unknown, Meta = unknown> = {
	/** Provide a unique display label per option. Supports HTML. */
	label: string;
	/** Provide a unique option value. */
	value: Value;
	/** Provide a comma separated list of keywords. */
	keywords?: string;
} & ([undefined] extends [Meta]
	? {
			/** Pass arbitrary data per option. */
			meta?: Meta;
	  }
	: {
			/** Pass arbitrary data per option. */
			meta: Meta;
	  });
```

Not sure it that's necessary though and even if it is - if there's a nicer way to do that. Don't think there's a way to do it with interfaces.

sites/skeleton.dev/src/routes/(inner)/components/autocomplete/+page.svelte
```diff
@@ -39,15 +39,15 @@
 
 	let inputDemo = '';
 	let inputAllowlist = '';
-	type FlavorOption = AutocompleteOption<string>;
-	const flavorOptions = [
```
HugeLetters: Had to change this because otherwise TS would infer `Meta` generic type as `{healthy: false}` instead of `boolean` and then would error that other meta tags do not satisfy the type.

endigo9740:
Hey @AdrianGonz97 not sure if you saw but @HugeLetters had a few follow up questions above. If we can get these addressed and give me a thumbs up when this is approved and ready to merge.

packages/skeleton/src/lib/components/Autocomplete/Autocomplete.svelte
AdrianGonz97: Given that the value it represents is named `options`, I think this is perfectly fine 😄

packages/skeleton/src/lib/components/Autocomplete/types.ts
AdrianGonz97: I don't think that'll be necessary since it'll likely be a breaking change for those that vary the type of their `meta` field. 

For instance, consider the following:
```ts
// note that we're not specifying a Meta type, so it defaults to `unknown`
type FlavorOption = AutocompleteOption<string>;
const flavorOptions: FlavorOption[] = [
	{ label: 'Vanilla', value: 'vanilla', keywords: 'plain, basic', meta: { healthy: false } },
	{ label: 'Chocolate', value: 'chocolate', keywords: 'dark, white', meta: { healthy: false } },
	{ label: 'Peach', value: 'peach', keywords: 'fruit' }
];
```
The last value in the array (Peach) will have a type error, despite `meta` being optional. It can be fixed by specifying the type for Meta in `AutocompleteOption`, but the default behavior will break.

Ideally, we'd want the default generic type of `Meta` to be `undefined`, but that's also going to result in breaking the types for people that have already implemented the `AutocompleteOption` type.

So for now, let's just keep it as is and I'll note that it should be changed to `undefined` in the next major version of Skeleton.

AdrianGonz97 (approved):
Excellent changes and thanks for the update! 

I've pushed a small change that adjusts the type for `options` so that it's appearance in the docs site makes a little more sense. And with that, everything else looks great and I think this PR is ready to be merged.

HugeLetters:
> Excellent changes and thanks for the update!
> 
> I've pushed a small change that adjusts the type for `options` so that it's appearance in the docs site makes a little more sense. And with that, everything else looks great and I think this PR is ready to be merged.

Great! Somebody needs to authorize Vercel's deployment to follow through.

PR #2023, opened 2023-09-08
docs: Fixed theme generator's `mono` font

AdrianGonz97:
`Courier New` was using a double quote instead of a single, so that's been changed.

I also made the output string for fonts use template literals instead of either a single or double quotes so that we can avoid the above issue entirely.


endigo9740:
Good catch! I mean to check the theme generator earlier and it slipped my mind. Thanks!

PR #2033, opened 2023-09-12
chore: Replaced vercel analytics for fathom

AdrianGonz97:
## TODO

- [ ] Register with [usefathom](https://app.usefathom.com/register)
- [ ] Acquire Fathom URL and ID
- [ ] Add `PUBLIC_FATHOM_URL` and `PUBLIC_FATHOM_ID` to Vercel's ENV vars


endigo9740:
@AdrianGonz97 this is still on my radar, but likely not something I'll have time for today. I'll aim to review later this week though.

endigo9740:
@AdrianGonz97 I'm going to ahead and make the decisions to put this one on ice for now. The goal will be to revisit the Skeleton docs alongside the "Next" stuff, so let's revisit this then and similar solutions then. We'll likely look to move to Cloudfront then and their analytics solution.

Thanks for the find and quick turn around on this though!

Issue #1381, opened 2023-04-26
Show how to reduce bundle size when using highlight.js in CodeBlocks


treo:
### Link to the Page

https://www.skeleton.dev/utilities/codeblocks

### Describe the Issue

When using highlight.js as described, it results in a huge bundle size increase (+1MB). 

It would be good to document a different way of loading it, such that languages can be loaded dynamically.

My crude workaround for it looks like the following, but that is bad in many different ways. And shouldn't be copied. Unfortunately, I don't know of a better way to make it work. Dynamically loading from a library isn't supported by vite, so I used a public CDN for it, but that comes with its own problems, like being bound to a particular version.

Layout.svelte:
```
	import hljs from 'highlight.js/lib/core';
	import plaintext from 'highlight.js/lib/languages/plaintext';
	import 'highlight.js/styles/tomorrow-night-bright.css';
	import { storeHighlightJs } from '@skeletonlabs/skeleton';

	hljs.registerLanguage('plaintext', plaintext);
	storeHighlightJs.set(hljs);
```

CodeRenderer.svelte:
```
<script lang="ts">
    import hljs from 'highlight.js/lib/core';

    import { CodeBlock } from '@skeletonlabs/skeleton';
    import {browser} from "$app/environment";
	export let lang;
	export let text;

    $: actualLang = lang ? lang : 'plaintext'
    let loadedLangs = {
        plaintext: true
    }
    $: loadLanguage(actualLang)

    async function loadLanguage(actualLang) {
        if (!loadedLangs[actualLang] && browser) {
            try {
                /* @vite-ignore */
                const languageModule = await import(`https://cdnjs.cloudflare.com/ajax/libs/highlight.js/11.7.0/es/languages/${actualLang}.min.js`);
                hljs.registerLanguage(actualLang, languageModule.default);
                loadedLangs[actualLang] = true;
            }catch (e) {
                console.info(`Language ${actualLang} not found for highlight.js`);
            }
        }
    }
</script>

{#if loadedLangs[actualLang]}
    <CodeBlock language={actualLang} code={text} />
{:else}
    <CodeBlock language="plaintext" code={text} />
{/if}
```

### Are you able to create a Pull Request with the fix?

None

endigo9740:
Hmm I haven't noticed bundle sizes reaching quite that scale. In fact the entire Skeleton documentation site, which utilizes nearly every feature we provide, is about 612 kb total.

[...]

PR #2047, opened 2023-09-14
Improve code blocks docs

### LINKED ISSUE #2024: Syntax highlighting in code blocks breaks with purgeCss
### Link to the Page

https://www.skeleton.dev/utilities/codeblocks#configure-your-project

The "**Configure Your Project**" section may benefit from an example of how to configure `PurgeCSS` correctly when importing a `highlight.js` style sheet. Especially because create-skeleton-app now uses `PurgeCSS` by default. 

image

The correct configuration is already documented in the [PurgeCSS section](https://www.skeleton.dev/docs/purgecss#safelisting). Without safe-listing classes with the `hljs-` prefix, syntax highlighting won't work in the bundled app. When using the vite dev server though, highlighting works as expected. This makes this harder to catch but is easily preventable by improving the documentation.  

I'd be happy to fix this and open a PR.

moritzhertler:
## Linked Issue

Closes #2024 and #1381 

## Description

Improves the Code Blocks documentation by adding an example of configuring PurgeCss and a note on reducing bundle size. 

## Changsets

## Checklist

### DIFF
--- a/packages/skeleton/src/lib/utilities/CodeBlock/CodeBlock.svelte
+++ b/packages/skeleton/src/lib/utilities/CodeBlock/CodeBlock.svelte
@@ -56,6 +56,7 @@
 	function languageFormatter(lang: string): string {
 		if (lang === 'js') return 'javascript';
 		if (lang === 'ts') return 'typescript';
+		if (lang === 'shell') return 'terminal';
 		return lang;
 	}
 
--- a/sites/skeleton.dev/src/routes/(inner)/docs/contributing/requirements/+page.svelte
+++ b/sites/skeleton.dev/src/routes/(inner)/docs/contributing/requirements/+page.svelte
@@ -190,9 +190,9 @@
 		<p>
 			To check for linting issues via <a class="anchor" href="https://prettier.io/" target="_blank" rel="noreferrer">Prettier</a>, run:
 		</p>
-		<CodeBlock language="console" code={`pnpm lint`} />
+		<CodeBlock language="shell" code={`pnpm lint`} />
 		<p>To automatically apply suggested linting changes, run:</p>
-		<CodeBlock language="console" code={`pnpm format`} />
+		<CodeBlock language="shell" code={`pnpm format`} />
 
 		<!-- Automated Tests -->
 		<h3 class="h3">Automated Tests</h3>
@@ -200,7 +200,7 @@
 			Tests are handled via <a class="anchor" href="https://vitest.dev/" target="_blank" rel="noreferrer">Vitest</a>, which is similar to
 			Jest. Make sure to run all tests before submitting new pull requests.
 		</p>
-		<CodeBlock language="console" code={`pnpm test`} />
+		<CodeBlock language="shell" code={`pnpm test`} />
 
 		<!-- Spell Checking -->
 		<h3 class="h3">Spell Checking</h3>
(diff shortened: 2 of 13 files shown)


moritzhertler:
What do you think about these changes?

Would you like me to add a changeset for these documentation changes? 
Should the second change (about reducing bundle size) be a separate PR?

Sarenor:
Seems like a nicely contained set of changes and since it just changes the docs, no changeset should be necessary.

endigo9740:
@moritzhertler I'll be aiming to run through pending PR reviews tomorrow. But looks good at a glance. I'll aim to provide feedback tomorrow if anything else needs to change.

endigo9740:
@moritzhertler to help practice what we preach, I've updated the documentation to show how to import all the common languages we use on the Skeleton docs: html, css, js, ts, shell. Additionally, I've updated the Skeleton documentation site to implement these changes as well, which should reduce the bundle size of our project as well.

Note that in order to properly support what was the "console" alias I've opted for "shell" instead. But rename this to `terminal` in the UI to make it more obvious where this is used.

Feel free to give this a once over and if you're happy with the changes, then I'm happy to merge. This should prove to be a very welcome improvement to our overall documentation bundle size!

@niktek pinging you so you might review how you might want to set this up via the CLI. In case you want to bring back the auto-install prompt option. I'd suggest using only the common languages I used as being the defaults, but it does require a good bit more boilerplate.

moritzhertler:
Thanks, looks good @endigo9740! I've added two improvements:
- `json` is used on the introduction page but was not imported/added which caused an error when opening the accordion.
- The `blog/[slug]` page also imports `hljs` including all supported languages which drastically increased the size of the corresponding bundle. I changed the import to the highlights.js core and checked all blog posts to make sure no language was missing. 

> Additionally, to help practice what we preach, I've updated the Skeleton doc project to implement these changes, which should reduce the bundle size as well.

I've done this as well and was just about to ask if you were interested in me opening another PR regarding this😄

Here is what I found:
Previously, Vite bundled all of highlight.js into a separate chunk which was ~913kB in size. The highlight.js core is now bundled into a chunk of ~21kB and the languages are moved to the main layout node (0.[hash].js) which is now ~22kb larger. This results in the same number of requests but an overall reduction in bundle size of ~870kB🎉  

> Note that I was having issues with the "console" alias, so I've updated all examples to use "shell" instead. I also rename this to `terminal` in the component UI to make it more obvious the intended use case. Somehow there

A few of the blog posts still use console instead of shell. I don't know if the issue you encountered applies here as they do not use the CodeBlock component.

moritzhertler:
While looking through the highlight.js docs, I have just noticed that languages can depend on each other. This is done by a simple string reference in the language function. Importing a language does **not** also import the dependent languages. 

I have gone through the languages we are using and the following dependent languages are missing:
- `bash` (used by `shell`)
- `handlebars` (used by `xml`)
- `graphql` (used by `javascript`)

Using an unregistered language directly with the `CodeBlock` component results in a runtime error. This is **NOT** the case with these dependent languages. I tested this with a GraphQL template string in a JavaScript code block. The missing dependency simply results in unformatted text, not in a runtime error. 

Should we add the missing languages? 

This more and more feels a bit hard to maintain. Not only does every language have to be imported manually but also the dependent languages have to be looked up manually. Maybe there is a better solution?

endigo9740:
@moritzhertler thank you for the extra set of eyes on this. Those are some great catches! I appreciate you making the changes directly as well.

> A few of the blog posts still use "console" instead of "shell". I don't know if the issue you encountered applies here as they do not use the CodeBlock component.

If they are not using CodeBlocks they should be fine. If you can link me to a specific post I'm happy to take a look at what's up. I figure it's just snippets generated through our CMS (we're using Ghost's headless API).

> While looking through the highlight.js docs, I have just noticed that languages can depend on each other. This is done by a simple string reference in the language function. Importing a language does not also import the dependent languages.

That's an interesting gotcha. Let's maybe circle back to expand the docs with a note about this and give folks a heads up in the near future. In order to get his update out next week I'm going to go ahead and merge what we have now. But I'll give our support team a heads up in case this comes up before then.

PR #2051, opened 2023-09-17
bugfix/modal-transition-layer-hardcoded-padding

### LINKED ISSUE #2050: Modals have hardcoded padding on transition element
### Current Behavior

image

As reported by Joost on Discord.

Let's review changing where this padding is set to ensure it's configurable per-instance. This is making it difficult for end users to create full screen components that fill the entire screen.

Mahmoud-zino:
## Linked Issue

Closes #2050

## Description

Added `TransitionLayerPadding` prop to enable full-screen modals.

## Changsets

bugfix: Added `transitionLayerPadding` prop in Modal to enable full-screen modals.

## Checklist

### DIFF
--- a/.changeset/curly-parrots-look.md
+++ b/.changeset/curly-parrots-look.md
@@ -1,0 +1,5 @@
+---
+"@skeletonlabs/skeleton": patch
+---
+
+bugfix: Added `transitionLayerPadding` prop in Modal to enable full-screen modals.
--- a/packages/skeleton/src/lib/utilities/Modal/Modal.svelte
+++ b/packages/skeleton/src/lib/utilities/Modal/Modal.svelte
@@ -50,6 +50,8 @@
 	export let shadow: CssClasses = 'shadow-xl';
 	/** Provide a class to override the z-index */
 	export let zIndex: CssClasses = 'z-[999]';
+	/** Provide classes to style the transition layer padding. */
+	export let transitionLayerPadding = 'p-4';
 
 	// Props (buttons)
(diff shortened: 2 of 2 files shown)


endigo9740:
@Mahmoud-zino I'm limited on time today so I won't be able to review this right now, but will try to circle back later this week. Thanks!

endigo9740:
@Mahmoud-zino I'm not sure this is going to work - it doesn't address the core issue in the manner suggested in the original ticket:

> Let's review changing where this padding is set and ensure it's configurable per-instance.

The padding remains set in the same place (on the transition layer) and the value cannot be controlled per instance - only globally with the component prop.

To be honest, I don't like that the transition layer exists at all. I understand why it was added, but it really complicates the layout here. The transition layer should handle ONLY that, transition values - but instead it's having to adapt portions of the visual layout that were previously handled by the backdrop. So styles are split between the backdrop and transition layer in this really awkward manner.

Overall I think this component needs a proper top down review and possible rewrite of the markup and styles specifically to properly handle all scenarios needed for modals, including full screen modals. These small changes are not taking the "big picture" view that we need to really solve the core issues here. So I suggest we do the following:

1. Close this PR - we don't want to introduce props that may or may not persist long term
2. Review our Issue backlog and gather all tickets related to full screen modals - linked them into a single ticket
3. Let me run through and handle the updated markup/styles to address theses issues in a definitive manner

I'd welcome your help in gathering those tickets together - I believe there's 2-3 off hand.

Mahmoud-zino:
@endigo9740 this makes sense, I misunderstood the request previously. 👍

PR #2064, opened 2023-09-21
feat/Modal component - better typing support

### LINKED ISSUE #1813: Code completion in ModalComponent

We can provide code completion by using Typescript generics in `ModalComponent`:
```ts
import { ComponentProps, SvelteComponent } from "svelte"

export declare type Constructor<T> = new (...args: any[]) => T;

export interface ModalComponent<T extends SvelteComponent> {
	/** Import and provide your component reference. */
	ref: Constructor<T>;
	/** Provide component props as key/value pairs. */
	props?: ComponentProps<T>;
	/** Provide an HTML template literal for the default slot. */
	slot?: string;
}
```

Enhancement

Code is taken from https://github.com/bryanmylee/svelte-render/blob/main/src/lib/createRender.ts

HugeLetters:
## Linked Issue

Closes #1813

## Description

This PR aims to improve typing on a modal component.
 - [x] response value type is restricted based on modal type
   - side note: with prompt modals the type is only a string. Even though an `input type="number"` can be provided through `vallueAttr` - Svelte will not parse its value as a number because its compiler only sees `type="text"` specified in the Modal component.
 - [x] triggering a component modal infers props type
   - [x] also for registering components on a Modal  
 - [ ] restrict modal trigger properties based on a modal type? For example: can't provide image if modal type is component - since it does nothing anyway, can't provide component if modal type is not a component etc.
 - [x] vallueAttr is typed with `HTMLInputAttributes`
 - [ ] autocomplete for registered components keys? At first glance seems impossible - Modal component and modalStore are not coupled with each other. Only if we change the API so that you register modal components through store, not Modal component maybe?

## Changsets

## Checklist


endigo9740:
@HugeLetters I've made a minor edit to the tagged issue above. Make sure to remove the `{}` quotes or this won't link properly and the original issue will not be closed automatically when this PR is merged.

HugeLetters:
@endigo9740 could you clear something up for me please? Is the purpose of the modalStore is to just track a queue of modals if a multiple of them has been triggered (like [here](https://www.skeleton.dev/utilities/modals#trigger:~:text=Alert%20Modal-,Queue%20Multiple,-Confirm%20Modal))? Is it also for nested modals or those are not supported? Or is there some other purpose?

endigo9740:
@HugeLetters in simplest terms the modal system is implemented globally in your app. It's not components created on-demand. We use a singleton pattern for this. There's a single instance of `<Modal />` that is placed in the root layout, then the queue is managed via a Svelte store. Triggering a modal is adding to the queue, closing a modal is removing from the queue.

HugeLetters:
> @HugeLetters in simplest terms the modal system is implemented globally in your app. It's implemented on-demand page to page. We use a singleton pattern for this. There's a single instance of `<Modal />` that is placed in the root layout, then the queue is managed via a Svelte store. Triggering a modal is adding to the queue, closing a modal is removing from the queue. It's first in, first out.
> 
> We ship with a handful of canned options that replace native dialogs, such as alert, prompt, confirm. Then provide a means for creating your own custom modals via custom components. You can either pass them on-demand (when triggering) or register them ahead of time for reuse.

Yeah, I am asking specifically about [this](https://github.com/skeletonlabs/skeleton/blob/0187fdbc30af91c2b2fb8e6ed75f6407d9e37267/packages/skeleton/src/lib/utilities/Modal/stores.ts#L46) - so it's just to keep track of a modal queue? What's the purpose of queue? Is it just for sequential modals like in the link in my previous message?

endigo9740:
Correct, the role of the modal/drawer/toast stores are each to maintain their respective reactive list (read: queue) of modal instances. If multiple portions of the app trigger multiple dialogs, then they need to queue. For example, if a dialog is triggered while another dialog is being displayed.

> I am asking specifically about this

Also FYI, the way these stores are initialized changed in v2 in order to prevent some known issues with [SvelteKit SSR](https://github.com/skeletonlabs/skeleton/wiki/SvelteKit-SSR-Warning). This change was implemented by @AdrianGonz97, who worked with us on your last PR, so he would be the guy to ask if you have any questions about this.

packages/skeleton/src/lib/utilities/Modal/Modal.svelte
```diff
@@ -218,11 +218,10 @@
 {#if $modalStore.length > 0}
 	{#key $modalStore}
 		<!-- Backdrop -->
-		<!-- FIXME: resolve a11y warnings -->
-		<!-- svelte-ignore a11y-no-static-element-interactions -->
 		<div
 			class="modal-backdrop {classesBackdrop}"
 			data-testid="modal-backdrop"
+			role="presentation"
```
HugeLetters, line 224: I think it's safe to set it like this since you can't use keyboard navigation to access backdrop - it's only for mk+b users and so it can be safely ignored by screen readers.

HugeLetters:
@endigo9740 I see there is an issue that you need to keep a track of properties to pass on - [here](https://github.com/skeletonlabs/skeleton/blob/724a88135dc9167689f1bdf2f197d3049b974060/packages/skeleton/src/lib/utilities/Modal/Modal.svelte#L186). 

We can do this to at least ensure we didn't forget to pass on any props to parent but it seems not all props should be passed on - what's the criteria? Anything that's not not related to Backdrop or Transition Layer? Docs claim that any Modal prop is available on the parent object - so maybe pass all of them through?
```ts
import type Modal from './Modal.svelte'; // you can import a module from itself
type ModalProps = Required<ComponentProps<Modal>>; // add Omit here to exclude unrelated props.
const parent: ModalProps = { ... };
```

HugeLetters:
@endigo9740 @AdrianGonz97 I would also like your input on this. 
https://github.com/skeletonlabs/skeleton/blob/0b7add15c16bd11cec313ce2d2176504e8ab266c/sites/skeleton.dev/src/lib/modals/examples/ModalExampleForm.svelte#L19

With my current changes this errors because the only type of response common between any modal type is `false|undefined` - so the error is correct since otherwise it would allow for unexpected behavior. Like you currently have a confirm modal active yet some other component sends you a string response through modal store even though you expected to receive only `boolean|undefined`. 
So if I keep my changes as is it would now require to first check the type of a modal before sending a response - but it would be a breaking change in a way, albeit a quite small one.

What approach would you recommend?

packages/skeleton/src/lib/utilities/Modal/Modal.svelte
endigo9740: I'd be cautious about adding a11y features in this PR, that's likely out of scope. Though I do agree we should likely include a role here. If we do, we probably need a means to set it globally and override per instance. Maybe create a new ticket for this?

A11y attributes can be tricky - we don't often add them without proper research. Essentially no attributes are typically better than wrong attributes when it comes to accessibility.

endigo9740:
@HugeLetters per the inquiry around `parent`. To help explain how that functions, this is a sort of duct tape fix we implemented to allow custom component modals be able to inherit and use properties set on the parent `<Modal />` component. Since they're props, they may or may not be set by the end user.

I'm not sure if you've looked at the changes Svelte 5 is introducing, but my gut is telling me there will be a better way to handle this than what we're doing now. Typing for `parent` may be best suited to wait until then.

sites/skeleton.dev/src/routes/+layout.svelte
```diff
@@ -187,7 +176,14 @@
 </svelte:head>
 
 <!-- Overlays -->
-<Modal components={modalComponentRegistry} />
```
HugeLetters, line 190: Stuff "has" to be moved inline - otherwise you don't get the convenience of generics.

packages/skeleton/src/lib/utilities/Modal/types.ts
```diff
@@ -1,45 +1,57 @@
...
-	meta?: any;
-}
+type ModalResponseRecord = {
+	alert: never;
+	confirm: boolean;
+	prompt: string;
+	component: any;
+};
+export type ModalType = keyof ModalResponseRecord;
+export type ModalSettings<Type extends ModalType = ModalType, Component extends SvelteComponent = SvelteComponent> = Type extends Type
```
HugeLetters, line 24: `Types extends Type` makes sure we get a union of objects instead of a single object with properties types as unions.
We want this
```ts
{ a: 1, b: 1 } | { a: 2, b: 2 }
```
not this
```ts
{ a: 1 | 2, b: 1 | 2 }
```

packages/skeleton/src/lib/utilities/Modal/types.ts
```diff
@@ -1,45 +1,57 @@
...
+			/** Provide arbitrary classes to the modal window. */
+			modalClasses?: string;
+			/** Override the Cancel button label. */
+			buttonTextCancel?: string;
+			/** Override the Confirm button label. */
+			buttonTextConfirm?: string;
+			/** Override the Submit button label. */
+			buttonTextSubmit?: string;
+			/** Pass arbitrary data per modal instance. */
+			meta?: any;
```
HugeLetters, line 55: This isn't used anywhere.

HugeLetters:
@AdrianGonz97 I think I'm more or less done for now - I've listed things I've done & considered. 
So once you have the time please let me know if anything else could be improved.

endigo9740:
> @AdrianGonz97 I think I'm more or less done for now - I've listed things I've done & considered. So once you have the time please let me know if anything else could be improved.

@AdrianGonz97 I'd welcome your review for this. If you give the thumbs up I'm happy to merge.

HugeLetters:
@endigo9740 I think Adrian gave his thumbs up with a reaction 😅

AdrianGonz97:
@HugeLetters heh, sorry not quite - I'll give it a review today! 😄

endigo9740:
@AdrianGonz97 I'm running through and reviews all pending PRs today, so I'll keep an eye out for your response! Thanks!

endigo9740:
@HugeLetters sorry for the delay on this one - I think @AdrianGonz97 and myself have both been swamped. We do still intend to review this, but given the larger scope of changes this will be a bit more involved. Thanks for your patience in the meantime.

HugeLetters:
It's alright, no hurry :)

sites/skeleton.dev/src/routes/+layout.svelte
AdrianGonz97: I don't particularly mind inlining props like this, but @endigo9740 may find this sacrilegious 😄. I'm fine with either.

sites/skeleton.dev/src/lib/modals/examples/ModalExampleForm.svelte
```diff
@@ -19,7 +19,7 @@
 
 	// We've created a custom submit function to pass the response and close the modal.
 	function onFormSubmit(): void {
-		if ($modalStore[0].response) $modalStore[0].response(formData);
+		if ($modalStore[0].response && $modalStore[0].type === 'component') $modalStore[0].response(formData);
```
AdrianGonz97, line 22: I'm afraid this may be the point where this change demonstrates that it's a breaking one for types.

AdrianGonz97 (commented):
cc: @endigo9740 @HugeLetters 

Overall, I really like the changes! While I pushed some really minor changes, this PR is pretty good as-is.

Unfortunately, I _believe_ that this PR may break some of the types for our users. For instance, in our example component `ModalExampleForm.svelte`, the following check needed to be added to discriminate on the Modal type, otherwise, we'd get a type-error when we try to pass `flavor` to `response`:
```ts
function onFormSubmit(): void {
	//                                                 👇
	if ($modalStore[0].response && $modalStore[0].type === 'component') 
		$modalStore[0].response(flavor);
	modalStore.close();
}
```

Now, whether changes to types should be considered 'breaking' is certainly a contentious topic, I believe that they should be. So while this PR is great as it is, I would hold off from merging until the next major version of Skeleton.

packages/skeleton/src/lib/utilities/Modal/Modal.svelte
```diff
@@ -153,18 +161,22 @@
 	}
 
 	function onClose(): void {
-		if ($modalStore[0].response) $modalStore[0].response(false);
+		if (currentModal.response) currentModal.response(false);
```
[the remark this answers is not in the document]
HugeLetters: I'm fine with either - I leave it up to you guys.

endigo9740:
> Now, whether changes to types should be considered 'breaking' is certainly a contentious topic, I believe that they should be. So while this PR is great as it is, I would hold off from merging until the next major version of Skeleton.

@AdrianGonz97 @HugeLetters Not to be the bearer of bad news, but if this is the case we might not be able to move forward with this PR (yet). v2 -> v3 will likely involve a major rewrite of this feature. If that's the case, we can continue to make updates, but only as reference for how we might handle this for the rewrite.

@HugeLetters if you would like to like to be a part of the rewrite for v3, I'm happy to bring you in for that. You and Adrian can help shape the feature as it's being written, if that makes sense. No pressure obviously, but consider it an open offer.

packages/skeleton/src/lib/utilities/Modal/Modal.svelte
```diff
@@ -153,18 +161,22 @@
...
 		modalStore.close();
 	}
 
 	function onConfirm(): void {
-		if ($modalStore[0].response) $modalStore[0].response(true);
+		if (currentModal.type !== 'confirm') return;
+
+		if (currentModal.response) currentModal.response(true);
```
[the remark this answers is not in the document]
HugeLetters: This is one isn't breaking change actually per se since this function is only called if modalType is "confirm" - unless we consider that it's type gets mutated somewhere between rerender and a callback call(impossible basically).

We could articulate this better if we introduce Modal sub-components to break-up the code a little bit - and this would eliminate the need to call this check.

HugeLetters:
I'm fine if this needs to be postponed until v3 - I'm not that familiar with what should be considered breaking and not. 

On one hand it is since there might exist a code base which would require some minor code changes with this update to avoid TS compilation errors(nothing changes behavior-wise).
On the other it only breaks stuff which wasn't intended to be used like that in the first place - you're not supposed to provide custom responses on confirm modals for example.

packages/skeleton/src/lib/utilities/Modal/Modal.svelte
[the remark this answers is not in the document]
AdrianGonz97: > This is one isn't breaking change actually per se since this function is only called if modalType is "confirm"

Sorry, I should've been more specific on this one. I meant that feel the same way as I did for the aside I wrote for using string literals instead of `true` or `false` for `onClose` and `onConfirm`. We would respond with `CLOSED` and `CONFIRMED`, respectively.

endigo9740:
Fair points @HugeLetters. FYI part of the reason for the v3 rewrite is Svelte 5 support. Our hope is this will greatly simplify a lot of things, which should make introducing better type safety a bit easier. At least in theory. Fingers crossed :)

endigo9740:
@HugeLetters @AdrianGonz97 Just a quick update after sleeping on this. I think we'll go ahead and bump this back to a draft state, I'll update to title to indicate this is not intended to be merged, and we'll use the lessons learned here to inform the design process for the v3 version of the modals.

Given all the other changes coming down the pipeline, I think we'll be in for a really nice update.

Huge, sorry for this taking so long. I suspected this might be the way this goes but wanted to make sure we vetted it properly to confirm. Your contribution is appreciated and I'll definitely ping you when we get to the new modal system. If you have time to jump in and review we would sure appreciate it!

Thanks guys!

PR #2124, opened 2023-10-07
Resolve Typescript Issues 1/2

### LINKED ISSUE #2115: Tech Debt: Fix all "any" types used throughout the docs
### Link to the Page

_No response_

When running ``pnpm lint`` you get a bunch of "any" warnings, this should be fixed.

jwatts777:
## Linked Issue

Closes #2115 

## Description

Cleanup and remove `any` types from documentation sections

## Changsets

## Checklist


jwatts777:
[This is the most updated list](https://github.com/skeletonlabs/skeleton/issues/2115#issuecomment-1751511685)
Calling it good for tonight.

sites/skeleton.dev/src/lib/components/DocsIcon/icons.ts
```diff
@@ -3,8 +3,18 @@
     License - https://fontawesome.com/license (Commercial License)
     Copyright 2022 Fonticons, Inc
 */
+export type Icon = {
+    viewBox?: string;
+    path: string;
+};
+
+export type Icons = {
```
Mahmoud-zino, line 11: @jwatts777 please add `[key: string]: Icon;` here so we can index `Icons` using strings. this solves 2 of svelte-check errors.

sites/skeleton.dev/src/lib/themes.ts
```diff
@@ -1,7 +1,22 @@
 // Preset Themes
 
-const ghPath = 'https://github.com/skeletonlabs/skeleton/blob/master/src/themes';
-export const themes: any[] = [
+interface Font {
```
Mahmoud-zino, line 3: We are missing a `url:string` here, the url is used as href for the buttons in the theme.css.
we should also provide a url for each font in `themes`.
this solves another 2 of the svelte-check errors.

sites/skeleton.dev/src/routes/(inner)/elements/forms/+page.svelte
```diff
@@ -7,6 +7,18 @@
...
 
+	interface InputSettings {
+    title: string;
+    type: HTMLInputElement['type'];
+    placeholder?: string;
+    readonly?: boolean;
+    disabled?: boolean;
+    multiple?: boolean;
+    autocomplete?: string;
+    tabindex?: string;
```
Mahmoud-zino, line 18: this should be `tabindex?: number;` instead.
the same change must be applied in `/elements/forms/settings.ts`

sites/skeleton.dev/src/lib/themes.ts
jwatts777: Still waiting on more explanation.

sites/skeleton.dev/src/routes/(inner)/elements/forms/+page.svelte
jwatts777: If I implement this change, line 16 needs a change as well

sites/skeleton.dev/src/lib/themes.ts
jwatts777: Tentative change in place.

jwatts777 (commented):
Is there something here that I am missing?

endigo9740:
@jwatts777 FYI you can leave these unticked when creating the issue, then tap the generated UI checkboxes afterwards. They are interactive. Much simpler and the proper markdown formatting will be applied automatically.

image

AdrianGonz97 (changes requested):
Looks good! I've made a couple of changes to fix some types, specifically those that involve `sveld`, rearranged the import order of some components, fixed some typos (these were not introduced in this PR, but we may as well squash them when noticed) and fixed a small logic bug with the attribute mapper. 

I've also added suggestions below to tidy up the `Theme` type so that we're not redefining it's type in 2 separate locations.

Thanks!

sites/skeleton.dev/src/routes/(inner)/docs/get-started/SectionThemes.svelte
```diff
@@ -14,8 +13,23 @@ plugins: [
...
+	}
+
+	interface ActiveTheme {
+		file: string;
+		name: string;
+		fonts: Font[];
+		colors: string[];
+		surface: string;
+	}
+
```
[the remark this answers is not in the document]
jwatts777: Thank you for doing this. I had the same idea, but was nervous about being proactive in following DRY principles.

jwatts777:
> Looks good! I've made a couple of changes to fix some types, specifically those that involve `sveld`, rearranged the import order of some components, fixed some typos (these were not introduced in this PR, but we may as well squash them when noticed) and fixed a small logic bug with the attribute mapper.
> 
> I've also added suggestions below to tidy up the `Theme` type so that we're not redefining it's type in 2 separate locations.
> 
> Thanks!

Just for my own info, you returned a secondary wrapping of the return array with Object.entries vs Object.keys? 
Why do we do this instead of just returning back the object? We're grabbing, wrapping in an array, then returning a map of that array.

AdrianGonz97:
> Just for my own info, you returned a secondary wrapping of the return array with Object.entries vs Object.keys?
Why do we do this instead of just returning back the object? We're grabbing, wrapping in an array, then returning a map of that array.

@jwatts777 

`Object.keys` was only returning the `key` string and not the `value`, so if you look at the deployment before the fix [here](https://skeleton-docs-ayur8fcnj-skeleton-labs.vercel.app/elements/forms#input), and view the code for `Input`, you'll see that the first two characters of each key were being applied:
img

when it should be this:
img

AdrianGonz97 (approved):

PR #2177, opened 2023-10-24
Docs/ Nav List example for highlighting to match main Nav List example

ddxv:
## Description
I struggled a bit understanding the Skeleton.dev > Explore > Tailwind > Lists > Nav List for highlighting. As a newbie I was unfamiliar with the shorthand format it was written in, and after searching Discord for that the `{href}` was shorthand for how it would be written inside a loop. This makes sense, but was a bit unclear as someone newer.

So I suggest changing the highlighting example to match the main Nav List example just above the highlighting example. 

Dominik and Niktek helped with some pointers in the Discord channel, as well as suggesting changing the icons.

{description}

## Changsets

## Checklist

### DIFF
--- a/sites/skeleton.dev/src/routes/(inner)/elements/lists/+page.svelte
+++ b/sites/skeleton.dev/src/routes/(inner)/elements/lists/+page.svelte
@@ -183,7 +183,7 @@
 	<ul>
 		<li>
 			<a href="/elements/lists">
-				<span class="badge bg-primary-500">💀</span>
+				<span class="badge bg-primary-500">📋</span>
 				<span class="flex-auto">Skeleton</span>
 			</a>
 		</li>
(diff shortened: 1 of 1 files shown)


Sarenor (approved):

endigo9740:
@ddxv @Sarenor @niktek

While I agree shorthand might be confusing to new users, I don't think all the other changes are justified. Appending the attribute key should be all that's needed to clarify the point of confusion. Something like this:

```html
<CodeBlock language="html" code={`<a href={href} class="{classesActive(href)}">Page</a>`} />
```

Per the emojii, we've maintained the documentation for a bit over a year and half. In that time we've changed our standards a bit. The emojii was a remnant of an old standard. We now just use `(icon)` as a placeholder. Skeleton is icon library agnostic, so we don't want to take a opinionated stance here. I'd welcome the skulls being replaced with `(icon)`.

I know many libraries like to repeat the entirety of code per each snippet, but this is not how Skeleton handles this. We're a small team, so we keep things as lean as possible. I'd recommend reverting the wrapping nav/ul/li tags. This makes it harder to focus on what's being demonstrated here.

Once these changes are made, let me know and I'll be happy to merge.

ddxv:
>  Something like this:
> 
> <CodeBlock language="html" code={`<a href={href} class="{classesActive(href)}">Page</a>`} />

@endigo9740 Thanks for the input. I updated to follow that.

endigo9740:
That'll work, thanks for your contribution @ddxv

PR #2193, opened 2023-10-26
Feat/paginator-optional-size

### LINKED ISSUE #2066: Paginators: make 'size' setting optional (fro Server-Side Pagination)

In a server-side pagination context, we often don't know in advance the size of the dataset.
Therefore, I should be great if the 'size' setting could be optional. In a such case, the showNumerals option would be 'false', and the paginator woudl show, say "11-20" and not "11-20 of 243".

Furthermore, this would be consistent with https://vincjo.fr/datatables components (the server-side pagination part of the API is https://vincjo.fr/datatables/remote).

Enhancement

_No response_

Mahmoud-zino:
## Linked Issue

Closes #2066

## Description

when the size is optional, all controls depending on it are hidden.

## Changsets

## Checklist


endigo9740:
@Mahmoud-zino it looks like we hav a bug. In fact, I'm not sure how we'll overcome this. But when no `size` is present we don't know the max range, so you can just navigate to the next page forever. Well beyond the limit of the content itself.

Screenshot 2023-10-27 at 10 22 28 AM

I think we may just want to trust our gut on this one and make the recommendation that we don't support pagination without a size defined. If if the end users have to query this separately, they should be able to know the max number of items, which is crucial for pagination to work at all.

Mahmoud-zino:
@endigo9740 I noticed this when testing but I didn't consider it a bug because if the size is not known, it should be expected that we can't limit the paginator.
With that being said, I can see why this could cause us more trouble than benefits and I agree that we should not implement it. 😄

endigo9740:
@Mahmoud-zino yeah this will look and feel busted to end users. They won't understand why this occurring since there's no indicator they've reached the end of the content.

PR #2194, opened 2023-10-26
Bugfix/table cells interactive focusable

### LINKED ISSUE #2168: Table cells are always focusable
### Current Behavior

When using `<Table interactive={false} />`, the first cell of each row is focusable (tabindex=0). This makes keyboard navigation on a page cumbersome, as each table row takes focus when tabbing through a webpage, when no interaction is possible with the table rows.

### Expected Behavior

When `<Table interactive={false} />`, no cells should be focusable. When `interactive={true}`, the table cells should be focusable.

### Steps To Reproduce

1. Using the Table-component
2. Using some data in the `source` attribute to populate the table.
3. Press the `Tab button`
4. A cell in the table gets focus.

_No response_

### More Information

_No response_

Mahmoud-zino:
## Linked Issue

Closes #2168

## Description

table cells are focusable only in interactive mode.

## Changsets

## Checklist

### DIFF
--- a/.changeset/rich-squids-decide.md
+++ b/.changeset/rich-squids-decide.md
@@ -1,0 +1,5 @@
+---
+"@skeletonlabs/skeleton": patch
+---
+
+bugfix: table cells are not focusable when `interactive` is not set.
--- a/packages/skeleton/src/lib/components/Table/Table.svelte
+++ b/packages/skeleton/src/lib/components/Table/Table.svelte
@@ -68,7 +68,7 @@
 	<table
 		class="{classesTable}"
 		class:table-interactive={interactive}
-		role="grid"
+		role={interactive ? "grid" : "table"}
 		use:tableA11y
 	>
 		<!-- on:keydown={(e) => onTableKeydown(elemTable, e)} -->
@@ -76,7 +76,7 @@
 		<thead class="table-head {regionHead}">
 			<tr>
 				{#each source.head as heading }
-					<th class="{regionHeadCell}">{@html heading}</th>
+					<th class="{regionHeadCell}" role="columnheader">{@html heading}</th>
 				{/each}
 			</tr>
 		</thead>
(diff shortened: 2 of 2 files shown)


endigo9740:
@Mahmoud-zino I think you're on the right track, but we need to account for a11y here too. The WAI APG define two types of interactive table-like elements:

1. [Table Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/table/) - like a presentation HTML table, but not interactive
2. [Grid Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/grid/) - these are like a table, but interactive

While changing the `tabindex` is likely correct, it doesn't quite go far enough. I think we'll likely also need to modify following two roles:

- `role="grid"` -> `role="table"`
- `role="gridcell"` -> `role="cell"`

I'm also questioning if we should introduce `row` and `columnheader` roles, or if our use of a native HTML table is semantic enough as is.

I originally tested stuff like this using this Chrome Extension. Feel free to use it to double check:
https://chrome.google.com/webstore/detail/axe-devtools-web-accessib/lhdoppojpmngadmnindnejefpokejbdd

endigo9740:
@ryceg I know you're fairly knowledgable with the a11y stuff, so I'd welcome your input here as well

Mahmoud-zino:
@endigo9740 I think it is better to separate the components to grid / table and handling all ARIA rules for each one of them instead of trying to handle both in the same component

endigo9740:
@Mahmoud-zino if it's only the couple things mentioned I don't think separate components would be needed. Honestly most of the ARIA stuff overlaps as far as I can tell. Anything further abstraction would just make these harder to maintain over time.

endigo9740:
Hey @Mahmoud-zino, I'm doing PR reviews on Sundays now. Just checking in on this. As far as I can tell we should just need to modify the tools per the requirements I mentioned above. I'll aim to check in on this again Tuesday before the new release. Thanks!

Mahmoud-zino:
@endigo9740 I implemented the roles you mentioned and gone through the ARIA articles and implemented those that are not redundant roles.
Please feel free to suggest changes if needed.

PR #2208, opened 2023-11-01
feat/autocomplete-custom-filter-function

### LINKED ISSUE #1966: Allow custom search solution for autocomplete component
Is it ok to export the filter function so we can use our own filtering mechanism? I like to use a Fuzzy search with fuse.js

https://github.com/skeletonlabs/skeleton/blob/20af3ae37c6792cef598b404e0f513dd8885daaf/packages/skeleton/src/lib/components/Autocomplete/Autocomplete.svelte#L117

Mahmoud-zino:
## Linked Issue

Closes #1966

## Description

Added prop `filterFunction` with default value of the Skeleton's original filter function.

## Changsets

## Checklist

### DIFF
--- a/.changeset/plenty-drinks-tan.md
+++ b/.changeset/plenty-drinks-tan.md
@@ -1,0 +1,5 @@
+---
+"@skeletonlabs/skeleton": minor
+---
+
+feat: Autocomplete now accepts custom filter function using the prop `filterFunction`.
--- a/.changeset/plenty-drinks-tan.md
+++ b/.changeset/plenty-drinks-tan.md
@@ -2,4 +2,4 @@
 "@skeletonlabs/skeleton": minor
 ---
 
-feat: Autocomplete now accepts custom filter function using the prop `filterFunction`.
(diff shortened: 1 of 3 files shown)


endigo9740:
@Mahmoud-zino a couple things on this:

1. Go ahead and rename the prop `filter` (one word). We should aim to keep props as slim as possible.
2. I'm not sure I understand what the sentence in the example is referring to - highlighted here:

> This example uses a custom filter to filter options after value only.

Screenshot 2023-11-05 at 10 43 36 AM

The other examples just show the filtered results - maybe we just go ahead and follow that pattern here?

If you can make these couple changes then this should be good to merge.

Mahmoud-zino:
@endigo9740 
the line was intended to convey that the filter is different from the standard one. The standard filter applies after value, label, or keyword filtering. However, I recognize that it might be more confusing than helpful for users to understand its meaning, so I will remove it

endigo9740:
@Mahmoud-zino looks good. Be sure to flip to "ready to review" when you're done.

PR #2231, opened 2023-11-09
docs update on pagination  with dynamic size

### LINKED ISSUE #2211: Add information on how to have server side pagination with a dynamic size
### Link to the Page

https://www.skeleton.dev/components/paginators

I was having an issue while using the paginator, as it broke (numbers stopped updating correctly when changing page) when I added filters to my table (which in turn makes the size parameter dynamic, as it changes based on the select filters). 

I found no information in the documentation for this rather common pattern and I was able to come up with a solution of my own, with some help from other developers in the Discord server. However, I believe that, given how common this pattern is, it would make sense to add an example in the documents.

Here is my final solution:

```
paginationSettings = {
        page: Number($page.url.searchParams.get('pageNumber')),
        limit: 5,
        size: Number(data.Count), // This is the total count of the filtered data at any time and it is dynamic.
        amounts: []
    } satisfies PaginationSettings;

$: paginationSettings.size = Number(data.Count); // This ensures that the size property is updated whenever the load function runs (hence, when the data is filtered).
```

Riley1101:
## Add information on how to have server side pagination with a dynamic size

Closes #2211

## Description
Documentation on updating pagination size accordingly to the dynamic data source

## Changsets

## Checklist

### DIFF
--- a/sites/skeleton.dev/src/routes/(inner)/components/paginators/+page.svelte
+++ b/sites/skeleton.dev/src/routes/(inner)/components/paginators/+page.svelte
@@ -177,6 +177,23 @@ let tableHeaders: string[] = ['Positions', 'Name', 'Weight', 'Symbol'];
 <Table source={{ head: tableHeaders, body: paginatedSource }} />
 				`}
 			/>
+			<p>
+				If the paginated data source is dynamic, you might want to update the size parameter accordingly with the data. To achieve this, you can
+				use a Svelte reactive declaration to update it accordingly.
+			</p>
+			<CodeBlock
+				language="ts"
+				code={`
+let paginationSettings = {
+    page: 0,
+    limit: 5,
+    size: source.length,
+    amounts: [1, 2, 5, 10],
+} satisfies PaginationSettings;
+
+$: paginationSettings.size= source.length;
+				`}
+			/>
 		</section>
 		<section class="space-y-4">
 			<h2 class="h2">Server-Side Pagination</h2>
--- a/sites/skeleton.dev/src/routes/(inner)/components/paginators/+page.svelte
+++ b/sites/skeleton.dev/src/routes/(inner)/components/paginators/+page.svelte
@@ -194,6 +194,10 @@ let paginationSettings = {
 $: paginationSettings.size= source.length;
 				`}
 			/>
+            <p>
+                With this pagination setup, it's important to update the paginationSettings directly for reactivity. 
+                Simply updating a reference to source will not trigger the reactivity. 
+            </p>
 		</section>
 		<section class="space-y-4">
 			<h2 class="h2">Server-Side Pagination</h2>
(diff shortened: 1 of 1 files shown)


endigo9740:
Hey @Riley1101, per your PR - can you confirm you've tested this and ensure it works? My understanding and 
recommendation to users has always been to recreate the object if data within it changes.

Svelte details this in their tutorial:
https://learn.svelte.dev/tutorial/updating-arrays-and-objects

The general idea being if you just modify the values within the existing array/object, it wouldn't trigger trigger an updated render of that data. The Svelte specifically monitors only when there's a new instance of the array/object provided.

So I'd expect the correct way to update this would be more like:

```ts
let paginationSettings = {
    page: 0,
    limit: 5,
    size: source.length,
    amounts: [1, 2, 5, 10],
} satisfies PaginationSettings;

$: if (source.length)  paginationSettings = { ... paginationSettings, ... { size: source.length } };
```

I'd recommend testing this in a local project to verify which method works as expected here.

endigo9740:
@Riley1101 quick update - I opened this up to testing from other contributors on our Discord. If you're a member take a look at the `#contributor` channel. Another contributor, Hugos, mentioned your method for updating this data should work. But this seems to run contrary to what the Svelte tutorial explains.

We're going to reach out in the Svelte discord for clarification on this. But currently it look like your solution would be more efficient. We just just wanna vet for any edge cases here.

endigo9740:
@Riley1101 ok, per the Discord discussion we're good to move forward with your change here. Seems the Svelte doc is trying to point out you can't update a REFERENCE. But direct updates to the object values is fine for reactivity. That means your change here will be fine.

I will, however, make the suggesting that we append a small note to clarify this though. If it tripped us up, it'll likely trip up others.

Just append a small note to the description of this snippet that says something like: "make sure to update `paginatorSettings` directly, updating a reference will not trigger reactivity" or something along those lines. Maybe even provide an example mirroring the Svelte docs here to be explicitly clear.

Once that's in place I'm happy to merge. Thanks!

Riley1101:
> @Riley1101 ok, per the Discord discussion we're good to move forward with your change here. Seems the Svelte doc is trying to point out you can't update a REFERENCE. But direct updates to the object values is fine for reactivity. That means your change here will be fine.
> 
> I will, however, make the suggesting that we append a small note to clarify this though. If it tripped us up, it'll likely trip up others.
> 
> Just append a small note to the description of this snippet that says something like: "make sure to update `paginatorSettings` directly, updating a reference will not trigger reactivity" or something along those lines. Maybe even provide an example mirroring the Svelte docs here to be explicitly clear.
> 
> Once that's in place I'm happy to merge. Thanks!

 I will try to add a bit more concise clarification to this Thanks @endigo9740

endigo9740:
@Riley1101 thank you for the quick turn around. i went ahead and made a few modifications here as well, including making this it's own standalone section. That'll help us on the support end if we need to deep link to this topic.

Thanks for your contribution!

PR #2237, opened 2023-11-13
Bugfix/popup empty close query

### LINKED ISSUE #2218: Popup `closeQuery` prop does not work as expected
Currently the `closeQuery` prop doesn't work as outlined in the documentation. It states: `You may provide a custom query or set '' to disable this feature.`

However, setting closeQuery to `''` throws an error in the console.

image

We should probably adjust the `onWindowClick` event to account for this:

```typescript
function onWindowClick(event: any): void {
	// Return if the popup is not yet open
	if (popupState.open === false) return;
	// Return if click is the trigger element
	if (triggerNode.contains(event.target)) return;
	// If click it outside the popup
	if (elemPopup && elemPopup.contains(event.target) === false) {
		close();
		return;
	}
	// Handle Close Query State
	const closeQueryString: string = args.closeQuery === undefined ? 'a[href], button' : args.closeQuery;
	// Respect not querying for things if the string is empty
	if (closeQueryString === '') {
		return;
	}
	const closableMenuElements = elemPopup?.querySelectorAll(closeQueryString);
	closableMenuElements?.forEach((elem) => {
		if (elem.contains(event.target)) close();
	});
}
```

Also I would propose that we could potentially solve some issues with `data-popup` conflicts by being able to pass in a `bind:this` reference instead of only a `data-popup` value.

Example:

```svelte
<script lang="ts">
// Let the user maintain their own reference to the target
let popupEl: HTMLElement | null = null;

const popupFeatured: PopupSettings = {
	event: 'focus-click',
	// Matches the real element from bind:this
	target: popupEl,
	placement: 'bottom',
};
</script>

<button class="btn variant-filled" use:popup={popupFeatured}>Show Popup</button>

<div class="card p-4 w-72 shadow-xl" bind:this={popupEl}>
	<div><p>Demo Content</p></div>
</div>
```

Accomplishing this would likely be fairly straightforward as well by modifying the `setDomElements` method.

```typescript
function setDomElements(): void {
	// If a real element was passed, use it as the elemPopup
	if (args.target instanceof Element) {
		elemPopup = args.target;
	} else {
		elemPopup = document.querySelector(`[data-popup="${args.target}"]`) ?? document.createElement('div');
	}
	elemArrow = elemPopup.querySelector(`.arrow`) ?? document.createElement('div');
}
```

_Originally posted by @DerrikMilligan in https://github.com/skeletonlabs/skeleton/issues/1916#issuecomment-1794324793_

DerrikMilligan:
## Linked Issue

Closes #2218 

## Description

Adjusted the `onWindowClick` method in `popup` so that it matches the functionality outlined in the documentation regarding the `queryString` option.

Went for a guard statement since the rest of the method doesn't need to run if it's an empty string.

## Changsets

## Checklist

### DIFF
--- a/.changeset/eight-toes-itch.md
+++ b/.changeset/eight-toes-itch.md
@@ -1,0 +1,5 @@
+---
+"@skeletonlabs/skeleton": patch
+---
+
+Adjust popup 'onWindowClick' to respect an empty 'queryString'
--- a/packages/skeleton/src/lib/utilities/Popup/popup.ts
+++ b/packages/skeleton/src/lib/utilities/Popup/popup.ts
@@ -148,6 +148,8 @@ export function popup(triggerNode: HTMLElement, args: PopupSettings) {
 		}
 		// Handle Close Query State
 		const closeQueryString: string = args.closeQuery === undefined ? 'a[href], button' : args.closeQuery;
+		// Return if no closeQuery is provided
+		if (closeQueryString === '') return;
 		const closableMenuElements = elemPopup?.querySelectorAll(closeQueryString);
 		closableMenuElements?.forEach((elem) => {
 			if (elem.contains(event.target)) close();


Sarenor:
Looks good to me 👍  Thanks!

Issue #1870, opened 2023-08-15
Modal height (h-fit / h-full)


Mahmoud-zino:
### Describe the feature in detail (code, mocks, or screenshots encouraged)

As discussed in Discord we need a way to make the user decide if they want `h-fit` or `h-full`.

## Potential solutions
1. provide a `CssClass` prop `height` with a default value and document both possibilities with their side effects.
2. provide an enum sort of prop that lets the user choose one of both solutions.

### What type of pull request would this be?

Enhancement

### Provide relevant links or additional information.

Discord support issue -> https://discord.com/channels/1003691521280856084/1140610247304679454

last PR targeting this issue (changed from h-full to h-fit) -> #1695 

issue (caused by h-full) -> #1683

endigo9740:
@Mahmoud-zino FYI per this, I've noted on the v2 docs when a modal opens or closes, the some scrollbars appear very briefly on the right side of the page, which affects the position of the modal slightly. It animates in a C shape, rather than sliding up in a straight line.

This only last for a few milliseconds though, so it's a blink and you miss it kind of thing.

So far I've only tested in Chrome on Mac, but I think it may be related to the scroll changes for modals recently. Mentioning it here just in case.

emmarvpol:
> @Mahmoud-zino FYI per this, I've noted on the v2 docs when a modal opens or closes, the some scrollbars appear very briefly on the right side of the page, which affects the position of the modal slightly. It animates in a C shape, rather than sliding up in a straight line.
> 
> This only last for a few milliseconds though, so it's a blink and you miss it kind of thing.
> 
> So far I've only tested in Chrome on Mac, but I think it may be related to the scroll changes for modals recently. Mentioning it here just in case.

I noticed the same behavior so I added this to app.postcss. Now there is no scrollbar blinking and the modal is scrollable. Tested on Opera, Firefox and Safari.

```css
.modal-backdrop {
  @apply overflow-hidden;
}

.modal-transition {
  @apply max-h-screen;
}
```

endigo9740:
Per Discord - here's a way to set view-specific sizing options temporarily. This should provide a workaround for now.

`w-[50vw] h-[70vh]`

The `v` stands for "viewport" in this case. Check MDN if you need more instruction how these sizing units operate.

Schroedi:
[...]

PR #2241, opened 2023-11-15
Bugfix/modal slide right on open

### LINKED ISSUE #2239: Modal sliding to the right while opening
### Current Behavior

With a basic modal component in use I am seeing the modal slide to the right by about 5px while opening. 

### Expected Behavior

When the modal opens it should transition in, but not slide to the right after it displays. 

### Steps To Reproduce

1. Open the stackblitz below
2. Click the "Open Modal" button
3. Watch the modal open and slide to the right as it becomes visible

https://stackblitz.com/edit/skeletonlabs-repl-g8eubt

### More Information

_No response_

CoreyT355:
## Linked Issue

Closes #2239

## Description

Add base classes to modal to prevent scrollbar on open

The scrollbar appearing quickly on modal open is causing the modal slide. A proposed solution here, https://github.com/skeletonlabs/skeleton/issues/1870#issuecomment-1691777511, fixes this issue. 
- Added `overflow-hidden` to `cBackdrop`
- Added `max-h-screen` to `cTransitionLayer`

## Changsets

## Checklist

### DIFF
--- a/packages/skeleton/src/lib/utilities/Modal/Modal.svelte
+++ b/packages/skeleton/src/lib/utilities/Modal/Modal.svelte
@@ -101,8 +101,8 @@
 	export let transitionOutParams: TransitionParams<TransitionOut> = { duration: 150, opacity: 0, x: 0, y: 100 };
 
 	// Base Styles
-	const cBackdrop = 'fixed top-0 left-0 right-0 bottom-0 overflow-y-auto';
-	const cTransitionLayer = 'w-full h-fit min-h-full p-4 overflow-y-auto flex justify-center';
+	const cBackdrop = 'fixed top-0 left-0 right-0 bottom-0 overflow-hidden';
+	const cTransitionLayer = 'w-full h-fit min-h-full max-h-screen p-4 overflow-y-auto flex justify-center';
 	const cModal = 'block overflow-y-auto'; // max-h-full overflow-y-auto overflow-x-hidden
 	const cModalImage = 'w-full h-auto';
 
--- a/.changeset/hot-houses-joke.md
+++ b/.changeset/hot-houses-joke.md
@@ -1,0 +1,5 @@
+---
+"@skeletonlabs/skeleton": patch
+---
+
+bugfix: Fixes Issue #2239, added overflow-hidden to the base backdrop class list, and max-h-screen to the base transition class list.


Mahmoud-zino:
@CoreyT355 The problem with this solution is when a modal is bigger than the screen, the user won't be able to scroll through the content.
That is why we changed this scroll overflow to auto in the first place...

CoreyT355:
> @CoreyT355 The problem with this solution is when a modal is bigger than the screen, ~the user won't be able to scroll through the content~. I tested it again and the scroll is working fine, but when the modal is bigger than the screen, we can't see the top of the modal and we can't scroll up.

@Mahmoud-zino The only other thing I can think of is to instead apply `scrollbar-gutter: stable;` to the `modal-backdrop` class, or on that div itself. That seems to work, though the scrollbar still flashes for a second it does not effect the modal position. That's still not supported by Safari however.

Mahmoud-zino:
@CoreyT355 

> That's still not supported by Safari however

I think this is a deal breaker

I would suggest we apply the overflow after the intro ends and delete it before the outro starts. this way we can avoid the trouble we are having at the moment. I didn't done any testing yet, but feel free to try it out and see if it works.

CoreyT355:
> @CoreyT355
> 
> > That's still not supported by Safari however
> 
> I think this is a deal breaker
> 
> I would suggest we apply the overflow after the intro ends and delete it before the outro starts. this way we can avoid the trouble we are having at the moment. I didn't done any testing yet, but feel free to try it out and see if it works.

@Mahmoud-zino That's a solid approach, but it ends up flipping the issue. Now with a modal taller than the viewport, they shift to the left as the scrollbar appears as `overflow-y-auto` is reapplied at the end of the transition.

This is what I tried working with for example:
```
  function onModalTransitionInEnd(event: Event): void {
    if (!(event.target instanceof Element)) return;
    event.target.classList.add('overflow-y-auto');
  }

  function onModalTransitionOutStart(event: Event): void {
    if (!(event.target instanceof Element)) return;
    event.target.classList.remove('overflow-y-auto');
  }
  
  on:introend={onModalTransitionInEnd}
  on:outrostart={onModalTransitionOutStart}
```

I'll look at it some more tomorrow. I tried playing around with the different transition events. It always comes down to that scrollbar width showing up after the modal is visible. Maybe I am missing something though.

Mahmoud-zino:
> but it ends up flipping the issue. Now with a modal taller than the viewport, they shift to the left as the scrollbar appears as overflow-y-auto is reapplied at the end of the transition.

I missed this part, I think our only option here is to check the client size and act depending on it in relation to the screen size.
@endigo9740 I would appreciate your input once you have time 😄

endigo9740:
Hey guys, it's on my agenda to review - just short on time atm. Worst case I'll jump on this weekend, because I do very much want to see this fixed!

endigo9740:
@CoreyT355 @Mahmoud-zino I keeping forgetting how much of a rabbit hole this issue is. Essentially modals have a lot of different expectations:

* Most will just be small floating windows in the middle of the screen, mobile responsive styling can be minimal
* Some folks need really tall modals that are expected to scroll vertically with an overflow scroll
* While others are expected to fill the fill height/width of the screen without allowing overflow scroll

Personally I like @CoreyT355's suggestion here in theory - only adjusting the overflow styling during transition. But my concern would be that this wouldn't cover all three use cases stated above.

If you guys can do me a favor and try testing this or any proposed solution against these three use cases. I think that's what will help determine what we move forward with.

Then long term, I do have it on my agenda to revisit the modal feature in depth, which will likely include a rewrite of the styles top to bottom. There's some new goodies coming as part of Svelte 5 that should reduce some of the required code here. And in doing that might make it easier to focus on the styling.

Hope this helps.

Mahmoud-zino:
@endigo9740 I think getting the client size will solve a lot of out problems, I am struggling to find the time to test it but I will try to do it tomorrow.

packages/skeleton/src/lib/utilities/Modal/Modal.svelte
```diff
@@ -101,8 +101,8 @@
 	export let transitionOutParams: TransitionParams<TransitionOut> = { duration: 150, opacity: 0, x: 0, y: 100 };
 
 	// Base Styles
-	const cBackdrop = 'fixed top-0 left-0 right-0 bottom-0 overflow-y-auto';
-	const cTransitionLayer = 'w-full h-fit min-h-full p-4 overflow-y-auto flex justify-center';
+	const cBackdrop = 'fixed top-0 left-0 right-0 bottom-0 overflow-hidden';
```
endigo9740, line 104: @CoreyT355 @Mahmoud-zino after reviewing Mahmoud's PR here:
https://github.com/skeletonlabs/skeleton/pull/2255

I've noted you're both modifying the `overflow` property here. Given the only other change is the `max-h-screen` setting below. I'd recommend that if we feel this is useful, to replicate it in @Mahmoud-zino PR and we'll favor that and close this PR.

Let me know if that works for both of you. Thanks!

CoreyT355: @endigo9740 Definitely better to unify the changes down to a single PR.

PR #2255, opened 2023-11-20
Bugfix: modal wobble on transition in/out

### LINKED ISSUE #1870: Modal height (h-fit / h-full)

### LINKED ISSUE #2239: Modal sliding to the right while opening

Mahmoud-zino:
## Linked Issue

Closes #2239
Closes #1870

## Description

Fixed the issue by checking the client height and comparing it to the window height and then adding the overflow based on the result.

> Note: `cOverflow` is added directly to the classes of the `backdrop` because otherwise the reactivity won't be triggered on every change.

## Changsets

## Checklist


Mahmoud-zino:
@endigo9740 this won't solve #1870, but I though separating them will simplify the review.
If this solution is accepted we can close #2241 . Thank you @CoreyT355 for helping with this bug.

CoreyT355:
> @endigo9740 this won't solve #1870, but I though separating them will simplify the review. If this solution is accepted we can close #2241 . Thank you @CoreyT355 for helping with this bug.

@Mahmoud-zino That's a nice solution. I haven't worked with the `svelte:window` piece yet, so I didn't think about it when the idea of catching the window height came up. 
Thanks for finding a solution here. Hopefully this gets merged in.

endigo9740:
@CoreyT355 @Mahmoud-zino I've been handling PR reviews on Sunday's (tomorrow for me) so expect a proper audit then.

packages/skeleton/src/lib/utilities/Modal/Modal.svelte
```diff
@@ -115,6 +115,9 @@
 	};
 	let currentComponent: ModalComponent | undefined;
 	let registeredInteractionWithBackdrop = false;
+	let modalElement: HTMLDivElement;
+	let windowHeight: number;
+	let cBackdropOverflow: CssClasses = 'overflow-y-hidden';
```
endigo9740, line 120: The `cX` classes are meant to be a fixed set of base classes that do not change. To avoid confusion, I'd rename and drop the "c". Likewise remove the CssClasses type designation - those are only needed for exposed props that users can override. It's an identifier that enables Tailwind Intellesense for those props. Given it's a privative, you can leave off the type entirely in fact!

packages/skeleton/src/lib/utilities/Modal/Modal.svelte
```diff
@@ -131,6 +134,19 @@
...
+		if (!modalHeight) return;
+
+		if (modalHeight > windowHeight) {
+			cBackdropOverflow = 'overflow-y-auto';
+		} else {
+			cBackdropOverflow = 'overflow-y-hidden';
+		}
+	}
+	// first child of the modal is the content.
+	$: onModalHeightChange((modalElement?.firstChild as HTMLElement)?.clientHeight);
```
endigo9740, line 148: I'm not sure `firstChild` is reliable here. While we suggest users created their custom modal components with a single wrapping element, there's no guarantee they will. Any reason not to just verify the size of the modal element itself? It should always be present and always wrap the entirety of the content.

If you're concerned about padding and borders - then I'd recommend the use of `offsetHeight`:
https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetHeight

endigo9740:
@Mahmoud-zino just a couple small nitpicks here, but overall I like this. It seems to be a great solution to the issue at hand.

packages/skeleton/src/lib/utilities/Modal/Modal.svelte
```diff
@@ -101,7 +101,7 @@
 	export let transitionOutParams: TransitionParams<TransitionOut> = { duration: 150, opacity: 0, x: 0, y: 100 };
 
 	// Base Styles
-	const cBackdrop = 'fixed top-0 left-0 right-0 bottom-0 overflow-y-auto';
+	const cBackdrop = 'fixed top-0 left-0 right-0 bottom-0';
 	const cTransitionLayer = 'w-full h-fit min-h-full p-4 overflow-y-auto flex justify-center';
```
endigo9740, line 105: Per my comment in #2241, I'd recommend appending testing with `max-h-screen` applied to the `cTransitionLayer` classes. Though off hand it seems redundant with `w-full`. I'm sure there's a good reason for this that's not obvious to me though.

packages/skeleton/src/lib/utilities/Modal/Modal.svelte
Mahmoud-zino: @endigo9740 
while you are correct about the `firstChild` not being reliable, the modal element itself is also not reliable, when using custom modals the modal-element has the height of (0) always. so I have changed the code to check both the parent and the child before deciding which one represents the size. this should work with every situation.

packages/skeleton/src/lib/utilities/Modal/Modal.svelte
Mahmoud-zino: @endigo9740 I also don't know why it is important  😄, but I have added it 👍

Mahmoud-zino:
@endigo9740 I just checked it again, this is the effect of having `max-h-screen`, deleting it solved the issue.

endigo9740:
@Mahmoud-zino That's awesome. In order to properly test this, can you do me a favor and implement a new component modal example that is full screen? Just follow the lead of the current examples. Without that we don't have a great way to test all three conditions I mentioned above. The contents of the modal can be anything - even just placeholder text that says: "Full Screen Example". I'll polish it a bit more later.

endigo9740:
@Mahmoud-zino I've moved, renamed, and updated the Modal docs to showcase the new "full" modal option. I've also update the contents of the modal to provide some simpler styles for showcasing how this can work.

Note there is currently no way to disable the backdrop > modal-transition built in padding. But I think that's an issue we should address in Skeleton v3.

I also see how you're handling modals that are oversized vertically. The scrollbar won't appear as you resize your window, but will appear if that size is too large when the modal appears. I think that's a happy medium for now. Again, I'd like to tackle the root issues for this in v3.

Assuming you're happy with my changes, feel free to approve this PR and I'll aim to merge right away. Thanks for the hard work on this one!

endigo9740 (approved):

Mahmoud-zino:
@endigo9740 I can't approve my own PR, but it looks good to me :+1:

PR #2270, opened 2023-11-29
Bugfix/set accordion item id when passed

### LINKED ISSUE #2265: Problem with setting id for AccordionItem
### Current Behavior

When trying to set an id for an AccordionItem explicitly to "X", the id is actually set to "accordion-control-X".

### Expected Behavior

If overriding the default the expectation is that the id is set to "X".

### Steps To Reproduce

_No response_

_No response_

### More Information

_No response_

phamduylong:
Add id checking for accordion item, it it was not passed in by user then we generate a random id

## Linked Issue

Closes #2265

## Description

Add a layer to check if id was manually passed in by user. If yes, use the id assigned, else generate a random one (the OG way)

Add id testing. Since the id will be assigned inside onMount calls, a slight delay is needed.

## Changsets

## Checklist


packages/skeleton/src/lib/components/Accordion/AccordionItem.svelte
```diff
@@ -33,7 +32,9 @@
 	 * Provide a unique input id. Auto-generated by default
 	 * @type {string}
 	 */
-	export let id = String(Math.random());
+	export let id = '';
```
Mahmoud-zino, line 35: @phamduylong instead of doing all these steps you implemented, we can achieve the same effect if you move the string `accordion-control-` to the default value.

so the only change would be:
```suggestion
	export let id = `accordion-control-${Math.random()}`;
```

Mahmoud-zino (changes requested):

packages/skeleton/src/lib/components/Accordion/AccordionItem.svelte
phamduylong: Dum dum me couldn't figure it out. Thanks, I'll implement this. Way cleaner.

Mahmoud-zino: I think we have to rethink the solution, we also have another Id like `accordion-panel-{id}` which should also be correctly replaced

phamduylong:
> @phamduylong I just deleted my last review, because it was not complete... so moving the string `accordion-control-` to the default value of the id itself will still work because the the panel id will be `accordion-panel-accordion-control-xxxx`. However this is still not clean and will be a pain for users trying to target the panel element for CSS-styles or similar stuff.

A solution I can think of to help with styling is to have props for both the control and the panel (like we have for tables). This should ease the pain of custom styling. But I agree that the panel could use an id as well.

endigo9740:
@phamduylong @Mahmoud-zino I think there may have been a failure of communication here guys. The original issue was that when we forward the `on:toggle` event the ID has `accordion-controls-${id}`, where as we just need the raw `id` value.

This change should be as simple as:

```ts
export let id = String(Math.random());

// ...

dispatch('toggle', { event, id: `accordion-control-${id}`, open: currentOpenState, autocollapse });
```

To:

```ts
export let id = `accordion-control-${String(Math.random())}`;

dispatch('toggle', { event, id, open: currentOpenState, autocollapse });
```

Where the `id` is passed verbatim without anything else appended.

You'll have to explain the purpose of all other changes - because they seem to be out of scope. And unfortunately we don't allow out of scope changes to be shoehorned in without prior discussion and approval.

Issue #1672, opened 2023-06-18
Vertical Radio button group


TSRChapman:
### Current Behavior

verticle radio group is slightly out of alignment from the top radio to the second radio

### Expected Behavior

verticle alignment on all radio buttons

### Steps To Reproduce

visible on the documentation page for radio button groups showing the vertical arrangement

### Link to Reproduction / Stackblitz

https://www.skeleton.dev/components/radio-groups

endigo9740:
Appending a quick screenshot to show the error - note the second item is a bit off when the hover state is triggered:

![Screenshot 2023-06-19 at 12 02 40 PM](https://github.com/skeletonlabs/skeleton/assets/1509726/1846d55b-9853-4745-ba0b-9869df1f2961)

The issue seems to be the spacing. Even in the vertical orientation it's still using `space-x`

PR #2309, opened 2023-12-13
Fix RadioGroup spacing

### LINKED ISSUE #2295: Spacing prop is ignored by RadioGroup
### Current Behavior

The RadioGroup component accepts a "spacing" prop that is supposed to override the default spacing. However the spacing relevant code in RadioGroup is as follows:
```
<script>
export let spacing = "";
//...
$:
  spacing = `${display.includes("flex-col") ? "" : "space-x-1"}`;
$:
  classesBase = `${cBase} ${display} ${background} ${border} ${spacing} ${rounded} ${$$props.class ?? ""}`;
</script>

<div class="radio-group {classesBase}" data-testid="radio-group" role="radiogroup" aria-labelledby={labelledby}>
    <slot />
</div>
```

This prevents the spacing prop being considered as the reactive line will always default spacing to "" or "space-x-1"

### Expected Behavior

Spacing prop should effect the css classes.

### Steps To Reproduce

Use spacing on RadioGroup component

https://stackblitz.com/edit/jq-starter-zrlfgn?file=my-skeleton-app%2Fsrc%2Froutes%2F%2Bpage.svelte

### More Information

_No response_

phamduylong:
## Linked Issue

Closes #2295

## Description

Made spacing props to adopt value passed in and set *space-x-1* as default value for it

## Changsets

## Checklist

### DIFF
--- a/packages/skeleton/src/lib/components/Radio/RadioGroup.svelte
+++ b/packages/skeleton/src/lib/components/Radio/RadioGroup.svelte
@@ -12,7 +12,7 @@
 	/** Provide classes to set the border styles. */
 	export let border: CssClasses = 'border-token border-surface-400-500-token';
 	/** Provide classes horizontal spacing between items. */
-	export let spacing: CssClasses = '';
+	export let spacing: CssClasses = 'space-x-1';
 	/** Provide classes to set the border radius. */
 	export let rounded: CssClasses = 'rounded-token';
 
@@ -47,8 +47,6 @@
 	const cBase = 'p-1';
 
 	// Reactive
-	// display space-x only on row flex as it is not needed on column flex.
-	$: spacing = `${display.includes('flex-col') ? '' : 'space-x-1'}`;
 	$: classesBase = `${cBase} ${display} ${background} ${border} ${spacing} ${rounded} ${$$props.class ?? ''}`;
 </script>
 
--- a/packages/skeleton/src/lib/components/Radio/RadioGroup.svelte
+++ b/packages/skeleton/src/lib/components/Radio/RadioGroup.svelte
@@ -5,14 +5,16 @@
 	import type { CssClasses } from '../../index.js';
 
 	// Props (Group)
-	/** Provide display classes. Set `flex` to stretch full width. */
+	/** Provide display classes. Use `flex` to stretch full width. */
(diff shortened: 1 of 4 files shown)


packages/skeleton/src/lib/components/Radio/RadioGroup.svelte
```diff
@@ -47,8 +47,6 @@
 	const cBase = 'p-1';
 
 	// Reactive
-	// display space-x only on row flex as it is not needed on column flex.
-	$: spacing = `${display.includes('flex-col') ? '' : 'space-x-1'}`;
```
Mahmoud-zino, line 51: @phamduylong we still need this because of this bug: #1672
I would suggest the following:
```ts
export let spacing: CssClasses = 'space-x-1';
//....
// display space-x only on row flex as it is not needed on column flex.
$: classesSpacing = `${display.includes('flex-col') ? '' : spacing}`;
$: classesBase = `${cBase} ${display} ${background} ${border} ${classesSpacing} ${rounded} ${$$props.class ?? ''}`;
```

Mahmoud-zino (changes requested):

endigo9740:
@phamduylong @Mahmoud-zino so we had two issues here, neither of which would be solved by adding a `spacing` prop:

1. `display` was not the appropriate prop to add `flex-col` for vertical layouts.
2. `gap` is the appropriate style/prop to handle both vertical and horizontal gap spacing.

To correct this I've made a few changes:

1. `display` is now properly labeled to handle `display: x` CSS styles
2. A new `flexDirection` prop has been added to handle `flex-col` classes
3. A new `gap` prop has been added to handle universal item spacing (vertical/horizontal)
4. The docs have been updated to denote all changes.

This should provide a much more better solution to the original issue, while fixing the misuse of the `display` property.

Mahmoud-zino:
@endigo9740 your solution with the ` gap` prop  is great! but are we ok with having duplicated ` inline-flex`  classes? 
it won't have an effect but it doesn't look clean!

endigo9740:
This `flexDirection` prop had an incorrect default value. It should now be set to `flex-row` on the `dev` branch. Thanks for the catch @Mahmoud-zino

PR #2317, opened 2023-12-15
feat/progressbar-animation-prop

### LINKED ISSUE #2281: Add ProgressBar props to control indeterminate animation settings

We have had a Discord user request a way to customize the `indeterminate` mode animation transition speed. We should strive to minimize the `<style>` block animation settings:

https://github.com/skeletonlabs/skeleton/blob/master/packages/skeleton/src/lib/components/ProgressBar/ProgressBar.svelte#L64

The goal is to split the shorthand:

```css
animation: animIndeterminate 2s infinite linear;
```

Into dedicated properties like so:

```css
animation-name: animIndeterminate;
animation-iteration-count: infinite;
```

Then implement new component props that handle styles for [transition duration](https://tailwindcss.com/docs/transition-duration) and [transition timing functions](https://tailwindcss.com/docs/transition-timing-function). Suggested prop names would be:

* `indeterminateDuration="duration-75"`
* `indeterminateTransitionDuration="linear"`

Enhancement

Discord thread:

https://discord.com/channels/1003691521280856084/1181418447696048218/1181418447696048218

Mahmoud-zino:
## Linked Issue

Closes #2281

## Description

Added Prop `indeterminateAnimation` which can be used to supply a custom indeterminate animation.

## Note
@endigo9740 
per your comment:
> @Mahmoud-zino I see what you mean, if there's not equivalent ways to handle these animation properties in a piecemeal fashion, then allowing folks to replace the entire animation may be the best route here. I'd suggest researching the former, and if not possible we go with the latter.

- the first option is not possible (see last option for something similar ^^).
- the second option is possible but requires users to define their animations in the global style-sheet, not great but sadly I can't think of anything else that could be easy to implement without introducing another dependency, so I implemented it but feel free to close the PR if this is something we don't want to do.
- A third option would be to define custom utilities for tailwind to target the animations. It would be nice but it feels like creating (again the next option :smile: ) from scratch...
- The last option I found was to use plugins, I didn't use any until now but at a quick glance [tailwindcss-animated](https://github.com/new-data-services/tailwindcss-animated) seems like a good addition.

## Changsets

## Checklist

### DIFF
--- a/.changeset/mighty-cycles-drive.md
+++ b/.changeset/mighty-cycles-drive.md
@@ -1,5 +1,5 @@
 ---
-"@skeletonlabs/skeleton": major
+"@skeletonlabs/skeleton": minor
 ---
 
 feat: Added the prop `indeterminateAnimation` to customize the ProgressBar indeterminate animation.
--- a/.changeset/mighty-cycles-drive.md
+++ b/.changeset/mighty-cycles-drive.md
@@ -1,0 +1,5 @@
+---
+"@skeletonlabs/skeleton": major
+---
+
+feat: Added the prop `indeterminateAnimation` to customize the ProgressBar indeterminate animation.
--- a/packages/skeleton/src/lib/components/ProgressBar/ProgressBar.svelte
+++ b/packages/skeleton/src/lib/components/ProgressBar/ProgressBar.svelte
@@ -18,6 +18,8 @@
 	export let rounded: CssClasses = 'rounded-token';
 	/** Provide classes to set the meter transition styles. */
 	export let transition: CssClasses = 'transition-[width]';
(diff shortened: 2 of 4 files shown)


endigo9740:
@Mahmoud-zino of those options, the second seems like the best. It's exactly what I had in mind. Define a reusable animation in your global stylesheet, then make sure to pass/use that for each progress bar instance. Global styles have their place, and that makes sense. It should be relatively easy to understand, implement, and reuse with no performance impacts.

endigo9740:
@Mahmoud-zino the functional implementation was spot on, I just updated the docs and prop name to simplify things a bit. Great job here!

PR #2334, opened 2023-12-20
Bugfix/modal prompt inconsistent type

### LINKED ISSUE #2283: Modal prompt type returning non-string value if default value is used
### Current Behavior

Creating a modal of type prompt might return a non-string value if user proceeds with the default value. This example below will return 12 as a number type when user doesn't modify the value, but "13" right as we click the increment button on input.

```ts
const modal: ModalSettings = {
  type: 'prompt',
  title: 'Enter Age',
  body: 'How old are you?',
  value: 12,
  valueAttr: { type: 'number', required: true },
  response: (r) => console.log('response:', r)
};
```
As a user, I'd expect to put value (the placeholder) same type with what valueAttr type is, hence it returned a number here. Maybe a note saying default value has to be of type string can help.

### Expected Behavior

Unsure, perhaps it should be string as well for default values? Would be nice to keep it consistent.

### Steps To Reproduce

_No response_

_No response_

### More Information

Found while documenting [#2127](https://github.com/skeletonlabs/skeleton/issues/2127)

Mahmoud-zino:
## Linked Issue

Closes #2283

## Description

when using "any" as the type of the variable we return number if the default value is not changed, but when it is changed the variable will have the type string (changed by the input) which is causing the inconsistency.
to solve this I converted any input to string (with accounting for null an undefined).

----

After further discussion, we decided to cast only numbers when the prompt modal of type number.
Discussion: https://discord.com/channels/1003691521280856084/1003699523522142399/1187099645684498512

## Changsets

## Checklist

### DIFF
--- a/packages/skeleton/src/lib/utilities/Modal/Modal.svelte
+++ b/packages/skeleton/src/lib/utilities/Modal/Modal.svelte
@@ -125,7 +125,7 @@
 	modalStore.subscribe((modals: ModalSettings[]) => {
 		if (!modals.length) return;
 		// Set Prompt input value and type
-		if (modals[0].type === 'prompt') promptValue = modals[0].value?.toString();
+		if (modals[0].type === 'prompt') promptValue = modals[0].value;
 		// Override button text per instance, if available
 		buttonTextCancel = modals[0].buttonTextCancel || buttonTextDefaults.buttonTextCancel;
 		buttonTextConfirm = modals[0].buttonTextConfirm || buttonTextDefaults.buttonTextConfirm;
@@ -183,7 +183,11 @@
 
 	function onPromptSubmit(event: SvelteEvent<SubmitEvent, HTMLFormElement>): void {
 		event.preventDefault();
-		if ($modalStore[0].response) $modalStore[0].response(promptValue);
+		if ($modalStore[0].response) {
+			if ($modalStore[0].valueAttr !== undefined && 'type' in $modalStore[0].valueAttr && $modalStore[0].valueAttr.type === 'number')
+				$modalStore[0].response(parseInt(promptValue));
+			else $modalStore[0].response(promptValue);
+		}
 		modalStore.close();
 	}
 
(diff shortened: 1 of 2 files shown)


packages/skeleton/src/lib/utilities/Modal/Modal.svelte
```diff
@@ -183,7 +183,11 @@
 
 	function onPromptSubmit(event: SvelteEvent<SubmitEvent, HTMLFormElement>): void {
 		event.preventDefault();
-		if ($modalStore[0].response) $modalStore[0].response(promptValue);
+		if ($modalStore[0].response) {
+			if ($modalStore[0].valueAttr !== undefined && 'type' in $modalStore[0].valueAttr && $modalStore[0].valueAttr.type === 'number')
+				$modalStore[0].response(parseInt(promptValue));
```
phamduylong, line 188: Why not Number constructor 🤔? I generally think it's better, not that it would make a huge performance leap in this context.
```ts
$modalStore[0].response(Number(promptValue));
```

Feel free to correct me if I'm wrong.

Mahmoud-zino: @phamduylong It is not better (performance depends on the string we are trying to parse), it just behaves differently when parsing, and to be honest both behaviors are totally Ok in our case, so I didn't think about it and would leave it as is if there is not a good argument behind the change.
but thanks for the suggestion and feel free to follow up if you have a better reason for the change :smile:

packages/skeleton/src/lib/utilities/Modal/Modal.svelte
```diff
@@ -183,7 +183,11 @@
 
 	function onPromptSubmit(event: SvelteEvent<SubmitEvent, HTMLFormElement>): void {
 		event.preventDefault();
-		if ($modalStore[0].response) $modalStore[0].response(promptValue);
+		if ($modalStore[0].response) {
```
endigo9740, line 186: @Mahmoud-zino add a comment just above this line explaining the reason for this change. Something like:

> Ensures input[number] retains numeric type when up/down arrows are pressed in browser.

Once that's add this is good to merge.

endigo9740 (changes requested):

endigo9740 (approved):

PR #2414, opened 2024-01-14
Bugfix/center

seltzerfish:
## Linked Issue
https://github.com/skeletonlabs/create-skeleton-app/issues/41

## Description

Applies the tailwind class h-full to html and body elements in the barebones template so that the content is centered.

## Checklist

- [x ] This PR targets the `dev` branch (NEVER `master`)
- [x ] Documentation reflects all relevant changes
- [x ] Branch is prefixed with: `docs/`, `feat/`, `chore/`, `bugfix/`
- [x ] Ensure Svelte and Typescript linting is current - run `pnpm ci:check`
- [x ] Ensure Prettier linting is current - run `pnpm format`
- [x ] All test cases are passing - run `pnpm test`
- [x ] Includes a changeset (if relevant; see above)

### DIFF
--- a/.changeset/rotten-stingrays-deny.md
+++ b/.changeset/rotten-stingrays-deny.md
@@ -1,0 +1,5 @@
+---
+"create-skeleton-app": patch
+---
+
+Vertically center content for barebones template.
--- a/.changeset/rotten-stingrays-deny.md
+++ b/.changeset/rotten-stingrays-deny.md
@@ -2,4 +2,4 @@
 "create-skeleton-app": patch
 ---
 
-Vertically center content for barebones template.
+bugfix:Vertically center content for barebones template.
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -8,7 +8,7 @@ Closes #{issueNumber}
 
 ## Changsets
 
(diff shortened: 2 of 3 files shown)


endigo9740:
Hey @seltzerfish, we do refer to it as a "barebones" template because it has the minimal changes needed to run. This might fall into the category of being too opinionated.

Also we typically strongly discourage multiple changes in a single PR, so the change to the PR template is both out of scope of this change, and also again something we would highly discourage. While `cs` is shorter, it's less descriptive of the command being run. We can't assume the majority of the audience knows what that is or does.

If you are interested in submitting more changes, I might recommend consulting with me or other members of the team beforehand going forward. Please and thanks.

seltzerfish:
Gotcha. The barebones template used to center this, so i assumed it was intended at some point. If this is working as intended now that's fine though :P

As for the `cs`, I added that because that's what the script is actually called in `package.json`, and I interpreted the instructions to run `pnpm changeset` as outdated/incorrect because that command won't actually do anything.

endigo9740:
@seltzerfish I can confirm the `pnpm changeset` command runs just fine. I use it often. You just have to make sure you run it in the root of the project. Never within a package or site project.

Issue #2360, opened 2023-12-26
NEXT: Tailwind Plugin


endigo9740:
This will act as a hub to centralize for this information.

## See Also

The following updates will likely need to be handled in coordination.

* https://github.com/skeletonlabs/skeleton/issues/1142
* https://github.com/skeletonlabs/skeleton/issues/2359

## Tailwind Directive Layers

- `base` - global styles, resets, etc
- `components` - for opt-in element styles, like `.btn`
- `utilities` - general purpose utility classes
- `variants` - to append new variants, like `hover:` and `md:`

## Maintainer Requests

The following requests are coming straight from the Skeleton team. These are highly likely be implemented:

- [x] Implement skeleton utility classes via `theme.extend` over standalone classes
- [x] Explore using `@apply` directly in the tailwind plugin's CSS-in-JS objects ([reference](https://twitter.com/adamwathan/status/1742248960843227567))
- [x] Reorganize current styles; distribute to Tailwind directive layers
- [x] Updated Tailwind > Blocks to something more akin to the [OpenAI cookbook](https://cookbook.openai.com/)
- [x] Update Chrome scrollbar styles to [new spec](https://developer.chrome.com/docs/css-ui/scrollbar-styling)
- [x] Update theme type definitions to support all changes (removing `properties_dark`, etc)
- [x] A fix for the OS dark mode setting. [See the issue outlined here](https://github.com/skeletonlabs/skeleton/issues/2359#issuecomment-1912633056).
- [x] Implement @AdrianGonz97 suggestion for a [simpler theme import method](https://github.com/skeletonlabs/skeleton/issues/2360#issuecomment-1894410516)
- [ ] @AdrianGonz97 please see @niktek [question regarding exec scripts here](https://github.com/skeletonlabs/skeleton/pull/2431#discussion_r1468276610)
- [x] ~~Aim to make `config.content` simpler for core/svelte/react ([reference](https://tailwindcss.com/docs/plugins#extending-the-configuration))~~ (wait until TW v4)

Design Tokens:

- [x] Revisit the `token` prefix/suffix concept after implementing with `theme.extend`
- [x] Design token background styles should follow [Material Design's naming scheme](https://m3.material.io/components/all-buttons) (ex: Ghost -> Tonal)
- [x] The concept of "Variants" should be revisited; consider folding these into design tokens
- [x] We need to make the concept of "color pairings" more clear; lots of confusion around these

Tailwind Components (aka "Elements")

[...]

Issue #2359, opened 2023-12-26
NEXT: Improved dark mode support


endigo9740:
> [!WARNING] 
> This issue is a work in progress.

This will act as a hub to centralize for this information.

## Maintainer Requests

The following requests are coming straight from the Skeleton team. These are highly likely be implemented:

* PROPOSAL: https://github.com/skeletonlabs/skeleton/discussions/951
* Consider doing away with the canned Lightswitch component in favor of better utility features
* (expect this to expand over time)

## Community Requests

The following requests have come from the community and are under consideration:

* ...

## Bugs and Issues

* https://github.com/skeletonlabs/skeleton/issues/1657

## Feedback

If you have additional updates or requests for this feature, please do so in the comments section below.

endigo9740:
@AdrianGonz97 I shared this on Discord, but to echo here for future reference:

---

@cokakoala so I've finally had a chance to sit down and look through the CSS compile scripts and I think I've found where our OS strategy dark mode bug is coming from. It seems you use a local Tailwind Config to determine the way the styles are processed:

https://github.com/skeletonlabs/skeleton/blob/feat/next-tailwind-plugin/packages/skeleton/scripts/compile-css-to-js.ts#L32

If I update both the doc site TW config AND this script TW config to remove the `darkMode: class` option, then all styles suddenly start working with the OS toggle. BUT, the two configs have to be in sync. If I try OS+Class or Class+OS things are out of sync and we get the weird style issues. Do we maybe need to pass that setting down to the plugin in some fashion?

Here's what class-based generates:
![image](https://github.com/skeletonlabs/skeleton/assets/1509726/1d5164cf-fad1-4fd0-953b-69fc45d34fc2)

Here's what OS-based generates:
![image](https://github.com/skeletonlabs/skeleton/assets/1509726/ce9b72be-93be-44ae-8370-7432218b2e6e)

So ideally we need some way for our plugin to capture the same `darkMode` setting the users are using in their local project. It's the only way to ensure this matches up. Zero clue how we handle this though!

endigo9740:
With #2431 we can now consider this task done as well. Closing this out!

Per the Lightswitch feature - this will likely take the form of a guide rather than a single component, so I'm going to move that over to the Documentation ticket for now. The instructions for setting this up may differ framework to framework (especially meta frameworks like Astro) so we'll handle this in a dedicated section.

PR #2431, opened 2024-01-17
NEXT Tailwind Plugin Refactor

endigo9740:
## Linked Issue

Closes https://github.com/skeletonlabs/skeleton/issues/2360

## Description

Refactor the theme format and Tailwind plugin.


endigo9740:
@niktek @AdrianGonz97 this PR is now ready to merge. This is a huge first step towards v3 and will help unblock everything that comes next. So I cannot under emphasize the importance of getting this in. That said, I'd welcome a review by you two, but otherwise I'd say we go ahead and merge and follow up with any other additional changes or improvements over time, in smaller PRs.

Note that once this is merged, any v2 `/site` apps that relied on `tw-plugin` will no longer be in a functional state on the `next` branch. This is becausethe theme format, Tailwind components, etc have changed dramatically. So I'll leave it up to your discretion how we move forward with that.

Currently we'll only have ONE functional theme, called `/themes/next.ts`. But once the generator is operational, we'll expand with additional options and can look into features such as theme switching for the Astro docs.

@niktek given the CI/CD errors on this page are Vercel-related, let me know if you need anything from me to help troubleshoot. But this SHOULD be deployed to the CF instance and should be operational here:
https://next.skeleton.dev/

Note I'll be away for part of the afternoon, but back on later tonight or this week.

Thanks guys!

packages/skeleton/src/plugin/components/legacy/horizontal-rules.css
niktek, line 1: Quick query as to how these legacy files are being used - are they just for reference ? if so, why are there new additions to them ?

endigo9740: Some of these are going to be moved to Resources > Cookbook. But a lot of them can be removed eventually.

packages/skeleton/src/plugin/extends/spacing.ts
niktek, line 1: This feels like a layer of abstraction on top of TW and that users should just be extending the TW config directly to adjust these values.  Is this because we need the var values to generate other Skeleton classes and can't get access to these values ?

niktek (commented):
Only some superficial comments to worry about.  Will get this deployed soon

packages/skeleton/src/plugin/extends/spacing.ts
endigo9740: This is a TW config extension, just provided by our plugin. We essentially overwrite Tailwind's default for all these setting, then add in the scale factor, which allows you to bump up the entire spacing paradigm in your app using a SINGLE setting in your theme. Found here:

https://github.com/skeletonlabs/skeleton/blob/feat/next-tailwind-plugin/packages/skeleton/src/plugin/themes/next.ts#L9

Try adjusting this to 0.75 or 1.25 and watch the entire doc site spacing structure change. It's super cool.

packages/skeleton/src/plugin/utilities/presets.ts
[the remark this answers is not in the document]
endigo9740: Mentioned on Discord - but the `CssClasses` type is not the one we know from the Svelte components. It's just happenstance that it has the same name. It's a record for the classes added to the Tailwind plugin via `core.ts`.

packages/skeleton/src/plugin/utilities/typography.ts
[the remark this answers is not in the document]
endigo9740: > Also this seems more like a Record<string, Record<string, string>>

It is!

sites/next.skeleton.dev/package.json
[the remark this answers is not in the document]
endigo9740: FYI 3.4.1 is required to work around this bug:
https://github.com/tailwindlabs/tailwindcss/issues/12633#issuecomment-1879260319

endigo9740:
@niktek @AdrianGonz97 I've implemented the changes discussed above and on Discord. Minus of course a few things that Adrian will need to be involved in:

1. A fix for the OS dark mode setting. [See the issue outlined here](https://github.com/skeletonlabs/skeleton/issues/2359#issuecomment-1912633056).
2. Nik's comment above around the exec method of the scripts
3. Updates to the Theme type, which I'd like Adrian to be involved in as well (removing `properties_dark`, etc)

That said, let's plan to follow these as follow ups. I will now move forward with merging this into the `next` branch. We can follow up on plans for the above anytime this upcoming week.

PR #2443, opened 2024-01-23
Svelte Simple Datatables Guide

kmalloy24:
## Description

Adds 3rd party integration guide to docs for [Svelte Simple Datatables (SSD)](https://vincjo.fr/datatables/home).

## Checklist

Please read and apply all [contribution requirements](https://www.skeleton.dev/docs/contributing).

- [x] This PR targets the `dev` branch (NEVER `master`)
- [x] Documentation reflects all relevant changes
- [x] Branch is prefixed with: `docs/`, `feat/`, `chore/`, `bugfix/`
- [x] Ensure Svelte and Typescript linting is current - run `pnpm ci:check`
- [x] Ensure Prettier linting is current - run `pnpm format`
- [x] All test cases are passing - run `pnpm test`
- [x] Includes a changeset (if relevant; see above)

PR #2460, opened 2024-02-02
Support nested keys with `tableSourceMapper`

### LINKED ISSUE #2459: Support nested keys with `tableSourceMapper`

In the table component, we have the option of binding data from our data source to the table via the corresponding keys - we can make use of the utility function `tableMapperValues` to accomplish this.

Here is an example from the documentation:
```ts
const tableSimple: TableSource = {
	// A list of heading labels.
	head: ['Name', 'Symbol', 'Weight'],
	// The data visibly shown in your table body UI.
	body: tableMapperValues(sourceData, ['name', 'symbol', 'weight']),
	// Optional: The data returned when interactive is enabled and a row is clicked.
	meta: tableMapperValues(sourceData, ['position', 'name', 'symbol', 'weight']),
	// Optional: A list of footer labels.
	foot: ['Total', '', '<code class="code">5</code>']
};
```

`tableMapperValues` isn't able to work with nested keys like this for example:
```ts
tableMapperValues(sourceData, ['id', 'name.first', 'name.last', 'age'])
```
It would be very useful to support nested keys so that data from more complex structures can also be bound to the table as cell values.

Enhancement

https://www.skeleton.dev/components/tables

chrispump:
## Linked Issue

Closes #2459

## Description

Adds the ability to use nested keys in `tableSourceMapper`

## Checklist

### DIFF
--- a/packages/skeleton/src/lib/components/Table/utils.ts
+++ b/packages/skeleton/src/lib/components/Table/utils.ts
@@ -23,7 +23,18 @@
 export function tableSourceMapper(source: any[], keys: string[]): any[] {
 	return source.map((row) => {
 		const mappedRow: any = {};
-		keys.forEach((key) => (mappedRow[key] = row[key]));
+		keys.forEach((key) => {
+			const nestedKeys = key.split('.');
+			let value = row;
+			nestedKeys.forEach(nestedKey => {
+				if (value && (nestedKey in value)) {
+					value = value[nestedKey];
+				} else {
+					value = undefined;
+				}
+			});
+			mappedRow[key] = value;
+		});
 		return mappedRow;
 	});
 }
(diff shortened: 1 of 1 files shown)


endigo9740:
@chrispump I think you were already in-progress on this before I had a chance to comment in the thread:

- https://github.com/skeletonlabs/skeleton/issues/2459#issuecomment-1922603232

However, I think for now we'll skip this. I will, however, keep this in consideration for v3. Just FYI in v3 we're going to go all in on native HTML table styled with our Table element styles:

https://skeleton-next-git-next-skeleton-labs.vercel.app/tailwind/tables

This will be paired with detailed integration guides for at least one datatable solution per framework for Svelte/React/etc over time. The Svelte version of this will likely be an evolution of this new v2 guide, which should release next week:

https://github.com/skeletonlabs/skeleton/pull/2443

If you haven't already, I'd consider moving away from the table mapper function and handling this directly, and/or moving to the native HTML table approach. This will better prepare you for the future.

Thanks for the contribution all the same though!

PR #2468, opened 2024-02-08
Fixed Wrong OS Mode Lightswitch with SSR off documentation

Sarenor:
## Linked Issue

Fixes an issue with the Lightswitch OS Mode described for SSR Off reported on Discord


endigo9740:
@Sarenor I'm not sure what's up with the linter. The error is not very helpful. Testing locally and I'm only seeing warnings, not errors. So I'm going to ahead and merge and then try to patch the linting warning separately. See if that helps.

PR #2485, opened 2024-02-15
Feat/add disabled to listbox listboxitem

### LINKED ISSUE #2484: Add disabled property to ListBoxItem

Currently, ListBoxItem component does not have a property to disable the inner radio input. 
The feature would add:
- `disabled` property on ListBox that allows to disable every ListBoxItem.
- `disabled` property on ListBoxItem that allows individual disable of an item, when disabled is false on ListBox

There will also be default styles for active disabled items, losing hover and cursor-pointer.

image

image

Enhancement

_No response_

fpribeiro3069:
## Linked Issue

Closes #2484

## Description

Added disabled prop to ListBox and ListBoxItem to prevent selection of inner radio/checkbox inputs.
A ListBoxItem disabled will:
- restrict selection
- lose hover style
- cursor defaults to cursor-default
- default style of `variant-ghost´

## Changsets

## Checklist

### DIFF
--- a/packages/skeleton/src/lib/components/ListBox/ListBox.svelte
+++ b/packages/skeleton/src/lib/components/ListBox/ListBox.svelte
@@ -20,9 +20,9 @@
 	/** Provide classes to set the listbox item active background styles. */
 	export let active: CssClasses = 'variant-filled';
 	/** Provide classes to set the listbox item active disabled background styles */
-	export let activeDisabled: CssClasses = 'variant-soft';
+	export let activeDisabled: CssClasses = 'variant-ghost';
 	/** Provide classes to set the listbox item hover background styles. */
-	export let hover: CssClasses = 'hover:variant-ghost';
+	export let hover: CssClasses = 'hover:variant-soft';
 	/** Provide classes to set the listbox item padding styles. */
 	export let padding: CssClasses = 'px-4 py-2';
 
--- a/packages/skeleton/src/lib/components/ListBox/ListBox.svelte
+++ b/packages/skeleton/src/lib/components/ListBox/ListBox.svelte
@@ -19,8 +19,6 @@
 	// Props (styles) - Item Only
 	/** Provide classes to set the listbox item active background styles. */
 	export let active: CssClasses = 'variant-filled';
-	/** Provide classes to set the listbox item inactive background styles */
-	export let inactive: CssClasses = 'variant-ghost';
 	/** Provide classes to set the listbox item hover background styles. */
 	export let hover: CssClasses = 'hover:variant-soft';
 	/** Provide classes to set the listbox item padding styles. */
@@ -43,7 +41,6 @@
 	setContext('multiple', multiple);
 	setContext('rounded', rounded);
 	setContext('active', active);
-	setContext('inactive', inactive);
 	setContext('hover', hover);
 	setContext('padding', padding);
 	setContext('regionLead', regionLead);
--- a/packages/skeleton/src/lib/components/ListBox/ListBox.svelte
+++ b/packages/skeleton/src/lib/components/ListBox/ListBox.svelte
@@ -7,6 +7,8 @@
 	// Props
 	/** Enable selection of multiple items. */
 	export let multiple = false;
+	/** Disables selection of items. */
+	export let disabled = false;
 
 	// Props (styles)
 	/** Provide class to set the vertical spacing style. */
@@ -17,8 +19,10 @@
 	// Props (styles) - Item Only
 	/** Provide classes to set the listbox item active background styles. */
 	export let active: CssClasses = 'variant-filled';
+	/** Provide classes to set the listbox item active disabled background styles */
+	export let activeDisabled: CssClasses = 'variant-soft';
 	/** Provide classes to set the listbox item hover background styles. */
-	export let hover: CssClasses = 'hover:variant-soft';
+	export let hover: CssClasses = 'hover:variant-ghost';
 	/** Provide classes to set the listbox item padding styles. */
 	export let padding: CssClasses = 'px-4 py-2';
 
@@ -35,9 +39,11 @@
 	export let labelledby = '';
 
 	// Context
+	setContext('disabledItems', disabled);
 	setContext('multiple', multiple);
 	setContext('rounded', rounded);
 	setContext('active', active);
+	setContext('activeDisabled', activeDisabled);
 	setContext('hover', hover);
 	setContext('padding', padding);
 	setContext('regionLead', regionLead);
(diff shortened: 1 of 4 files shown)


packages/skeleton/src/lib/components/ListBox/ListBox.svelte
```diff
@@ -17,6 +19,8 @@
 	// Props (styles) - Item Only
 	/** Provide classes to set the listbox item active background styles. */
 	export let active: CssClasses = 'variant-filled';
+	/** Provide classes to set the listbox item active disabled background styles */
+	export let activeDisabled: CssClasses = 'variant-ghost';
```
Mahmoud-zino, line 23: Typically, we don't offer a "disabled-style" prop. 
Instead, we provide a default styling for disabled elements (* users can always override these styles). However, if we decide to introduce such a feature, let's consider renaming the prop and updating the corresponding comment for clarity.

packages/skeleton/src/lib/components/ListBox/ListBox.svelte
```diff
@@ -35,9 +39,11 @@
 	export let labelledby = '';
 
 	// Context
+	setContext('disabledItems', disabled);
```
Mahmoud-zino, line 42: Please ensure that the context key matches the context value to ensure consistency in styling.

packages/skeleton/src/lib/components/ListBox/ListBoxItem.svelte
```diff
@@ -17,19 +17,23 @@
 	export let name: string;
 	/** Set the input's value. */
 	export let value: any;
+	/** Disables selection of the item. */
+	export let disabled: boolean = false;
 
 	// Context
+	export let disabledItems: boolean = getContext('disabledItems');
```
Mahmoud-zino, line 24: Instead of introducing a new prop to obtain the context, we can utilize the following approach:

```suggestion
	export let disabled: boolean = getContext('disabled');
```
This way, the disabled status will be inferred from the parent component, while still allowing users the flexibility to override this value for individual listbox items.

packages/skeleton/src/lib/components/ListBox/ListBoxItem.svelte
```diff
@@ -17,19 +17,23 @@
...
+	export let activeDisabled: CssClasses = getContext('activeDisabled');
 	export let hover: CssClasses = getContext('hover');
 	export let padding: CssClasses = getContext('padding');
 	export let regionLead: CssClasses = getContext('regionLead');
 	export let regionDefault: CssClasses = getContext('regionDefault');
 	export let regionTrail: CssClasses = getContext('regionTrail');
 
 	// Classes
-	const cBase = 'cursor-pointer -outline-offset-[3px]';
+	const cBase = '-outline-offset-[3px]';
```
Mahmoud-zino, line 36: let's re-add `cursor-pointer` here and override it using the disabled style instead.

sites/skeleton.dev/src/routes/(inner)/components/listboxes/+page.svelte
```diff
@@ -113,6 +114,46 @@
...
+</ListBox>
+			`}
+					/>
+				</svelte:fragment>
+			</DocsPreview>
+		</section>
+		<section class="space-y-4">
+			<h3 class="h3">Disabled</h3>
+			<p>
+				By adding the <code class="code">disabled</code> property,
```
Mahmoud-zino, line 126: using the one prop approach, this should be delete because it is easy to understand and doesn't require additional description.
look at https://www.skeleton.dev/components/tree-views#disabled for example

Mahmoud-zino (changes requested):

packages/skeleton/src/lib/components/ListBox/ListBox.svelte
fpribeiro3069: I've renamed the class and holding on removing if there is an intention in providing this feature, as you mentioned.

packages/skeleton/src/lib/components/ListBox/ListBox.svelte
fpribeiro3069: done

packages/skeleton/src/lib/components/ListBox/ListBoxItem.svelte
fpribeiro3069: yes that was the behavior I was looking, got confused somehow, thanks

packages/skeleton/src/lib/components/ListBox/ListBoxItem.svelte
fpribeiro3069: I've re added it but I'm not sure how can i override the disabled style. Could you elaborate more on that?

sites/skeleton.dev/src/routes/(inner)/components/listboxes/+page.svelte
fpribeiro3069: I've narrowed down the description according to the example given

packages/skeleton/src/lib/components/ListBox/ListBoxItem.svelte
Mahmoud-zino: look at https://github.com/skeletonlabs/skeleton/blob/c6a4ddc02d2883e2ff0c350cc5f02f3b69d0d9ce/packages/skeleton/src/lib/components/TreeView/TreeViewItem.svelte#L320C33-L320C51
for how we handled this in other components :+1:

fpribeiro3069: Ok, let me know if i'm doing this correctly. I've removed entirely the possibility for styling the active disabled items and replaced it with opacity which seems better, using the cDisabled approach.

packages/skeleton/src/lib/components/ListBox/ListBox.svelte
fpribeiro3069: This can be closed if you agree with https://github.com/skeletonlabs/skeleton/pull/2485#discussion_r1501894198

endigo9740:
@fpribeiro3069 the PR was great, just a couple fixes needed:

1. We don't need a changeset for `/sites/skeleton.dev` (the docs), only for packages like `/packages/skeleton`. So I've updated and corrected your changeset.
2. Given disabled is such a common featured in interfaces and form elements, the big detailed example was probably overkill, so I've simplified this down to two simple snippets. This provides less to maintain over time too.

I've made the changes for you and this satisfies all requirements on our end, so I'll merge and this will be part of tomorrow's release. Thanks for your contribution!

PR #2494, opened 2024-02-19
chore: Migrated our TW plugin watcher into its own package

AdrianGonz97:
## Description

- Added the Vite plugin as it's own package in `packages/vite-plugin-tw-plugin-watcher`
- Installed the plugin in the docs site, as well as the svelte and react sites
	- NOTE: AFAICT, since HMR doesn't work in Svelte 5 yet, updates to the TW plugin will not be reflected on the site until the vite server restarts. Hopefully that's resolved soon
- I also adjusted the `vitePreprocess` settings in `@skeletonlabs/skeleton-svelte` for monorepo support for TW (this is mainly to fix IDE related errors)

## Changsets

## Checklist

- [ ] This PR targets the `dev` branch (NEVER `master`)
- [ ] Ensure Svelte and Typescript linting is current - run `pnpm ci:check`


endigo9740:
@AdrianGonz97 this looks great. My only point of concern is limiting this new package to a specific feature. Do you think we should create a catch in case we need multiple utility features like this in the future? Who knows what kind of fun quirks other frameworks might present in the future after all :D

AdrianGonz97:
@endigo9740 I think we can keep it the way it is now as it's in a very simple state. When the time comes that we need more utilities from it, I can transform the package into a generic `utils` package and add more stuff to it like TS instead of JSDoc and what not, which requires additional deps for tooling.

endigo9740:
@AdrianGonz97 sounds like a plan. I'll plan to merge asap.

PR #2498, opened 2024-02-21
feat: Next-avatar

### LINKED ISSUE #2384: NEXT: Avatars
> [!WARNING] 
> This issue is a work in progress.

## Maintainer Requests

- [x] Port forward the [prototype iteration](https://github.com/skeletonlabs/skeleton-next/pull/6).
- [x] Limit the display to a either a built-in image or a fallback slot
- [x] The fallback slot can be used for initials, icons, or even framework-specific `<Image />` components
- [x] Anchor tag support is optional and not advised. This will be hard to replicate in non-Svelte frameworks.
- [x] Use the [new Tailwind Size property](https://tailwindcss.com/docs/size) to control width/height via a single property
- [x] Consider a cross-framework solution for supporting [SVG filters](https://www.skeleton.dev/actions/filters). Actions are Svelte-only!

## Community Requests

* [x] https://github.com/skeletonlabs/skeleton/issues/2446

## Bugs and Issues

* n/a

## Feedback

Mahmoud-zino:
# :warning:  WIP :warning: 
## Linked Issue

Closes #2384

## Description

Avatar

### DIFF
--- a/packages/skeleton-react/src/App.tsx
+++ b/packages/skeleton-react/src/App.tsx
@@ -21,6 +21,9 @@ function App() {
             <a className="anchor" href="/components/accordions">
               Accordions
             </a>
+            <a className="anchor" href="/components/avatars">
+              Avatars
+            </a>
             {/* <a className="anchor" href="/components/app-bars">
               App Bars
             </a> */}
--- a/packages/skeleton-react/src/components/Avatar/Avatar.tsx
+++ b/packages/skeleton-react/src/components/Avatar/Avatar.tsx
@@ -1,0 +1,82 @@
+"use client";
+
+import React from "react";
+import { AvatarProps, FigureProps } from "./types";
+
+// Components
+const Figure: React.FC<FigureProps> = ({
+    // Root
+    base = '',
+    background = '',
+    size = '',
+    font = '',
+    border = '',
+    rounded = '',
+    shadow = '',
+    classes = '',
+    // Image
+    imageSrc = '',
+    imageAlt = '',
+    imageFilter = '',
+    imageBase = '',
+    imageClasses = '',
+    // Children
+    children
+}): React.ReactElement => {
+    return(
+        <figure className={`${base} ${background} ${size} ${font} ${border} ${rounded} ${shadow} ${classes}`}>
+            {imageSrc ? (
+                <img className={`${imageBase} ${imageClasses}`} src={imageSrc} alt={imageAlt} style={{ filter: `url(${imageFilter})` }} />
+            ) : (
+                children
+            )}
+	    </figure>  
+    );
+};
+
+const Avatar: React.FC<AvatarProps> = ({
+    href = '',
+    // Root
+    base = 'overflow-hidden isolate',
+    background = 'bg-surface-400-600',
+    size = 'size-16',
+    font = '',
+    border = '',
+    rounded = 'rounded-full',
+    shadow = '',
+    classes = '',
+    // Image
+    imageSrc = '',
+    imageAlt = '',
+    imageFilter = '',
+    imageBase = 'w-full object-cover',
+    imageClasses = '',
+    // Children
+    children
+}): React.ReactElement => {
+    const figureProps = {
+        base,
+        background,
+        size,
+        font,
+        border,
+        rounded,
+        shadow,
+        classes,
+        imageSrc,
+        imageAlt,
+        imageFilter,
+        imageBase,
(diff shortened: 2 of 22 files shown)


endigo9740:
@Mahmoud-zino I've removed the extra boilerplate in the original post above (changesets and checklist) and approved the PR deployment for the next branch. This should remain in affect for Preview URLS, but let me know if it reverts.

endigo9740:
@Mahmoud-zino I'm going to list out the issues I come across below:

## Documentation

> NOTE: great job following the doc page structure conversions, including multiple markdown files!

1. We should avoid terms like "slots" in the common space, as this is Svelte-specific. Substitute with "Children"
2. Multi-line paragraphs can be a bit odd when there's only 1-2 lines of text. Combine the Fallback descriptions please.
3. When create the code examples, go ahead and abstract the image URL to a variable. Then pass that down to the avatar. It makes for less text the user has to delete when they copy the code. See how I did this for cards.
4. We may want to provide examples using the fallback slot for initials. We should look into recommendations for scaling font size based on container width (it's difficult without Container Queries or SVG Text)

## Cross Framework

1. Demonstrating the interactive border is nice, but it was a remnant of a pre-release or v1 feature we deprecated long ago. Let's go ahead and sunset the code example. Keep the border prop though!
2. After raising this topic with the Skeleton team, we've agreed that the wrapping anchor/href would be detrimental to cross-framework support. Let's go ahead and nix that, which should simplify the component structure.
    - It won't be relevant here since we're removing it, but the href prop should be treated as a "functional" prop, not "style" prop. If we were keeping it this would mean switching from `anchorHref` -> `href` and adding it near the top of the prop definitions.
3. Looks like we're missing parity with the [v2 event handlers](https://www.skeleton.dev/components/avatars?tab=events). Make sure to reference the new Svelte 5 way to handle this (basically a callback function prop). React matches this pattern.
4. I like the inclusion of `imageFilter` but having `style="filter:url()"` hardcoded seems like a bad practice here. We should dynamically add this attribute and value only when populated.
6. For the `types.ts` file, note the convention I used for groups of props, like `Foo ---` with the dashes. Just a small thing, but helps readability.
5. Some of your component props are out of order. `base` is always the first a group, `classes` always the last. Order style props in between in the rough order they appear in the template top-to-bottom. 

## Svelte

1. The Svelte `types.ts` file imports `Action`, but is never used. Drop that.
2. Note the Accordion/Item components include a special `@component` comment above the template. This provides a semantic description when you over the `<Avatar>` component.

## React

1. Since we're dropping anchors, drop the `<Figure>` component in favor of just the single `<Avatar>` component
2. Likewise you can drop `FigureProps` from `types.ts`

endigo9740:
@Mahmoud-zino great job, just a few more nitpicks:

1. In the code examples go ahead and make the component + props single line instead of multi-line. It should fit!
2. At least for Svelte, maybe look into using the [style directive](https://learn.svelte.dev/tutorial/styles) to handle `filter`. React may have an equivalent.
3. Per the Avatar image itself, go ahead and copy the same image source I'm using for Tailwind > Badges > Overlap
4. In most components the "functional" props like src/alt/filter will only appear once. I think by default we should not prefix these. [I've run a team poll on Discord and the team seems to agree](https://discord.com/channels/1003691521280856084/1192531144823025765/1212127482732613733).

Per number 4, the result would be:

```html
<Avatar src="..." alt="..." filter="..." />
<Avatar>SK</Avatar>
<Avatar>(icon)</Avatar>
```

If we had multiple images embedded within the avatar somehow I think the prefix would make more sense. But here I think we can keep it simple. Most folks will understand this and appreciate the shorter attributes.

Mahmoud-zino:
@endigo9740 

> At least for Svelte, maybe look into using the [style directive](https://learn.svelte.dev/tutorial/styles) to handle filter. React may have an equivalent.

I searcher for a react equivalent and didn't find any other reasonable option, if you have found one please let me know.

endigo9740:
@Mahmoud-zino looks like the style attribute recommendation for React is an object with key/value pairs, which is what you've already done!

https://stackoverflow.com/questions/41503150/adding-style-attributes-to-a-css-class-dynamically-in-react-app

So that should be fine!

endigo9740:
@Mahmoud-zino I've pushed a couple minor text edits to the docs. With this, the component should be done! We'll obviously revisit once we have our auto-doc features available, but for now this should be perfect. Thanks for working with me on this!

PR #2528, opened 2024-02-29
appbar-next

### LINKED ISSUE #2381: NEXT: App Bar
> [!WARNING] 
> This issue is a work in progress.

## Maintainer Requests

- [x] Port forward the [prototype iteration](https://github.com/skeletonlabs/skeleton-next/pull/5). Though note the pending PR feedback not yet implemented.
- [x] See [Material Design Top App Bar](https://m3.material.io/components/top-app-bar/overview) for reference and guidelines
- [ ] Try to keep the positions/slots as simple as possible
- [x] Aim to make layout adjustments as simple as possible (maybe Grid is too difficult?)

## Community Requests

* n/a

## Bugs and Issues

* n/a

## Feedback

Mahmoud-zino:
## Linked Issue

Closes #2381

## Description

Ipmlements the v3 App Bar component for Svelte and React.


Mahmoud-zino:
@endigo9740 Some notes and open questions for the first iteration of the appbar:
1. from your feedback on the implementation by Hugos > I think the Main component might be an extra level of abstraction here. (https://github.com/skeletonlabs/skeleton-next/pull/5#discussion_r1443489310)
if we want to get rid of the abstraction layer here, we will have to use grid with a new prop in the Headline so we can add grid `col-span-x`.
2. I've changed the name rowMain to toolbar which in my opinion is clearer.
3. I have applied implemented all a11y keyboard shortcuts, the optional ones too (home, end).
4. should we introduce a prop to opt out a11y keyboard shortcuts for users wanting to implement their own version of it?
5. I have added full RTL support including inverting the keyboard arrow-keys (since they should be inverted in RTL) and included a RTL example in svelte and react projects but not in the docs, let me know what you think about it and if I should change or even delete the implementation.
6. I took a look at the examples from https://mui.com/material-ui/react-app-bar/ , all features included can be modified and added by our users with minimal effort, I have included an example of sticky in svelte and react projects to show that.

endigo9740:
@Mahmoud-zino thanks for submitting! First off this looks great at first glance!

First off, it looks like we have a few merge conflicts. If you could please try to get those sorted. It looks like you should be able to do this through the GitHub interface. Just tap the "resolve conflicts" button here. If you can't access this let me know.

Screenshot 2024-03-04 at 11 47 55 AM

## Docs

* I noticed you abstracted the primary example to it's own fine, but not the two follow ups (per framework). Those should be abstracted as well. We're making a commitment in v3 to show a fully functional implementation per each flushed out example, imports and all. See the other components as reference.
* If we're going that far we might as well make use of the Preview component too.
* The preview icons you used are great, but 24px may be too large. Drop the right-side down to 20px. Though consider keeping the arrow icon at it's current size. It'll likely feel more balanced this way.
* For the featured example, let's do order of imports as: React, then Skeleton, then icons.
* For the featured example, you can likely combine the `center` component by doing something like:
```tsx
<ToolBar.Center><!-- (optional) --></ToolBar.Center>
```
* Make sure when you create headings you keep consistent casing. Should be "Toolbars and Title"
* It's not immediately obvious to me that the Headline component should proceed the Toolbar.
* Take a look at the indentation in your Svelte example. Everything is indented by 8 spaces instead of 4. Prettier should be able to fix this (though your branch behind the update to implement that)
* In the Svelte example, make sure to add the "FIXME" comment to the empty props for the Astro bug. It'll make it easier to find and remove in the future.

> NOTE: the only exception to the abstracted example code block is when the change can be summarized in a single line. Just in case you're wonder why I did that else where!

## General

* I have mixed feelings on the a11y implementation. Not your fault, I think this is on me for designating the previous version as following the ARIA APG Toolbar guidelines. I don't think that's technically accurate. Let's go ahead and remove that designation and keep this as a simple presentational component. No JS logic for now. For tabbing, consider swapping to [tabindex](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex) which should support LTR/RTL
* We may not need the label/labeledBy props with this change btw
* Not to be too much of a pain, but the more I think about it the more I think `grid` is the right option for the top row styles. There's going to be use cases where folks want an even 3 columns (so the middle slot can center properly) and flex just will never satisfy that. Given we have the `base` styles now, folks can opt to replace with that if they choose! So default grid, flex optional. This will help simplify the styling for a few of the children (removing classes)
* The wrapping parent element should be a semantic `<header>` instead of `<div>`
* I'd also consider making the Toolbar/Headline rows `<section>` elements instead of `<div>`

## React

* This was an issue I missed in the previous component PR audits, but we can drop the `React.ReactElement` return type. This should be implicit! Feel free to patch the other React components if you don't mind! This will help improve readability.
* In the `types.ts` file, this is React-specific so we can use the term "component". So let's update descriptions to read like "Set's the lead component's base styles", etc.

## Svelte

* Make sure to add the Svelte-specific `@component` comment above the template. This adds a IntelliSense description when hovering the component.
* In the `types.ts` file, this is Svelte-specific so we can use the term "snippet". So let's update descriptions to read like "Set's the lead snippet element's base styles", etc.

endigo9740:
@Mahmoud-zino this is looking so good!!

A few follow ups:

- The headline example feels redundant with the main example at the top. Maybe it's not needed?
- The indentation for the React code snippets is messed up. Icons are only indented a single space? They look fine in VS Code. Maybe try deleting the indentation and recreating it. Sometimes it can get a little funky.
- It doesn't look like `tabindex` was implemented. I might still recommend that to ensure basic RTL support.
- The prop `toolbarColumns` should likely be renamed to follow the Tailwind convention of grid-cols-[...]`. So `toolbarCols` or `toolbarGridCols`. I'm fine with either as long as we remain consistent going forward. The latter might work better for React since it won't include the "toolbar" prefix.

The only other idea we might want to consider is we've had a few recent inquiries about mobile responsive styles for toolbars. We of course make the toolbar flexible, but the user is responsible for responsibly showing/hiding columns and resizing slot content as needed. Maybe we explain that in a quick blurb and perhaps replace the third example showing how you can show/hide columns at certain breakpoints via Tailwind's breakpoint variants?

Everything else looks great. Great job on this one!

Mahmoud-zino:
@endigo9740 I have applied the changes and added the responsive example.

> It doesn't look like tabindex was implemented. I might still recommend that to ensure basic RTL support.

this is not required to support RTL, it is automatically applied.

endigo9740:
@Mahmoud-zino Ive pushed a number of final tweaks to the PR branch. Check the commit here:

- https://github.com/skeletonlabs/skeleton/pull/2528/commits/b0909d64781ca9271c10e0ba8f677f818cbc6be4

This includes:

- Let's use a file naming convention of `ExampleFoo.svelte|tsx` for example components, where Example is listed first.
- I've fixed an issue where a code block was using syntax of `react`, which doesn't exist (should be `tsx`)
- I've made minor copy changes to the Svelte/React doc pages
- I moved a couple styles around in the responsive examples to use our new prefixed props

I love how you're handling the responsive example though where the title moves down to the headline position!

PR #2541, opened 2024-03-09
feat: tabs-next

### LINKED ISSUE #2395: NEXT: Tabs

## Structure

React:

```html
<Tabs>
    <Tabs.List>
        <Tabs.Control name="foo">...</Tabs.Control>
        <Tabs.Control name="bar">...</Tabs.Control>
        <Tabs.Anchor>...</Tabs.Anchor>
    </Tabs.list>
    <Tabs.Content>
        <Tabs.Panel value="foo">...</Tabs.Panel>
        <Tabs.Panel value="bar">...</Tabs.Panel>
    </Tabs.Content>
</Tabs>
```

Svelte:

```html
<Tabs>
    {#snippet list()}
        <Tabs.Control name="foo">...</Tabs.Control>
        <Tabs.Control name="bar">...</Tabs.Control>
        <Tabs.Anchor>...</Tabs.Anchor>
    {/snippet}
    {#snippet content()}
        <Tabs.Panel value="foo">...</Tabs.Panel>
        <Tabs.Panel value="bar">...</Tabs.Panel>
    {/snippet}
</Tabs>
```

## Maintainer Requests

- [x] Should follow the [ARIA APG guidelines](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/)
- [x] Tabs should embed a radio input to control state. See v2 and our new React reference repo for help with this.
- [x] Reviewing other frameworks ([ref 1](https://ui.shadcn.com/docs/components/tabs), [ref 2](https://mantine.dev/core/tabs/)), they tend to have more explicit sectioning between the parent, list, panels, controls, and anchors. See the proposed structure in the code blocks above.
- [x] The default styling can be based on [Material Design](https://m3.material.io/components/tabs/guidelines) (wide underline) until we can provide a proper Figma mock. This should still be customizable to represent a more "enclosed" style like [Shadcn shows here](https://ui.shadcn.com/docs/components/tabs) (or our v2 "bottom nav" reference).
- [x] Tabs should support either a alignment (ex: left) or evenly spaced across the width ([ref](https://mantine.dev/core/tabs/#tabs-position))
- [x] Should support setting a default tab - which should be possible through the radio input state
- [x] should replicate most/all relevant selection event handlers from v2
- [ ] Make considerations for how icons can be implemented. Both in a above/below and side-by-side approach.
- [ ] We should support Tabs for anchor links in addition to local state management (see v2 for reference)
- [ ] Bohdan has some requests regarding more unique animations, with a sliding underline bar. We should consider this as part of the larger discussion around v3 animations that work cross-framework.

### Optional Requests

- [ ] (optional) Consider support for Tab Anchors to support [React Router](https://mantine.dev/core/tabs/#usage-with-react-router) and the [Next.js Router](https://nextjs.org/docs/pages/building-your-application/routing/redirecting). This may be redundant with the above event handlers request.
- [ ] (optional) Let's follow up on whether the React tabs can replace a couple doc-site only components we're using - including Preview and Framework selection. There's some known gotchas here.
- [ ] (optional) I'd love for the content panels to slide left/right on selection (ala Material Design). But don't stress if this is too difficult for the initial draft.

## Community Requests

* n/a

## Bugs and Issues

* n/a

## Feedback

Mahmoud-zino:
## Linked Issue
Closes #2395

## Description
tabs-next


endigo9740:
@Mahmoud-zino this is great, but since the feature is not added to the main doc site it won't generate a preview for it. So I can't check until I'm back to my workstation - likely on Monday. In the meantime can you help remind me to flush out the tickets before we start implementing features?

https://github.com/skeletonlabs/skeleton/issues/2395

I realize you've got time to get a head start and that's great, but I want to avoid a situation where there's a major change not documented that doesn't get missed. To avoid a lot of duplicated effort or even requiring a feature to be rebuilt from the ground up.

There's a few things for Tabs specifically I need to put some thought into:

- If we want to stick with the TabGroup/Tab naming scheme. TabAnchor came in later, and to some degree it feels a bit clunky.
- The default styling. I'm working with Bohdan on the Figma side to try and get ahead of us so we have some references soon
- Any other notable tweaks or feature additions or changes

Figma stuff aside, I'll try to get the rest asap.

endigo9740:
@Mahmoud-zino I've had a moment to run through and write out my proposed updates and changes for the Tabs component feature:

- https://github.com/skeletonlabs/skeleton/issues/2395

Let me know if you have any specific questions or feedback regarding these.

Just FYI I'm going to try to devote some time this afternoon to flushing this out for a few other upcoming components so this is ready before you start next time (sorry about that!).

Likewise I'm going to ping Bohdan to see if he can work up a mock for the tab style for us that you can follow. If you need help with the design I can always jump in an help!

Mahmoud-zino:
@endigo9740 
Since the component is quite substantial and I've encountered some new cases (not documented), I'd like to get your opinion before implementing the React part and the documentation for the component. You can check out the test page in Svelte for all the different examples.
## general Notes
1. I anticipate there might be some requests for changes in naming and possibly reducing the number of props. I'm open to applying these changes since naming was one of the trickier aspects during the implementation of this component.
2. I've implemented a focus outline similar to the one provided in the mocks, but like you, I'm not entirely satisfied with it. Perhaps you have some suggestions in this area.
## functional Notes
1. I noticed there's no binding for groups in the example you provided. I wonder if you forgot about it or if you have something else in mind.
2. The events are implemented similarly to context props. You can capture events from the parent component or specify a different event for each of the children.
## ARIA Notes
1. I've followed most of the v2 ARIA functionalities since it already had a good implementation.
2. The only difference would be if a panel contains a focusable element, the panel itself won't be focusable, which is mentioned in the notes of:
> When the tabpanel does not contain any focusable elements or the first element with content is not focusable, the tabpanel should set tabindex="0" to include it in the tab sequence of the page.

3. I've also included 'Home' and 'End' for the keyboard handling of tabs.
## RTL Notes
1. I've ensured that RTL is functional. You can see the last example in the Svelte page.
2. I considered following your advice on handling the arrow functionality in RTL by using tabIndex, but it's not possible for this example since the tabs have the tabindex set to 0 only when they are active.

@Hugos68 the same goes for you. I'd appreciate your feedback on the Svelte component whenever you have the time :wink:

endigo9740:
@Mahmoud-zino I'll review and provide a more in-depth response when I do my full review. But I did want to clarify on this point:

> I noticed there's no binding for groups in the example you provided. I wonder if you forgot about it or if you have something else in mind.

I can't list every single feature that is available in v2's components. So please do make sure you're taking that as the foundation of requirements, and the reqs listed in the ticket as the additions/changes to make in transit to v3. To put it another way, we 110% want to keep the binding model. It worked perfectly for state management.

Plus this is why we put all the work into the React equivalents. To ensure we could replicate this on that side as well!

This goes for any components that embed form elements.

Sarenor:
I've looked through this, there's nothing that jumped out at me. I'd only suggest to check the conversation regarding ``rxPropname`` in #team-core :)

packages/skeleton-svelte/src/lib/components/Tab/Tabs.svelte
```diff
@@ -0,0 +1,92 @@
...
+	setContext('padding', controlsPadding);
+	setContext('rounded', controlsRounded);
+	setContext('spacingY', controlsSpacingY);
+	setContext('cursor', controlsCursor);
+	setContext('classes', controlsClasses);
+	setContext('onclick', onclick);
+	setContext('onkeypress', onkeypress);
+	setContext('onkeydown', onkeydown);
+	setContext('onkeyup', onkeyup);
+	setContext('onchange', onchange);
```
Hugos68, line 60: I see that we're utilizing context a bunch, but all these properties won't be reactively set through context.
If the user would change the `controlsBase` for example that change wouldn't get to all the consumers of this context.

[Here is an example of reactive vs non-reactive context in svelte 5](https://svelte-5-preview.vercel.app/#H4sIAAAAAAAACpWSTWvDMAyG_4pnBkkha-5pFhhlx42x67xDlqidIZGNrZSO4P8-YidNuo9CT0avXknWg3q-kw1Ynr31HMsWeMYftOYJpy89BPYADQFPuFWdqQYlt5WRmgqBgmSrlSH2UhpAYjujWhat0xCuQ2m0WRi3n7KpTz4fLW0oqAFih7LpgN2zW0slQRw91pLYE9xEq43APJ3nYy5Rd8Q-JNZZqEqDPv6o96Lzf839uCIPY4uhU3AVPOGtquVOQs0zMh245AQjWK7k0TMLtFVIcCTmxn2XOKZN-3FXN2yrjdI2Xo0g5g5xhArvDJQVyQNESaj5y7fw9H4M7Sec8WqSBBmgzmDQg-aGx_3CKzC3jSKWXkTkeV5NaH-REAqqFNr5Fvb_4fAcRnNFxx_WM9v56ejiedEnm24lT3Xhs69zpqLjes7-ZvHuvgEU4SWJRwMAAA==)

We should probably all get together and defined some guidelines around this.

endigo9740: @Mahmoud-zino I'm partially with Hugo here, we have a lot of context values. The issue is we're trying to feed everything down through the parent. Where as we should localize those props to come in on the child components themselves.

Yes, there's some benefits to having a single entry point from from a consuming developer's perspective, but this at the cost of our code maintainability. This was a pattern we followed in v2 that I don't think we should continue forward in v3.

My suggestion as a rule of thumb is: if the parent doesn't use the value, it should not contain a prop for it.

As per reactivity. We shouldn't add this just because we can. We should add this only when it's absolutely required given the extra hoops required to make it work. To me, most style prop don't need to be reactive. That's a luxury, not a necessity. Honestly same goes for even props. I can't think of a lot of scenarios where your actual handler function will change, as opposed to just handling the result within the handler in a different manner.

endigo9740:
@Mahmoud-zino this wasn't marked as "ready for review" so feel free to ignore if things are still in progress. But thought I'd chime in with a quick overview. I'll approach this from three angles: design, UX/A11y, and implementation (the code)

## Design

I agree with some of the concerns around hover/focus states in in the mocked designs, but it looks like Bohdan's been flushing this out a bit more. I actually rather like this current iteration:

Screenshot 2024-03-19 at 2 40 55 PM

I'd also recommend being mindful of focus outline styles. Though this should be less of an issue with the more padded version in the mock.
Screenshot 2024-03-19 at 3 07 21 PM

1. The hover/focus portion would be be detached from the surrounding "clickable" area. The clickable region would be teh full cell, while the visible portion is isolated within
2. To accomplish this you can likely use [Tailwind group styles](https://tailwindcss.com/docs/hover-focus-and-other-states#styling-based-on-parent-state).
3. Additionally, your version is missing the this line that stretches below all tabs, and potentially the width of the container
4. The hover/focus states could of course use the preset called: `preset-tonal-primary`

## UX/A11y

- So I like that you're taking so many considerations for interaction, but strangely tabbing is not functional. That was the first interaction I tested. Arrow keys are cool, they are part of the aria spec, but tabbing should always be standard.
- I believe you stated in a previous message that this is because you're disabling `tabindex` for the active element, but I'm not sure why. This is not harmful. If that's restricting this use case then do away with that. Labels should be naturally tab enabled, so a lot of this should work automatically
- The panel element is currently focusable. This should not be the case. This is not a selectable or interactive element, the focus should jump to whatever content is within the panel (assuming there is something). I see your comments mentioning the ARIA guidelines specify this, but I'm worried that may be misinterpreted. It doesn't seem right. Their own demos don't do this.

## Implementation

Just FYI I'll highlight some specific examples in the code itself separate from this message, but a few odds and ends:

1. I think in the future we should focus on getting the structure and design set first, THEN review options for a11y and keyboard interaction. The two overlapped and make it feel rigid or difficult to change things.
2. We have to be cautious of using `preventDefault` within any sort of keyboard interactions. There are good natural interactions that might be interrupted by doing this. Leaving us to have to reinvent in the wheel. For example, space/enter should just work naturally with an element using Labels for state. We get that for "free".
3. I may revisit the keyboard interactions for this feature in the future. Maybe for the time go ahead and comment these out and I'll do a separate dedicated review or possibly implement that portion myself.

packages/skeleton-svelte/src/lib/components/Tab/Tabs.svelte
```diff
@@ -0,0 +1,92 @@
...
+		if (!panelElem) return;
+		const hasFocusableElements =
+			panelElem.querySelectorAll('button, [href], input, select, textarea, [tabindex]:not([tabindex="-1"])').length > 0;
+		panelTabIndex = hasFocusableElements ? -1 : 0;
+	});
+</script>
+
+<!-- @component A Tab parent component. -->
+
+<div class="{base} {spaceY} {classes}" data-testid="tab-group">
```
endigo9740, line 75: `data-testId` should like just be set to a value of `tabs` to match the component naming convention.

packages/skeleton-svelte/src/lib/components/Tab/TabsControl.svelte
```diff
@@ -0,0 +1,126 @@
+<script lang="ts">
+	import { getContext } from 'svelte';
+	import type { TabsControlProps } from './types.js';
+
+	let {
+		name,
+		group,
+		title = '',
```
endigo9740, line 8: The `title` prop feels out of place here. Let's maybe drop this. The ARAI attributes already provide enough metadata here.

packages/skeleton-svelte/src/lib/components/Tab/TabsControl.svelte
```diff
@@ -0,0 +1,126 @@
...
+	const selected = $derived(group === name);
+	const rxActive = $derived(selected ? active : inactive);
+
+	let elemInput: HTMLInputElement;
+
+	function onTabKeyDown(event: KeyboardEvent) {
+		onkeydown(event);
+		if (!['Enter', 'Space', 'ArrowRight', 'ArrowLeft', 'Home', 'End'].includes(event.code)) return;
+
+		event.preventDefault();
```
endigo9740, line 51: Avoid this in the global event handler. Keep it localized only when overwriting very specific behavior. This can inadvertently damage built in features like enter/space to select.

packages/skeleton-svelte/src/lib/components/Tab/TabsControl.svelte
```diff
@@ -0,0 +1,126 @@
...
+
+		if (['Enter', 'Space'].includes(event.code)) {
```
endigo9740, line 53: See above, this should be unnecessary if implemented properly.

endigo9740: One again, I'd say let's disable and rethink our keyboard interaction strategy. I'll save my comments for the rest of this section.

packages/skeleton-svelte/src/lib/components/Tab/TabsControl.svelte
```diff
@@ -0,0 +1,126 @@
...
+			nextTabInput.click();
+			(nextTab as HTMLDivElement).focus();
+		}
+	}
+</script>
+
+<!-- @component A Tab Control component. -->
+
+<label class="{base} {text} {justify} {rxActive} {background} {padding} {rounded} {spacingY} {cursor} {classes}" {title}>
+	<li
```
endigo9740, line 106: Let's change `li` -> `div`. I understand the goal is improved semantics, but the label is already fulfilling this role for the most part. Which replaces the `button` in the ARIA APG guidelines.

packages/skeleton-svelte/src/lib/components/Tab/TabsControl.svelte
```diff
@@ -0,0 +1,126 @@
...
+		if (nextTabInput) {
+			nextTabInput.click();
+			(nextTab as HTMLDivElement).focus();
+		}
+	}
+</script>
+
+<!-- @component A Tab Control component. -->
+
+<label class="{base} {text} {justify} {rxActive} {background} {padding} {rounded} {spacingY} {cursor} {classes}" {title}>
```
endigo9740, line 105: Echoing: avoid style duplication between the label and wrapping element within. Pick one or the other. This adds extra overhead for users to understand.

packages/skeleton-svelte/src/lib/components/Tab/TabsControl.svelte
```diff
@@ -0,0 +1,126 @@
...
+		tabindex={selected ? 0 : -1}
+		aria-selected={selected}
+		data-testid="tabs-control"
+		role="tab"
+		class="{tabBase} {tabRounded} {tabOutline} {tabClasses}"
+		onkeydown={onTabKeyDown}
+		{onkeypress}
+		{onkeyup}
+	>
+		<div class="h-0 w-0 flex-none overflow-hidden">
```
endigo9740, line 117: Can you provide this comment to add context to why we use this method to hide the input:
https://github.com/skeletonlabs/skeleton/blob/master/packages/skeleton/src/lib/components/Tab/Tab.svelte#L128

packages/skeleton-svelte/src/lib/components/Tab/TabsControl.svelte
```diff
@@ -0,0 +1,126 @@
...
+			<input bind:this={elemInput} type="radio" bind:group {name} value={name} onchange={() => onchange(group)} {onclick} tabindex="-1" />
```
endigo9740, line 118: The `tabindex` feels redundant here. We never had trouble with focus on the v2 version afaik. Not sure why this is present.

packages/skeleton-svelte/src/lib/components/Tab/TabsControl.svelte
```diff
@@ -0,0 +1,126 @@
...
+		</div>
+		{#if children}
+			<div class="{contentBase} {contentSpaceX} {contentClasses}">
```
endigo9740, line 121: This extra div seems redundant. We already have so many layers here. It's like an onion heh.

packages/skeleton-svelte/src/lib/components/Tab/TabsPanel.svelte
```diff
@@ -0,0 +1,21 @@
...
+		classes = '',
+		// Snippets
+		children
+	}: TabsPanelProps = $props();
+</script>
+
+<!-- @component A Tab Panel component. -->
+
+{#if value === group && children}
+	<section class="{base} {classes}">
```
endigo9740, line 18: `section` might be too opinionated here. The ARIA example uses a `div`

packages/skeleton-svelte/src/lib/components/Tab/TabsPanel.svelte
```diff
@@ -0,0 +1,21 @@
+<script lang="ts">
+	import type { TabsPanelProps } from './types.js';
+
+	let {
+		value,
+		group,
+		// Root
+		base = '',
```
endigo9740, line 8: If there are no inherent base/classes styles then having both feels redundant. I'd say we default to one or the other. `bases` seems most obvious.

endigo9740 (changes requested):

packages/skeleton-svelte/src/lib/components/Tab/TabsControl.svelte
```diff
@@ -0,0 +1,126 @@
...
+	}: TabsControlProps = $props();
+
+	const selected = $derived(group === name);
+	const rxActive = $derived(selected ? active : inactive);
+
+	let elemInput: HTMLInputElement;
+
+	function onTabKeyDown(event: KeyboardEvent) {
+		onkeydown(event);
+		if (!['Enter', 'Space', 'ArrowRight', 'ArrowLeft', 'Home', 'End'].includes(event.code)) return;
```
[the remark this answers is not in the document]
Mahmoud-zino: this will simplify the code a lot... I made some changes but I find it necessary for multiple keys that share some functionalities.

packages/skeleton-svelte/src/lib/components/Tab/TabsControl.svelte
Mahmoud-zino: It is not redundant, without it tabbing to next element will require you to tab twice (hidden elements will still be tabbed)

packages/skeleton-svelte/src/lib/components/Tab/TabsControl.svelte
Mahmoud-zino: this helps providing styles for the children (flex, grid, etc..), I think it is nice to have and doesn't require much maintenance, otherwise when users wants to have tabs with text and icons they will have to wrap it every time.
let me know if this doesn't convince you and I will delete it :+1:

packages/skeleton-svelte/src/lib/components/Tab/TabsPanel.svelte
Mahmoud-zino: I would actually suggest to keep classes instead, since users will most likely avoid writing to base because in most of the other components we have some styles in base, so we will get a lot of questions on where is classes for this component.
Again let me know if you are not OK with the decision and I will change it.

Mahmoud-zino:
@endigo9740 

some follow-ups

> So I like that you're taking so many considerations for interaction, but strangely tabbing is not functional. That was the first interaction I tested. Arrow keys are cool, they are part of the aria spec, but tabbing should always be standard.
> I believe you stated in a previous message that this is because you're disabling tabindex for the active element, but I'm not sure why. This is not harmful. If that's restricting this use case then do away with that. Labels should be naturally tab enabled, so a lot of this should work automatically

Just like Skeleton-v2 tabs, I followed the [ARIA tabs with automatic activation](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-automatic/) example and built tabbing after that model. if you want we can for sure opt for [ARIA tabs with manual activation](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-manual/) which feels more natural. (I wanted to match as much as possible of v2 :wink: )

The panel element is currently focusable. This should not be the case. This is not a selectable or interactive element, the focus should jump to whatever content is within the panel (assuming there is something). I see your comments mentioning the ARIA guidelines specify this, but I'm worried that may be misinterpreted. It doesn't seem right. Their own demos don't do this.
Implementation
Just FYI I'll highlight some specific examples in the code itself separate from this message, but a few odds and ends:

> I think in the future we should focus on getting the structure and design set first, THEN review options for a11y and keyboard interaction. The two overlapped and make it feel rigid or difficult to change things.

I agree, it will makes it easier to review on batches.

> We have to be cautious of using preventDefault within any sort of keyboard interactions. There are good natural interactions that might be interrupted by doing this. Leaving us to have to reinvent in the wheel. For example, space/enter should just work naturally with an element using Labels for state. We get that for "free".

again totally agree, I wanted to have it after the enter, space keys but it seems I forgot to do so.

> I may revisit the keyboard interactions for this feature in the future. Maybe for the time go ahead and comment these out and I'll do a separate dedicated review or possibly implement that portion myself.

I will leave them uncommented for the time-being because it will make the automatic activation of the tabs nonfunctional. I would appreciate your review on them and I believe you will get to a very similar result as the one I have (in case we go forward with the automatic activation of tabs) because I spent some time optimizing the code to make it as readable and functional as possible.

packages/skeleton-svelte/src/lib/components/Tab/TabsControl.svelte
```diff
@@ -0,0 +1,126 @@
...
+		active = getContext<string>('active'),
+		inactive = getContext<string>('inactive'),
+		background = getContext<string>('background'),
+		padding = getContext<string>('padding'),
+		rounded = getContext<string>('rounded'),
+		spacingY = getContext<string>('spacingY'),
+		cursor = getContext<string>('cursor'),
+		classes = getContext<string>('classes'),
+		// Tab
+		tabBase = '',
```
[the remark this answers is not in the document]
Mahmoud-zino: reference https://github.com/skeletonlabs/skeleton/pull/2541#discussion_r1541900137

packages/skeleton-svelte/src/lib/components/Tab/Tabs.svelte
```diff
@@ -0,0 +1,92 @@
...
+	setContext('classes', controlsClasses);
+	setContext('onclick', onclick);
+	setContext('onkeypress', onkeypress);
+	setContext('onkeydown', onkeydown);
+	setContext('onkeyup', onkeyup);
+	setContext('onchange', onchange);
+
+	let panelElem: HTMLDivElement | undefined = $state(undefined);
+	let panelTabIndex = $state(0);
+	// if the tabpanel doesn't contain focusable element, it will be focusable following the ARIA guidelines.
```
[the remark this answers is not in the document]
Mahmoud-zino: reference https://github.com/skeletonlabs/skeleton/pull/2541#issuecomment-2023826220

packages/skeleton-svelte/src/lib/components/Tab/TabsControl.svelte
Mahmoud-zino: this is a common pattern to group keys functionality, using a switch for tow cases with the same behavior will make the code harder to read.

endigo9740:
@Mahmoud-zino I've gone ahead and close all feedback comments above so we can start "fresh". During my review I noted a lot of areas where would clean up and minimize the markup and styles, so I've done this. Likewise I've made a few cosmetic updates to try and more closely replicate the styles in Figma. I'll likely record a video to showcase some tips on how inspect the Figma doc to find specific values (colors, sizes, etc), which should make this process easier going forward.

## Preview Page

- Created variables for the icons
- Set the icons to 16px to match Figma
- Renamed the `panel()` snippet to `panels()` to be more semantic
- Added an "Icons Only" set to preview

## Design

> NOTE: this just details more "sweeping" changes. There's lots of micro adjustments via default prop values.

- I need to communicate with Bohdan to explain how Tailwind treats focus/outline styles. FYI for now we'll be off-spec.
- I've changed the default Tab width to `w-full` versus `w-fit`.
- I'm now using [Tailwind's group modifier](https://tailwindcss.com/docs/hover-focus-and-other-states#styling-based-on-parent-state) to affect the hover state. Now uses a tonal preset to match Figma.

## Components

## General

- Swapped from `space` properties to `gap`. This better prepares us for a vertical layout.

## Tabs (root)

- Removed the wrapping div around `tabpanels`, which is not part of the ARIA example. The `role` and `tabindex` have been moved to the panel component. I also removed the `$effect()` logic, which introduces an issue I'll talk about below.
- Renamed snippet `panel` -> `panels` to be more semantic that this contains multiple items

## Tabs.Control

- Removed nearly all classes from the first `label > div`. This element should be considered structural, not cosmetic. All styles have been redistributed the root label or the "content" element within. This greatly simplifies the number of style props users have to deal with.
- Modifying a number of default styles to better match the Figma docs. This includes using a preset-tonal style for group-hover, removing the outline styles (for now), and adding a missing base class for one element.
- I've updated the `onKeyDown` handler to strip out the space/enter features - which are never possible to use
- Updated the `onKeyDown` handler to document each portion of the process. There's a lot going on here, and there may still be room for further optimization

## Tabs.Panel

- Moved the `role` and `tabindex` attributes here to better match the ARIA example.
- Added an optional `id` prop to match the ARAIA example

--- 

## Tab Panel Tabindex Change

Per the panel change - I know you implemented the extra wrapping div to handle the auto-selection between the panel and panel contents (when a focusable element is present). We can likely migrate this logic from the Tabs -> Panel component. I unfortunately have run out of time to do this today though (sorry!)

---

## Next Steps

I'm like 90% happy with the state of this now, but I want to revisit tomorrow to try and simplify the styles a bit more. The reason we want to strive for the minimal markup and style props, is because it provides less overhead for users to need to know about, and makes the component easier to restyle if they want a dramatically different look. If we have too many options, this becomes really difficult to do.

Let's plan to sync tomorrow on some of these changes. But overall great job bud!

endigo9740:
@Mahmoud-zino I've go ahead and merged this PR. The reason being is this had some linting changes that spilled over into the documentation section. We need the doc section as "clean" as possible before https://github.com/skeletonlabs/skeleton/pull/2561 is merged.

I'm going to related this PR to make it specific to the Svelte component. Feel free to start a follow up PR to port the React version next. By doing this (assuming you start after #2561 is merged) you'll actually be able to implement both the Svelte and React components in the docs again!

Feel free to check with me or Hugo if you need additional details around the doc updates.

PR #2549, opened 2024-03-13
NEXT theme type improvements

endigo9740:
## Linked Issue

#2360

## Description

Consolidates the preset and custom theme types into a single `Theme` type. Removes `properties_dark` and `enhancements` from the type. Updates the `properties` to match the current v3 structure.

### DIFF
--- a/packages/skeleton/scripts/convert-theme.ts
+++ b/packages/skeleton/scripts/convert-theme.ts
@@ -2,7 +2,7 @@ import { readFileSync } from 'fs';
 import { writeFile } from 'fs/promises';
 import postcss from 'postcss';
 import postcssJs from 'postcss-js';
-import type { PresetTheme } from '../src/plugin/themes/index.js';
+import type { Theme } from '../src/plugin/themes/index.js';
 
 // Converts a theme's .css file into a .ts file.
 export async function convertTheme(name: string) {
@@ -18,10 +18,10 @@ export async function convertTheme(name: string) {
 
 	const theme = {
 		name,
-		properties: properties,
-		enhancements: { ...cssInJs },
-		properties_dark: {}
-	} satisfies PresetTheme;
+		properties: properties
+		// enhancements: { ...cssInJs },
+		// properties_dark: {}
+	} satisfies Theme;
 
 	// Creates the generated CSS-in-JS file
 	await writeFile(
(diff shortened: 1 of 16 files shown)


endigo9740:
@AdrianGonz97 @ryceg @Hugos68 tagging you for review on this as you guys are our resident Typescript pros. The changes are fairly simple, so hopefully nothing to egregious here!

Just FYI I do expect us to add a few additional properties between now and launch. But I'll do my best to ensure those are addressed evenly between all v3 themes.

I also have `/themes/legacy` for our v2 and prior themes. Those are just set aside for the moment while we finalize the v3 theme structure. Then I plan to either port those forward or recreate spiritual successors to take advantage of the additional settings.

packages/skeleton/scripts/convert-theme.ts
```diff
@@ -18,10 +18,10 @@ export async function convertTheme(name: string) {
 
 	const theme = {
 		name,
-		properties: properties,
-		enhancements: { ...cssInJs },
-		properties_dark: {}
-	} satisfies PresetTheme;
+		properties: properties
```
ryceg, line 21:
```suggestion
		properties
```

packages/skeleton/src/plugin/themes/legacy/gold-nouveau.ts
```diff
@@ -1,4 +1,4 @@
-import type { PresetTheme } from '../index.js';
+// import type { Theme } from '../index.js';
 
 const goldNouveau = {
```
ryceg, line 3: I think an extension of Theme that had lines 6 through 19 would be a good idea, so that the themes are typechecked

AdrianGonz97 (approved):
Deleted one more thing that wasn't needed anymore, but the rest LGTM

ryceg (commented):
LGTM!

packages/skeleton/src/plugin/themes/legacy/gold-nouveau.ts
endigo9740: I don't think GitHub is aligning this message in the right place, but if you're referring to the `properties_dark` and `enhancements` - those aren't carrying over to v3!

Hugos68:
Was asleep but looks good 😁

Issue #2390, opened 2024-01-03
NEXT: Progress Bars


endigo9740:
This will act as a hub to centralize this information.

## Structure

 ```html
 <Progress id="progress" value={70} max={100}  />
 ```

## Maintainer Requests

The following requests are coming straight from the Skeleton team. These are highly likely be implemented:

- [ ] Must fulfill all [a11y guidelines](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/progressbar_role)
- [ ] Rename to `<Progress>` to follow common convention
- [ ] Supports both a determinate and indeterminate values; the latter is animated
- [ ] DO NOT include a label - this should not be part of the component
- [ ] Ensure the docs link to the native alternative (see references below)
- [ ] Provide style props to enable customizing the animation settings (effect/duration/etc)
- [ ] Material provides a great reference for [default animation styles](https://m3.material.io/components/progress-indicators/guidelines)
- [ ] Ensure consistency with the [Progress Radial](https://github.com/skeletonlabs/skeleton/issues/2391) feature

## Community Requests

The following requests have come from the community and are under consideration:
 
* n/a

## Bugs and Issues

* n/a

## References

- Native: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/progress
- Skeleton v2: https://www.skeleton.dev/components/progress-bars
- Shadecn: https://ui.shadcn.com/docs/components/progress
- Mantine: https://mantine.dev/core/progress/
- Material Design: https://m3.material.io/components/progress-indicators/overview
- Bits UI: https://www.bits-ui.com/docs/components/progress
- Melt UI: https://melt-ui.com/docs/builders/progress

## Feedback

If you have additional updates or requests for this feature, please do so in the comments section below.

PR #2556, opened 2024-03-15
Feat/progress

Hugos68:
Closes: #2390


packages/skeleton-svelte/src/lib/components/Progress/Progress.svelte
```diff
@@ -0,0 +1,49 @@
...
+		value,
+		min = 0,
+		max = 100,
+		// Root
+		base = 'overflow-x-hidden bg-surface-200-800',
+		height = 'h-2',
+		rounded = 'rounded',
+		classes = '',
+		// Indicator
+		indicatorBase = 'bg-primary-500 h-full',
```
Mahmoud-zino, line 14: same here for the background classes, extract it to it's own prop.
I would leave the height in base since there no other values in base and users will not lose anything if they decide to override it.

packages/skeleton-svelte/src/lib/components/Progress/Progress.svelte
```diff
@@ -0,0 +1,49 @@
...
+		indicatorRounded = 'rounded',
+		indicatorTransition = 'transition-[width]',
+		indicatorIndeterminateAnimation = 'indeterminate',
+		indicatorClasses = ''
+	}: ProgressProps = $props();
+
+	const indeterminate = $derived(value === undefined);
+	const fillPercentage = $derived(indeterminate ? 50 : ((value! - min) / (max - min)) * 100);
+	const width = $derived(`${fillPercentage}%`);
```
Mahmoud-zino, line 23: for reactive variables, it seems that we should prefix it with rx, see the docs https://next.skeleton.dev/resources/contribute/components#dynamic-style-props

packages/skeleton-svelte/src/lib/components/Progress/Progress.svelte
```diff
@@ -0,0 +1,49 @@
...
+</script>
+
+<!-- @component Track -->
+<div class="{base} {height} {rounded} {classes}">
+	<!-- @component Indicator -->
+	<div
+		class="{indicatorBase} {indicatorRounded} {indicatorTransition} {indeterminate
+			? indicatorIndeterminateAnimation
+			: ''} {indicatorClasses}"
```
Mahmoud-zino, line 32: since the ternary operator here has only 1 option, I would prefer to use this instead:
```suggestion
		class="{indicatorBase} {indicatorRounded} {indicatorTransition} {indeterminate && indicatorIndeterminateAnimation} {indicatorClasses}"
```
This is still readable and cleaner IMHO.

packages/skeleton-svelte/src/routes/components/progress/+page.svelte
```diff
@@ -0,0 +1,93 @@
...
+<h3 class="h3">Indeterminate</h3>
+
+<h4 class="h4">Default (indeterminate)</h4>
+
+<Progress {min} {max} />
+
+<h4 class="h4">custom-indeterminate</h4>
+
+<Progress {min} {max} indicatorIndeterminateAnimation="custom-indeterminate" />
+
```
Mahmoud-zino, line 80: Please add an example displaying how RTL is fully supported. for this component I think everything should be fine but we have to make sure.
look at the AppBar for examples, but basically just include a div with the dir="rtl" and you should see if it works as expected.

Mahmoud-zino (commented):
@Hugos68 Great job on this one, I have included some notes that I think should be applied, but please note I am not the expert here, so for every point check if it makes sense and apply it only if you think so, otherwise we can discuss it further :+1:

packages/skeleton-svelte/src/lib/components/Progress/types.ts
```diff
@@ -0,0 +1,30 @@
+export interface ProgressProps {
+	/** Set the value */
+	value?: number;
+	/** Set the minimum value. */
+	min?: number;
+	/** Set the maximum value. */
+	max?: number;
+
```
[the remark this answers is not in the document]
Hugos68: The progress NEXT ticket said to remove the labelledBy prop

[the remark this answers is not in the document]
endigo9740: @Hugos68 the comment was not to include a the label itself as part of the component. But `labelledBy` should exist! This indicates the external element that contains the label.

packages/skeleton-svelte/src/lib/components/Progress/Progress.svelte
Hugos68: I had this before, but this would create a `false` class, not what you want

Mahmoud-zino: > I had this before, but this would create a `false` class, not what you want

Are you sure? I tested it yesterday and it worked as expected!

packages/skeleton-svelte/src/lib/components/Progress/Progress.svelte
Hugos68: I think we both mistunderstood when to use `rx`, see: https://discord.com/channels/1003691521280856084/1010694301883318326/1219296077434585198 where Chris explains it more thoroughly

packages/skeleton-svelte/src/lib/components/Progress/Progress.svelte
Hugos68: It does work but when `indeterminate` is `false` it will literally add a `false` string to the classes.

packages/skeleton-svelte/src/lib/components/Progress/Progress.svelte
```diff
@@ -0,0 +1,49 @@
+<script lang="ts">
+	import type { ProgressProps } from './types.js';
+
+	let {
+		value,
+		min = 0,
+		max = 100,
+		// Root
+		base = 'overflow-x-hidden bg-surface-200-800',
```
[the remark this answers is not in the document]
Hugos68: Good call, done.

packages/skeleton-svelte/src/lib/components/Progress/types.ts
```diff
@@ -0,0 +1,30 @@
...
+	// Indicator ---
+	/** Provide base classes. */
+	indicatorBase?: string;
+	/** Provide rounded classes. */
+	indicatorRounded?: string;
+	/** Provide transition classes. */
+	indicatorTransition?: string;
+	/** Provide indeterminate animation classes. */
+	indicatorIndeterminateAnimation?: string;
+	/** Provide arbitrary classes. */
```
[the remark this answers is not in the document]
Hugos68: Done.

packages/skeleton-svelte/src/lib/components/Progress/Progress.svelte
```diff
@@ -0,0 +1,49 @@
...
+	}: ProgressProps = $props();
+
+	const indeterminate = $derived(value === undefined);
+	const fillPercentage = $derived(indeterminate ? 50 : ((value! - min) / (max - min)) * 100);
+	const width = $derived(`${fillPercentage}%`);
+</script>
+
+<!-- @component Track -->
+<div class="{base} {height} {rounded} {classes}">
+	<!-- @component Indicator -->
```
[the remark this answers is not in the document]
Hugos68: I deleted both right now, I'm not sure what to put there. I'll leave that upto Chris or You before it gets merged

packages/skeleton-svelte/src/lib/components/Progress/Progress.svelte
```diff
@@ -0,0 +1,49 @@
...
+		indicatorClasses = ''
+	}: ProgressProps = $props();
+
+	const indeterminate = $derived(value === undefined);
+	const fillPercentage = $derived(indeterminate ? 50 : ((value! - min) / (max - min)) * 100);
+	const width = $derived(`${fillPercentage}%`);
+</script>
+
+<!-- @component Track -->
+<div class="{base} {height} {rounded} {classes}">
```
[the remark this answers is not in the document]
Hugos68: Done.

packages/skeleton-svelte/src/lib/components/Progress/types.ts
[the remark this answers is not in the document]
Hugos68: Done.

packages/skeleton-svelte/src/lib/components/Progress/Progress.svelte
Hugos68: Done.

packages/skeleton-svelte/src/lib/components/Progress/Progress.svelte
Hugos68: Done.

packages/skeleton-svelte/src/lib/components/Progress/Progress.svelte
Hugos68: Done. Moved to an `rx` variable that's either the animation or an empty string.

packages/skeleton-react/src/components/Progress/Progress.tsx
```diff
@@ -0,0 +1,64 @@
...
+                @keyframes anim-progress-indeterminate {
+                    from {
+                        transform: translateX(-200%);
+                    }
+                    to {
+                        transform: translateX(200%);
+                    }
+                }
+            `}</style>
+        </>
```
Sarenor, line 62: I feel like the indentation here is wrong, but if that's what the formatter says it has to be then so it shall be

TazorDE: I don't think the formatter has been run yet. See line 34

TazorDE (commented):
looks good to me :)
nice work

packages/skeleton-react/src/components/Progress/Progress.tsx
```diff
@@ -0,0 +1,64 @@
...
+		meterClasses = ''
+	}: ProgressProps) {
+
+    useEffect(() => {
+        if (min > max) {
+            console.error(`The minimum value: "${min}" is greater than the maximum value: "${max}"`);
+        }
+    }, [min, max]);
+
+    const indeterminate = useMemo(() => value === undefined, [value]);
```
Sarenor, line 30: Cheap enough to not require useMemo probably

packages/skeleton-react/src/components/Progress/Progress.tsx
```diff
@@ -0,0 +1,64 @@
...
+    const fillPercentage = useMemo(() => (indeterminate ? 50 : ((value! - min) / (max - min)) * 100), [value, min, max, indeterminate]);
+    const width = useMemo(() =>`${fillPercentage}%`, [fillPercentage]);
```
Sarenor, line 32: Cheap enough to not require useMemo probably

packages/skeleton-react/src/components/Progress/Progress.tsx
```diff
@@ -0,0 +1,64 @@
...
+	}: ProgressProps) {
+
+    useEffect(() => {
+        if (min > max) {
+            console.error(`The minimum value: "${min}" is greater than the maximum value: "${max}"`);
+        }
+    }, [min, max]);
+
+    const indeterminate = useMemo(() => value === undefined, [value]);
+    const fillPercentage = useMemo(() => (indeterminate ? 50 : ((value! - min) / (max - min)) * 100), [value, min, max, indeterminate]);
```
Sarenor, line 31: Probably also pretty cheap but I think it's fine to leave it in. On the other hand, it's probably going to be the thing that changes the most anyways and will make memoization useless in most cases.

I would tend to leave it in but I'm not 100% convinced. Haven't spent enough time doing low level react optimizations like this.

Sarenor (dismissed):

endigo9740:
@Hugos68 similar to the Tabs review, I'll approach this from a few perspectives:

## Design

No issues with the design, everything matches my expectations! Great job!

## DX

FYI the native `<progress>` element does not allow a minimum value. I'm not sure it provides any meaningful benefit here since progress should always start a "none".

Screenshot 2024-03-19 at 4 05 39 PM

## RTL

You can implement simple RTL support using `classes="-scale-x-[100%]"` to mirror (flip) the element on the X-axis. I'd suggest doing this with a `rtl: boolean` prop. When activated append this to the parent element class list.

## Code

Look for my code review comments shortly after this message appears.

packages/skeleton-react/src/components/Progress/Progress.tsx
```diff
@@ -0,0 +1,68 @@
...
+	meterBase = "h-full",
+	meterBg = "bg-primary-500",
+	meterRounded = "rounded",
+	meterTransition = "transition-[width]",
+	meterIndeterminateAnim = "indeterminate",
+	meterClasses = "",
+}: ProgressProps) {
+	useEffect(() => {
+		if (min > max) {
+			console.error(
```
endigo9740, line 25: FYI this does not appear to be triggering for me on either the Svelte or React components.

For Svelte I tested:

1. Setting `value: 150` and `max: 100` in the preview page `$state()` rune
2. Setting the same settings via the inputs on the page
3. Manually passing the values via props

packages/skeleton-react/src/components/Progress/Progress.tsx
```diff
@@ -0,0 +1,68 @@
+import { useEffect } from "react";
+import { ProgressProps } from "./types";
+
+export default function Progress({
+	value,
```
endigo9740, line 5: I might suggest considering appending an optional `id` prop. This could be useful your console error to identify *which* component is erroring. Apply it to the parent element.

packages/skeleton-react/src/components/Progress/Progress.tsx
```diff
@@ -0,0 +1,68 @@
...
+		<>
+			<div
+				role="progressbar"
+				aria-labelledby={ariaLabelledby}
+				aria-valuenow={value}
+				aria-valuemin={min}
+				aria-valuemax={max}
+				className={`${base} ${bg} ${height} ${rounded} ${classes}`}
+			>
+				<div
```
endigo9740, line 47: I think the styled div is more than adequate here, but MDN has an interesting take. Using an SVG!
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/progressbar_role#examples

The benefit here is we could enable more complex visual styling. Such as using a Stroke Linecap like the Radial Progress bars:
https://www.skeleton.dev/components/progress-radials#styling
(change the option from `butt` at the bottom-right of the preview)

I'd only suggest this if you're feeling like an extra challenge!

packages/skeleton-react/src/components/Progress/types.ts
```diff
@@ -0,0 +1,36 @@
...
+	// Meter ---
+	/** Set meter base classes. */
+	meterBase?: string;
+	/** Set meter bg classes */
+	meterBg?: string;
+	/** Set meter rounded classes. */
+	meterRounded?: string;
+	/** Set meter transition classes. */
+	meterTransition?: string;
+	/** Set meter indeterminate animation classes */
```
endigo9740, line 32: I'd suggest a slightly longer description here where we indicate the inherent animation is embedded within the component. Not sure how we word that exactly though to be succinct.

packages/skeleton-react/src/components/Progress/Progress.tsx
```diff
@@ -0,0 +1,68 @@
+import { useEffect } from "react";
+import { ProgressProps } from "./types";
+
+export default function Progress({
```
endigo9740, line 4: We can add typing to indicate this is a React function component. See the Avatar component as reference

```ts
export const Avatar: React.FC<AvatarProps> = ()
```

Might be worth checking if this is better to use this explicitly or not.

packages/skeleton-svelte/src/lib/components/Progress/Progress.svelte
```diff
@@ -0,0 +1,60 @@
+<script lang="ts">
```
endigo9740, line 1: FYI the React component appeared higher in the the list when reviewing. Please make sure to echo all relevant changes to the Svelte side too please!

endigo9740 (changes requested):

packages/skeleton-react/src/components/Progress/Progress.tsx
Hugos68: Weird, using the playground and just entering any value where this condition would be true results in:
image

packages/skeleton-react/src/components/Progress/Progress.tsx
Hugos68: This feels like overkill since svelte already shows where you are using the component (file and linenumber)

packages/skeleton-react/src/components/Progress/Progress.tsx
Hugos68: This definitely seems interesting, I'll take a look once the components are pretty much done

packages/skeleton-svelte/src/lib/components/Progress/Progress.svelte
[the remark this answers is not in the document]
endigo9740: @Hugos68 @Mahmoud-zino here's the documentation around the built-in Svelte documentation features:
https://svelte.dev/docs/faq#how-do-i-document-my-components

endigo9740:
@Hugos68 one more note - just noticed a couple merge conflicts. Make sure those are sorted before the next PR review as well. Please and thanks!

packages/skeleton-react/src/components/Progress/types.ts
Hugos68: Added a longer name, could you take a look again?

packages/skeleton-react/src/components/Progress/Progress.tsx
Hugos68: Could you confirm this is working as intended?

endigo9740:
@Hugos68 I've completed my review and implemented a few changes.

See all changes in the commit here:
https://github.com/skeletonlabs/skeleton/pull/2556/commits/be24ed70d725bb9c062ac1f0b4962466ca5e004a

- Cleaned up the preview page styles (because I have OCD heh)
- Removed any references to `min`
- Setup RTL demos (see below regarding animations)
- Extended Tailwind to include `animate-indeterminate`via the Skeleton Tailwind plugin
- Renamed `meterIndeterminateAnim` to `meterAnimate` to follow Tailwind convention

The only issue I could not resolve was why the validation error for a value that exceeds a max is not working for me. But if it's working for you I'm not going to stress about it.

I'll follow up in a moment to resolve the merge conflicts.

endigo9740 (approved):

PR #2561, opened 2024-03-20
Chore: Restructure docs

Hugos68:
Changes:
1. All collections are now categories under the `docs` collection.
2. All categories are accompanied with a `meta.mdx` with metadata related to that category.
3. Added `src/content/config.ts` with `zod` schema's to create type safe `frontmatter`.
4. Added `src/content/docs.ts` with helpers to retrieve stuff like metadata from a category, all the categories, etc


Hugos68:
Added coming soon page for blog:
image

Just FYI @endigo9740 we should add a 404 page that for example links back to a familiar place.
See: https://docs.astro.build/en/basics/astro-pages/#custom-404-error-page

Hugos68:
I would also like to revisit the current Layout and frontmatter structure, it seems overly complex which was neccesary at the time but I doubt this is still neccesary

endigo9740:
@Hugos68 sounds good, please continue as you can and keep me in the loop when this is ready for the next review.

Thanks for adding the blog page, and per the 404 page, I've noted that on the Astro Docs ticket pinned on our Issue tracker. I agree, we need that soon as well.

endigo9740:
@Hugos68 this is is fairly important and one we should try to wrap asap. It's likely going to start impeding the progress of other PRs soon - such as our ability to add or update new doc sections, including those needed for new components. Let me know if I can help with anything here.

Issue #2363, opened 2023-12-26
NEXT: Automated Testing


endigo9740:
As part of Skeleton v3, we'll implement support for multiple frontend frameworks. We'll likely want to consolidate on a common and universal testing tool. For this we'll opt for Vitest.

## Tasks

- [x] Install and configure Vitest: https://vitest.dev/
- [x] Install and configure Testing Library: https://testing-library.com/docs/
- [x] Implement for React components library
- [x] Implement for Svelte components library

## Goal

See: https://github.com/skeletonlabs/skeleton/issues/150

We need to ensure we have an adequate depth and quality of test cases per component. This will make it easier to introduce new features, handle refactors, and of course improve general QA testing. With the ultimate goal of ensuring we have a set of simple and robust test cases that are not too rigid.

Some component design and features are still in flux, so aim to ensure that tests may need to be improved over time.

## References:

- [React + Vitest + Testing Library Guide](https://victorbruce82.medium.com/vitest-with-react-testing-library-in-react-created-with-vite-3552f0a9a19a)

## Feedback

If you have additional updates or requests for this feature, please do so in the comments section below.

Hugos68:
I'd think [Testing Library](https://testing-library.com/docs/) will be beneficial for cross framework component testing because the sheer amount of frameworks they support. This will allow for a familliar testing API no matter what framework is being tested on.

endigo9740:
Yeah that's been my though as well. We actually use Testing Library already. Though I really hate\ the name - it's so generic you can't google for anything you need! Grr! :D

endigo9740:
See also type testing: https://youtu.be/mWK3Y_1kmaM?si=64rMT3KiSCz7MQ0p&t=29214

endigo9740:
Svelte Testing Library maintainer @mcous has shared a super helpful audit of changes recommended when we return to Svelte component testing:

- https://github.com/skeletonlabs/skeleton/discussions/2891#discussioncomment-11127131

PR #2566, opened 2024-03-21
NEXT React Package Vitest and Testing Library Setup

endigo9740:
## Linked Issue

#2363

## Description

React Package Vitest and Testing Library Setup

### DIFF
--- a/packages/skeleton-react/package.json
+++ b/packages/skeleton-react/package.json
@@ -7,7 +7,8 @@
 		"build": "tsc && vite build",
 		"lint": "eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
 		"preview": "vite preview",
-		"test": "vitest"
+		"test": "vitest run",
+		"test:watch": "vitest"
 	},
 	"publishConfig": {
 		"types": "./dist/index.d.ts",
--- a/packages/skeleton-react/package.json
+++ b/packages/skeleton-react/package.json
@@ -6,7 +6,8 @@
 		"dev": "vite",
 		"build": "tsc && vite build",
 		"lint": "eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
-		"preview": "vite preview"
+		"preview": "vite preview",
+		"test": "vitest"
 	},
 	"publishConfig": {
 		"types": "./dist/index.d.ts",
@@ -48,6 +49,9 @@
 	},
 	"devDependencies": {
 		"@skeletonlabs/skeleton": "workspace:*",
+		"@testing-library/jest-dom": "^6.4.2",
+		"@testing-library/react": "^14.2.2",
+		"@testing-library/user-event": "^14.5.2",
 		"@types/react": "^18.2.66",
 		"@types/react-dom": "^18.2.22",
 		"@typescript-eslint/eslint-plugin": "^6.21.0",
@@ -57,12 +61,14 @@
 		"eslint": "^8.57.0",
 		"eslint-plugin-react-hooks": "^4.6.0",
 		"eslint-plugin-react-refresh": "^0.4.6",
+		"jsdom": "^24.0.0",
 		"postcss": "^8.4.35",
 		"react-router-dom": "^6.22.3",
 		"tailwindcss": "^3.4.1",
 		"typescript": "^5.4.2",
 		"vite": "^5.1.6",
 		"vite-plugin-remix-router": "^2.0.0",
-		"vite-plugin-tw-plugin-watcher": "workspace:*"
+		"vite-plugin-tw-plugin-watcher": "workspace:*",
+		"vitest": "^1.4.0"
 	}
 }
--- a/packages/skeleton-react/src/App.test.tsx
+++ b/packages/skeleton-react/src/App.test.tsx
@@ -1,0 +1,10 @@
+import { describe, expect, it } from "vitest";
+import { render } from "@testing-library/react";
+import App from "./App";
+
+describe("App", () => {
+  it("renders the component", () => {
+    const component = render(<App />);
+    expect(component).toBeTruthy();
+  });
+});
--- a/packages/skeleton-react/src/App.test.tsx
+++ b/packages/skeleton-react/src/App.test.tsx
@@ -4,7 +4,7 @@ import App from "./App";
 
 describe("App", () => {
   it("renders the component", () => {
-    const component = render(<App />);
-    expect(component).toBeTruthy();
+    const { getByTestId } = render(<App />);
+    expect(getByTestId("app")).toBeInTheDocument();
   });
 });
--- a/packages/skeleton-react/src/components/Accordion/Accordion.test.tsx
+++ b/packages/skeleton-react/src/components/Accordion/Accordion.test.tsx
@@ -2,9 +2,135 @@ import { describe, expect, it } from "vitest";
 import { render } from "@testing-library/react";
 import { Accordion } from "./Accordion";
 
-describe("Accordion", () => {
-  it("renders the component", () => {
+// Accordion ---
+
+describe("<Accordion>", () => {
+  it("should render the component", () => {
     const component = render(<Accordion />);
     expect(component).toBeTruthy();
   });
+
+  it("should render with `multiple` prop enabled", () => {
+    const component = render(<Accordion multiple />);
+    expect(component).toBeTruthy();
+  });
+
+  it("should allow for children", () => {
+    const value = "foobar";
+    const { getByTestId } = render(<Accordion>{value}</Accordion>);
+    expect(getByTestId("accordion").innerHTML).toContain(value);
+  });
+
+  it("should allow you to set the `base` style prop", () => {
+    const tailwindClasses = "bg-red-500";
+    const { getByTestId } = render(<Accordion base={tailwindClasses} />);
+    expect(getByTestId("accordion").classList).toContain(tailwindClasses);
+  });
+
+  it("should allow you to set the `classes` style prop", () => {
+    const tailwindClasses = "bg-green-500";
+    const { getByTestId } = render(<Accordion classes={tailwindClasses} />);
+    expect(getByTestId("accordion").classList).toContain(tailwindClasses);
+  });
+});
+
+// Accordion.Item ---
+
+describe("<Accordion.Item>", () => {
+  it("should render the component", () => {
+    const component = render(<Accordion.Item />);
+    expect(component).toBeTruthy();
+  });
+
+  it("should allow for children", () => {
+    const value = "foobar";
+    const { getByTestId } = render(<Accordion.Item>{value}</Accordion.Item>);
+    expect(getByTestId("accordion-item").innerHTML).toContain(value);
+  });
+
+  it("should allow you to set the `base` style prop", () => {
+    const tailwindClasses = "bg-red-500";
+    const { getByTestId } = render(<Accordion.Item base={tailwindClasses} />);
+    expect(getByTestId("accordion-item").classList).toContain(tailwindClasses);
+  });
+
+  it("should allow you to set the `classes` style prop", () => {
+    const tailwindClasses = "bg-green-500";
+    const { getByTestId } = render(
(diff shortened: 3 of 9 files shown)


endigo9740:
@Hugos68 @TazorDE you guys have some experience with testing so I thought I'd request a review if you get a chance. I still need to add and tweak a few tests. But let me know what you think overall.

packages/skeleton-react/package.json
```diff
@@ -6,7 +6,8 @@
 		"dev": "vite",
 		"build": "tsc && vite build",
 		"lint": "eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
-		"preview": "vite preview"
+		"preview": "vite preview",
+		"test": "vitest"
```
TazorDE, line 10: I suggest using the command `vitest run` if you plan on using this command in a CI environment. Just the `vitest` command starts the testing in watch mode which doesn't terminate a pipeline.

packages/skeleton-react/src/App.test.tsx
```diff
@@ -0,0 +1,10 @@
+import { describe, expect, it } from "vitest";
+import { render } from "@testing-library/react";
+import App from "./App";
+
+describe("App", () => {
+  it("renders the component", () => {
+    const component = render(<App />);
+    expect(component).toBeTruthy();
```
TazorDE, line 8: I know the example I shared with you didn't do this but using the `toBeInTheDocument()` asseartion appears to be a cleaner way to handle tests like this.
`toBeVisible()` is also a great assertion to use in such a usecase

packages/skeleton-react/src/components/Accordion/Accordion.test.tsx
```diff
@@ -0,0 +1,179 @@
+import { describe, expect, it } from "vitest";
+import { render, waitFor } from "@testing-library/react";
+
+import { Accordion, AccordionContext } from "./Accordion";
+
+// Accordion ---
+
```
TazorDE, line 7: As with the App.test.tsx file I recommend replacing `toBeTruthy()` with more specific tests

packages/skeleton-react/src/components/Accordion/Accordion.test.tsx
```diff
@@ -0,0 +1,179 @@
...
+  it("should allow for children", () => {
+    const value = "foobar";
+    const { getByTestId } = render(<Accordion>{value}</Accordion>);
+    expect(getByTestId("accordion").innerHTML).toContain(value);
+  });
+
+  it("should allow you to set the `base` style prop", () => {
+    const tailwindClasses = "bg-red-500";
+    const { getByTestId } = render(<Accordion base={tailwindClasses} />);
+    expect(getByTestId("accordion").classList).toContain(tailwindClasses);
```
TazorDE, line 28: you actually don't need the `.classList` approach to check if classes are applied you can use the `toHaveClass(tailwindClasses)` assertion

packages/skeleton-react/src/components/Accordion/Accordion.test.tsx
```diff
@@ -0,0 +1,179 @@
+import { describe, expect, it } from "vitest";
+import { render, waitFor } from "@testing-library/react";
+
```
TazorDE, line 3: You can also test the functionality of the accordion with the `act()` function similarly to this:
```ts
import { act } from '@testing-library/svelte';

// describe > it > ...
await act(async () => {
    // user selects the first input
    await userEvent.click(inputElements[0]);
    // user types a number
    await userEvent.type(inputElements[0], '1');
});
// expect something to have changed
```
I'd like to see a test like this

packages/skeleton-react/tests/setup.js
TazorDE, line 1: you might want to add these imports to the setup file to unlock additional functionality during testing. 
Although I'm not sure if the first one exists for react I can only reference a SvelteKit project atm.
```ts
import '@testing-library/react/vitest';
import '@testing-library/user-event';
```

packages/skeleton-react/vite.config.ts
```diff
@@ -1,19 +1,37 @@
-import { defineConfig } from "vite";
-import path from "node:path";
+import { defineConfig } from "vitest/config";
```
TazorDE, line 1: you actually don't need to change the import from 'vite' to 'vitest' since you are in the vite.config.ts file.
I'd prefer to use the 'vite' import if they update something important and 'vitest' lacks behind with updating their config

packages/skeleton-react/vite.config.ts
```diff
@@ -1,19 +1,37 @@
...
+        path.join("..", "..", "packages", "skeleton", "src", "plugin")
+      )
+    ),
+  ],
+  test: {
+    // https://victorbruce82.medium.com/vitest-with-react-testing-library-in-react-created-with-vite-3552f0a9a19a
+    environment: "jsdom",
+    globals: true,
+    setupFiles: "./tests/setup.js",
+    exclude: [
```
TazorDE, line 24: I'd recommend to add `...configDefaults.exclude` to this list. 
```ts
import { configDefaults } from 'vitest/config';
```

TazorDE (commented):
Mostly looking good but some infos I didn't share with you earlier will definitely improve testing.

packages/skeleton-react/src/components/Accordion/Accordion.test.tsx
endigo9740: This would use `fireEvent.click()` accordion to the Testing Library docs:
https://testing-library.com/docs/react-testing-library/example-intro#full-example

I actually had an example of this, but the way the component works in React clicking updates the Context API state. But nothing I tried would actually modify the DOM state related to the Context values. I spent half the afternoon on this alone heh!

Learning how to test React's Content Provider is completely new to me, so it's a bit of a black box. But I definitely want to implement something along these lines!

packages/skeleton-react/tests/setup.js
endigo9740: Can you provide more specifics around this - what it would provide?

packages/skeleton-react/vite.config.ts
endigo9740: Tell that to Typescript! Red squiggles until I modified this in the project.

packages/skeleton-react/tests/setup.js
TazorDE: the additional assertions I mentioned above like `toBeInDocument()` or `toBeVisible()` and the user event one is for when you want to simulate user behaviour, like clicking to open the accordion

packages/skeleton-react/src/components/Accordion/Accordion.test.tsx
TazorDE: If I remember and can find the time I'll have a look at it tomorrow too. Maybe an additional pair of eyes can help

packages/skeleton-react/vite.config.ts
TazorDE: Interesting... I'll look at it tomorrow alongside the userEvent stuff

TazorDE:
@endigo9740 Ok I figured out how to add user interaction to your tests and do all that configuration stuff. 
- Add `@testing-library/jest-dom` to the `types` field in the `tsconfig.json`. This provides intellisense-hints for the testing-library DOM assertions.
- Add the following imports to `/tests/setup.js` (this one could also be a .ts file but I didn't check that). They add user events and more assertions.
```ts
import '@testing-library/react';
import '@testing-library/user-event';
```
- Update `vite.config.ts`. No major updates but it now uses the `vite` import over the `vitest/config` import
```ts
import { defineConfig } from "vite";
import { configDefaults } from "vitest/config";
import react from "@vitejs/plugin-react";
import RemixRouter from "vite-plugin-remix-router";
import skeletonPluginWatcher from "vite-plugin-tw-plugin-watcher";
import path from "node:path";

// https://vitejs.dev/config/
export default defineConfig({
  plugins: [
    react(),
    // https://vite-remix-router.vercel.app/
    RemixRouter(),
    skeletonPluginWatcher(
      path.resolve(
        path.join("..", "..", "packages", "skeleton", "src", "plugin")
      )
    ),
  ],
  test: {
    // https://victorbruce82.medium.com/vitest-with-react-testing-library-in-react-created-with-vite-3552f0a9a19a
    environment: 'jsdom',
    globals: true,
    setupFiles: ['./tests/setup.js'],
    exclude: [
      ...configDefaults.exclude,
      "**/node_modules/**",
      "**/tests/**",
      "**/build/**",
      "**/coverage/**",
      "**/src/lib/icons/**",
      "**/.eslintrc.cjs",
      "**/vite.config.ts",
      "**/tailwind.config.ts",
      "**/postcss.config.cjs",
      "**/playwright.config.ts"
    ]
  }
});
```
- Add user interaction tests to your test file. The following is an example for the Accordion component.
```ts
// packages/skeleton-react/src/components/Accordion/Accordion.test.tsx
import { act, render, waitFor } from "@testing-library/react";
import userEvent from '@testing-library/user-event';
// ...
// ...
// ...
describe("Accordion usage", () => {
  it('should show the panel when the control is clicked', async () => {

    const { queryByText, getByText } = render(
      <Accordion>
        <Accordion.Item>
          <Accordion.Control controls="testControl">Test Control 1</Accordion.Control>
          <Accordion.Panel id="testControl">Test Panel 1</Accordion.Panel>
        </Accordion.Item>
        <Accordion.Item>
          <Accordion.Control controls="testControl2">Test Control 2</Accordion.Control>
          <Accordion.Panel id="testControl2">Test Panel 2</Accordion.Panel>
        </Accordion.Item>
      </Accordion>
    );

    // expect controls to be visible
    const control1 = getByText('Test Control 1');
    expect(control1).toBeInTheDocument();
    const control2 = getByText('Test Control 2');
    expect(control2).toBeInTheDocument();


    // expect panels to be hidden
    expect(queryByText('Test Panel 1')).not.toBeInTheDocument();
    expect(queryByText('Test Panel 2')).not.toBeInTheDocument();

    // click the first control
    await act(async () => {
      await userEvent.click(control1);
    });

    // expect first panel to be visible
    expect(queryByText('Test Panel 1')).toBeInTheDocument();
    // the second panel should still be hidden
    expect(queryByText('Test Panel 2')).not.toBeInTheDocument();

    // click the second control
    await act(async () => {
      await userEvent.click(control2);
    });

    // expect second panel to be visible
    expect(queryByText('Test Panel 2')).toBeInTheDocument();
    // the first panel should be hidden
    expect(queryByText('Test Panel 1')).not.toBeInTheDocument();
  });
});
´´´

packages/skeleton-react/src/App.test.tsx
endigo9740: @TazorDE I'm down to implement these, unfortunately neither assertions appear to exist. Can you confirm how these should be setup and used?

packages/skeleton-react/src/components/Accordion/Accordion.test.tsx
endigo9740: See: https://github.com/skeletonlabs/skeleton/pull/2566#discussion_r1539983435

packages/skeleton-react/src/components/Accordion/Accordion.test.tsx
endigo9740: See: https://github.com/skeletonlabs/skeleton/pull/2566#discussion_r1539983435

packages/skeleton-react/vite.config.ts
endigo9740: @TazorDE just a reminder about this issue! When I revert the change to default settings the `defineConig.test` is not supported. The tests functionally run, but there appears to be a Type error.

endigo9740:
@TazorDE just FYI if you got extra pings for comments that no longer exist. Your final comment above resolved a number of issues both mentioned here and on Discord. Sorry for not thinking to scan head. I tend to go through in a linear order top-to-bottom and work on issues one-by-one. I'll try to be more careful about this in the future!

TazorDE (approved):
lgtm

PR #2583, opened 2024-03-31
NEXT React Tabs component

### LINKED ISSUE #2395: NEXT: Tabs

Mahmoud-zino:
## Linked Issue

Closes #2395

## Description

Tabs React component


Mahmoud-zino:
@endigo9740 notable changes:
The current implementation of reactCompose utility is not accepting required fields, I had to add `<any>` to make it work, I searched for other alternatives but didn't find any easier way to avoid using any.

Also I will update you on the rest of the tasks tomorrow.

endigo9740:
@Hugos68 do you review the issue Mahmoud described above for the `reactCompose` feature and provide your recommendations. No rush, just when you can.

Hugos68:
We can remove `reactCompose` all together, it's merely a wrapper for `Object.assign`, `Object.assign` is far more leanient aswell so will probably happily accept `any`.

endigo9740:
@Mahmoud-zino given you mention work on this will continue onto the weekend I've bumped this back to the "draft" state.

endigo9740:
@Mahmoud-zino unfortunately I've run out of time for a full review for this today, so I'll plan to circle back tomorrow. In the meantime, here's a few superficial bits of feedback:

- [ ] I'd suggest trying to match the number of examples you had on the Svelte/React sandbox apps, including RTL!
- [ ] Some of the panel content is a bit too long and wraps. Maybe keep it short [like Mantine](https://mantine.dev/core/tabs/)?
- [ ] I'd suggest adding one example with focusable content within the panel to test a11y. Even an `<a>` would work!
- [ ] On the Svelte Docs the `T` for "Television" is lowercase in the first example.
- [ ] In the icon with text + icons stacked, put the icons on top. This is how Material UI handles it.
- [ ] The Svelte docs > Layout snippets don't have syntax highlighting. Verify the language is set to `svelte`

I'll do a proper dive into the code tomorrow! Overall things are looking great!

sites/next.skeleton.dev/src/content/docs/components/tab/meta.mdx
```diff
@@ -0,0 +1,8 @@
+---
+title: Tab
```
endigo9740, line 2: The page section should be named "Tabs" (plural) instead of "Tab" singular.

This is one of those odd situations in English where the single group entity is typically referred to in plural. This is because it's a group of things. Feel free to compare to a few other libraries to see what I mean.

Make sure to update the meta info, directory (for the route, etc

sites/next.skeleton.dev/src/examples/components/tabs/VerticalIconsExample.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { Tabs, TabsList, TabsControl } from '@skeletonlabs/skeleton-react';
+import { LibraryBig, Film, Tv } from 'lucide-react';
+
+export const Page: React.FC<any> = () => {
+	const [group, setGroup] = useState('books');
+	return (
+		<Tabs>
+			<Tabs.List>
+				<TabsList.Control name="books" group={group} onchange={setGroup} title="Books">
+					<TabsControl.Item classes="flex-col">
```
endigo9740, line 11: Each `TableControl.Item` needs to use`rounded="rounded-container"` to properly handle the hover state. The default `rounded` class is much too rounded.

Before:

image

After:

Screenshot 2024-04-09 at 2 26 10 PM

sites/next.skeleton.dev/src/content/docs/components/tab/react.mdx
```diff
@@ -0,0 +1,88 @@
...
+The tabs component complies with accessibility guidelines, ensuring an optimal user experience for everyone.
+
+<Preview client:load>
+	<Fragment slot="preview">
+		<FocusableContentExample client:load />
+	</Fragment>
+	<Fragment slot="code">
+		<Code code={FocusableContentExampleRaw} lang="tsx" />
+	</Fragment>
+</Preview>
```
endigo9740, line 88: I'd suggest adding a RTL examples similar to the text sandbox.

sites/next.skeleton.dev/src/content/docs/components/tab/react.mdx
```diff
@@ -0,0 +1,88 @@
...
+<Preview client:load>
+	<Fragment slot="preview">
+		<JustifyExample client:load />
+	</Fragment>
+	<Fragment slot="code">
+		<Code code={JustifyExampleRaw} lang="tsx" />
+	</Fragment>
+</Preview>
+
+## A11y
```
endigo9740, line 77: Rename to "Accessibility"

sites/next.skeleton.dev/src/content/docs/components/tab/react.mdx
```diff
@@ -0,0 +1,88 @@
...
+
+The tabs component complies with accessibility guidelines, ensuring an optimal user experience for everyone.
```
endigo9740, line 79: Suggested Text: "This component adheres to the ARIA APG. Use the preview below to test tabbing and keyboard navigation"

sites/next.skeleton.dev/src/content/docs/components/tab/react.mdx
```diff
@@ -0,0 +1,88 @@
...
+<Preview client:load>
+	<Fragment slot="preview">
+		<Example client:load />
+	</Fragment>
+	<Fragment slot="code">
+		<Code code={ExampleRaw} lang="tsx" />
+	</Fragment>
+</Preview>
+
+## Layout
```
endigo9740, line 31: Rename from "Layout" -> "Icons". The text below can remain as is.

sites/next.skeleton.dev/src/content/docs/components/tab/react.mdx
```diff
@@ -0,0 +1,88 @@
+---
+layout: '@layouts/LayoutDoc.astro'
+---
+
+export const components = componentSet;
+
+import { Page as Example } from '@examples/components/tabs/Example.tsx';
+import ExampleRaw from '@examples/components/tabs/Example.tsx?raw';
+import { Page as IconsExample } from '@examples/components/tabs/IconsExample.tsx';
+import IconsExampleRaw from '@examples/components/tabs/IconsExample.tsx?raw';
```
endigo9740, line 10: For all other components the example components are named "ExampleFoo" - with the word "example" at the start. I suggest we match that convention.

sites/next.skeleton.dev/src/content/docs/components/tab/react.mdx
```diff
@@ -0,0 +1,88 @@
...
+		<VerticalIconsExample client:load />
+	</Fragment>
+	<Fragment slot="code">
+		<Code code={VerticalIconsExampleRaw} lang="tsx" />
+	</Fragment>
+</Preview>
+
+## Stretch
+
+Tabs can be stretched to fill the available space.
```
endigo9740, line 55: Append at the end: "...using the `x` prop." (whichever prop is relevant here)

packages/skeleton-react/src/components/Tab/Tabs.tsx
```diff
@@ -0,0 +1,216 @@
...
+    children
+}) => {
+    return value === group && children && (
+        <div id={id} role="tabpanel" tabIndex={0} aria-labelledby={labelledBy} className={classes}>
+            {children}
+        </div> 
+    );
+};
+
+export const TabsControl = Object.assign(TabsControlRoot, { Item: TabsControlItem });
```
endigo9740, line 209: I'm not sure we're following the dot notation convention properly.

- The root should be `<Tabs>`
- All children should be prefixed `<Tabs.Foo>`

You can see this in practice for Mantine:
https://mantine.dev/core/tabs/

Instead we have multiple prefixes, including:

`<Tabs.Foo>`, `<TabsList.Foo>`, and `<TabsControl.foo>`

Which results in:

```ts
import { Tabs, TabsList, TabsControl } from "...";
```

There should only be a single import item per tree of components components. Something like:

```ts
import { Tabs } from "...";
```
```html
<Tabs>
    <Tabs.List>
        <Tabs.Control>(labelText)</Tabs.Control>
        <Tabs.Control>(labelText)</Tabs.Control>
        <Tabs.Control>(labelText)</Tabs.Control>
    </Tabs.List>
    <Tabs.Panel>(panelContent)</Tabs.Panel>
    <Tabs.Panel>(panelContent)</Tabs.Panel>
    <Tabs.Panel>(panelContent)</Tabs.Panel>
</Tabs>
```

Additionally, the `<TabsControl.Item>` seems like a redundant abstraction. I'm assuming you were trying to replicate the structure for icon + label (like Svelte). However, icons should be passed via a prop. See the AccordionItem for reference.

endigo9740: Feel free to ping me if we need to talk this one out! Given the scope of changes here I'll pause my review until these changes are implemented.

endigo9740 (changes requested):
@Mahmoud-zino here's another round of changes. I might suggest you start with the structural change for the React components dot notation syntax. That'll have a dramatic effect on the code, so I'll pause my review and pick back up once that's in place.

packages/skeleton-react/src/components/Tab/Tabs.tsx
```diff
@@ -0,0 +1,216 @@
...
+    background = '',
+    border = 'border-b-[1px]',
+    text = 'type-scale-3',
+    padding = 'pb-2',
+    rounded = '',
+    gap = 'gap-1',
+    cursor = 'cursor-pointer',
+    classes = '',
+    // Events
+    onclick = () => {},
```
endigo9740, line 60: @Hugos68 caught a convention issue here. It looks like events differ between Svelte 5 and React:

* Svelte: `onclick` -> all lowercase
* React: `onClick` -> camel-case

Hugo made the suggest that we lean into the framework convention and I think I agree here. We're going to move this to approach for the Accordion and update the contribution docs with a note for this asap.

Make sure to update all events here - click, keydown, etc. Note this change is ONLY needed for the React components. Svelte should remain as is!

endigo9740 (changes requested):

sites/next.skeleton.dev/src/examples/components/tabs/VerticalIconsExample.tsx
Mahmoud-zino: sorry it looks like I have changed it only for svelte on your initial request, I will include it for react too.

sites/next.skeleton.dev/src/content/docs/components/tab/react.mdx
Mahmoud-zino: @endigo9740 we are using classes to control the width of the tab control, so I am not sure what to write here...

endigo9740: @Mahmoud-zino seem like we would want to implement a `width` prop, right? This illustrates a use case for it.

endigo9740:
@Mahmoud-zino no further changes here, everything looks great. You nailed it!

The only thing we might consider in the future is avoiding loose text within children slots. For example:

```html
<SomeComponent>Skeleton</SomeComponent>
```
```html
<SomeComponent>
    <SomeIcon />
    Skeleton
</SomeComponent>
```

See where the text "Skeleton" is implemented above. This can incur two issues:

1.  Loose text isn't treated as block-like element, which can have some adverse affects when it comes to mixing text, icons, etc for flexbox or grid styles.
2. There's no guarantee the text will adhere to global styles. So if the text is intended to be a heading, we need to declare it as such.

But none of the issues appear to be present here - just wanted to give a heads up going forward!

PR #2599, opened 2024-04-08
Next: Add Ratings Component

### LINKED ISSUE #2392: NEXT: Ratings

## Maintainer Requests

- [x] Should adhere to the [ARIA APG](https://www.w3.org/WAI/ARIA/apg/patterns/radio/examples/radio-rating/)
- [x] In the v2 version we had a "wobble" due to the position of the Font Awesome star icons. Lucide should not suffer from this, but let's pay close attention to ensure this is not the case in v3.
- [x] Recommend a smaller display size by default in the docs. The icons were way too big in v2!
- [x] (optional) [Fractional support](https://mantine.dev/core/rating/#fractions) is interesting, but not required

## Community Requests

* https://github.com/skeletonlabs/skeleton/issues/2558

## Bugs and Issues

* n/a

## References

- Skeleton v2: https://www.skeleton.dev/components/ratings
- Mantine: https://mantine.dev/core/rating/

## Feedback

Mahmoud-zino:
## Linked Issue

Closes #2392


endigo9740:
@Mahmoud-zino just took a quick look at this. I won't provide a full review as I know it's still a draft - but make sure you're not hardcoding SVGs within components if it can be avoid. While we're taking an opinionated stance on Lucide, that doesn't mean we're keen to embed icons from it.

Mahmoud-zino:
> @Mahmoud-zino just took a quick look at this. I won't provide a full review as I know it's still a draft - but make sure you're not hardcoding SVGs within components if it can be avoid. While we're taking an opinionated stance on Lucide, that doesn't mean we're keen to embed icons from it.

This is just a fallback svg (star since it is the most common one for rating) but if you want to use another approach I am totally ok with it, just let me know what would you prefer.

endigo9740:
Yeah, no fallback SVG. Users will be required to add their own icons. Placeholder text is probably fine though.

Mahmoud-zino:
@endigo9740 Please ignore the fallback SVG at the moment, I want to get your opinion on the fractional approach before continuing the implementation.
so this approach is using clip-path to hide parts of the full icon to display the empty icon under it, this will enable us to support fractions fully with ease, you can look at the example svelte file to see it in action...
please let me know if I should continue this approach or you have something else in mind

endigo9740:
@Mahmoud-zino `clip-path` is a great way to handle this! I like it. Let's do it!

endigo9740:
@Mahmoud-zino as part of prepping for tomorrow's release I'm doing a pass through all pending PRs. Let me know a status update on this one. What's remaining and ETA for delivery. If you're bogged down, let me or the team know and someone can likely jump in and pick this up! (Hugo, etc)

Mahmoud-zino:
@endigo9740 Functionality and design of the component is now implemented and ready for review :+1:

endigo9740:
@Mahmoud-zino I'm so sorry, I've been swamped with the Floating UI Svelte stuff this week and completely missed this.

I'll aim to review in the next day or two, could you go ahead and do me a huge favor - could you install Lucide for the Svelte and React packages as dev dependencies. Then go ahead and swap out SVGs with icons from those?

Remember, we will NEVER used Lucide in our components, but it's perfectly valid to use in the SvelteKit and Vite/React sandboxes. I've been meaning to do this for a while, and it'll be helpful for your component here.

Please and thanks sir!

endigo9740:
Thanks @Mahmoud-zino this is on my agenda to review today, assuming no further interruptions!

packages/skeleton-svelte/src/lib/components/Rating/Rating.svelte
```diff
@@ -0,0 +1,125 @@
...
+			<div class="clip-right absolute left-0 top-0 w-fit" style="--clip_value: {100 - percentage * 100}%">
+				{#if fullIcon}
+					{@render fullIcon()}
+				{/if}
+			</div>
+		</div>
+	{/if}
+{/snippet}
+
+<div
```
endigo9740, line 104: Let's make the wrapping element a `<figure>` instead of div.

packages/skeleton-svelte/src/lib/components/Rating/Rating.svelte
```diff
@@ -0,0 +1,125 @@
...
+	role="group"
```
endigo9740, line 105: Let's drop the `role` here. The rule of thumb with a11y is no-a11y attributes are better than bad or incorrect a11y attributes. This will become a Figure role per the above comment.

packages/skeleton-svelte/src/lib/components/Rating/Rating.svelte
```diff
@@ -0,0 +1,125 @@
...
+	class="{base} {text} {fill} {justify} {spaceX} {classes}"
+	data-testid="rating"
+	onmouseleave={onRatingLeave}
+	onblur={onRatingLeave}
+>
+	{#each { length: max } as _, i}
+		<div class="{itemBase} {itemAspect} {itemClasses}">
```
endigo9740, line 112: I'm not sure I understand the extra wrapping `div` here. Overall structure should be:

```html
<figure>
    <button /> (or div)
    <!-- repeat per instance -->
</figure>
```

packages/skeleton-svelte/src/lib/components/Rating/Rating.svelte
```diff
@@ -0,0 +1,125 @@
...
+			</div>
+		</div>
+	{/if}
+{/snippet}
+
+<div
+	role="group"
+	class="{base} {text} {fill} {justify} {spaceX} {classes}"
+	data-testid="rating"
+	onmouseleave={onRatingLeave}
```
endigo9740, line 108: Not sure I understand the extra events on the parent. Neither `<div>` or `<figure>` are directly interactive, which is why I'm assuming a role was given. But this is typically an indicator things are not setup properly. Keep interaction limited to interactive elements only.

For example, when I hover or click a button, handle the interaction there only.

packages/skeleton-svelte/src/lib/components/Rating/types.ts
```diff
@@ -0,0 +1,40 @@
...
+	/** Sets the item base styles. */
+	itemBase?: string;
+	/** Sets the item aspect ratio styles. */
+	itemAspect?: string;
+	/** Provide arbitrary CSS classes to the rating item. */
+	itemClasses?: string;
+
+	// Snippets ---
+	/** The empty icon slot. */
+	emptyIcon?: Snippet;
```
endigo9740, line 37: It's easier to identify related snippets with prefixes versus suffixes. At least in LTR languages! :)

- `iconEmpty`
- `iconFull`

packages/skeleton-svelte/src/lib/components/Rating/Rating.svelte
```diff
@@ -0,0 +1,125 @@
...
+		hoverValue = order + lastSelectedFraction / fraction;
+	}
+
+	function onRatingLeave() {
+		hovering = false;
+	}
+</script>
+
+{#snippet rating(order, percentage)}
+	{#if interactive}
```
endigo9740, line 65: There's a lot of boilerplate being replicated between interactive vs non-interactive.

I'm trying to think of a way to handle this. Maybe look into `<svelte:element>`, which lets you conditionally set the HTML tag to use:

https://svelte.dev/docs/special-elements#svelte-element

The issue would be dynamically applying the attributes and event handlers. Maybe you could use a spread operator like:

```html
<svelte:element bind:this={elemTag} {...attributes}>...</ svelte:element>
```

packages/skeleton-svelte/src/lib/components/Rating/Rating.svelte
```diff
@@ -0,0 +1,125 @@
...
+	function onRatingHover(event: Event, order: number) {
+		// handling mouse
+		if (event instanceof MouseEvent) {
+			const ratingRect = (event.currentTarget as HTMLElement).getBoundingClientRect();
+			const fractionWidth = ratingRect.width / fraction;
+			const left = event.clientX - ratingRect.left;
+			lastSelectedFraction = Math.floor(left / fractionWidth) + 1;
+		}
+		// handling keyboard
+		else {
```
endigo9740, line 53: Looks like some logic is missing from the `else` condition?

packages/skeleton-svelte/src/lib/components/Rating/Rating.svelte
```diff
@@ -0,0 +1,125 @@
...
+
+	let hoverValue = $state(0);
+	let hovering = $state(false);
+	let lastSelectedFraction = $state(1);
+
+	function onRatingClick(event: Event, order: number) {
+		// handling mouse
+		if (event instanceof PointerEvent) {
+			const ratingRect = (event.currentTarget as HTMLElement).getBoundingClientRect();
+			console.log(event);
```
endigo9740, line 33: Console log was committed

packages/skeleton-svelte/src/lib/components/Rating/Rating.svelte
```diff
@@ -0,0 +1,125 @@
...
+		// Root
+		base = 'flex w-full',
+		text = 'text-token',
+		fill = 'fill-token',
+		justify = 'justify-center',
+		spaceX = 'space-x-2',
+		classes = '',
+		// Item ---
+		itemBase = 'w-full',
+		itemAspect = 'aspect-[1/1]',
```
endigo9740, line 18: The `aspect-square` Tailwind utility class is equivalent to 1:1

packages/skeleton-svelte/src/lib/components/Rating/Rating.svelte
```diff
@@ -0,0 +1,125 @@
...
+	{#if interactive}
+		<button
+			type="button"
+			class="relative h-full w-full"
+			onclick={(event) => onRatingClick(event, order)}
+			onmousemove={(event) => onRatingHover(event, order)}
+			onmousedown={(event) => onRatingHover(event, order)}
+			onfocus={(event) => onRatingHover(event, order)}
+		>
+			<div class="clip-left absolute left-0 top-0 flex h-full w-full items-center justify-center" style="--clip_value: {percentage * 100}%">
```
endigo9740, line 74: There's a ton of inline classes that should likely be abstracted to `baseX` style props.

packages/skeleton-svelte/src/lib/components/Rating/types.ts
```diff
@@ -0,0 +1,40 @@
...
+	/** Sets horizontal spacing styles. */
+	spaceX?: string;
+	/** Provide arbitrary CSS classes. */
+	classes?: string;
+
+	// Item ---
+	/** Sets the item base styles. */
+	itemBase?: string;
+	/** Sets the item aspect ratio styles. */
+	itemAspect?: string;
```
endigo9740, line 31: To just to confirm, icon size is set externally, correct? We don't need an `itemSize` attribute, is that right?

https://tailwindcss.com/docs/size

packages/skeleton-svelte/src/routes/components/ratings/+page.svelte
```diff
@@ -0,0 +1,147 @@
...
+						<Bone size={24} />
+					{/snippet}
+				</Rating>
+			</div>
+			<h3 class="h3">{interractiveValue.toFixed(2)}</h3>
+		</div>
+	</section>
+
+	<section class="space-y-4">
+		<h2 class="h2">Interractive</h2>
```
endigo9740, line 102: Typo: "Interactive"

endigo9740 (changes requested):
@Mahmoud-zino here's my review. A few things of note:

- I LOVE the fractional units. That's so much cleaner.
- I'm kind of so-so on the hover states, that may be overkill and adding extra complexity
- Keep the interaction though. Clicking to set state is nice.

Let me know if I can clarify any points raised below. I'd suggest we go through another round of review before proceeding to tests and React components.

packages/skeleton-svelte/src/lib/components/Rating/Rating.svelte
Mahmoud-zino: We agreed last time to handle keyboard logic after the first review, and I have some questions about how are we going to handle fractions with Keyboard, so I would wait until the first wave of changes are approved :+1:

packages/skeleton-svelte/src/lib/components/Rating/Rating.svelte
Mahmoud-zino: I don't think it makes sense to abstract those classes, they are crucial for the functionality of the component and should not be changed by the user IMHO.
And I am not sure what to call the element if abstracted, maybe you have a suggestion...

endigo9740:
@Mahmoud-zino I've bumped this back to a draft to help better indicate when it's ready for review again!

Mahmoud-zino:
@endigo9740 I have left you a couple of comments back that needs some clarifications :smile: Thanks.
Also about the hover, I removed it but I am really thinking we should keep it, it is a lot easier to select fractions when you are getting immediate feedback.

> If we can reduce complexity enough, I'd like us to consider inserting input state so this can be used in forms!

Do you mean adding a prop so users can change the type of the buttons to submit a form? if not, could you please provide more details?

endigo9740:
@Mahmoud-zino this is still on my list to review. Just tied up with the auto-doc stuff atm.

> Do you mean adding a prop so users can change the type of the buttons to submit a form? if not, could you please provide more details?

I believe buttons and inputs cannot have their type dynamically set. Svelte will complain about this. In this case I think we should rely on users to handle submit themselves and keep the buttons at `type="button"`

packages/skeleton-svelte/src/lib/components/Rating/Rating.svelte
endigo9740: The problem her is we can't promote the library as having faux-headless support if some classes are hardcoded. Even if they're structurally important, consider moving them to a `base` prop of some sort.

packages/skeleton-svelte/src/lib/components/Rating/Rating.svelte
Mahmoud-zino: @endigo9740 Do you have any suggestions on how to handle the keyboard events with fractions? do we ignore the fractions or use something like arrow to set the different fractions?

endigo9740: @Mahmoud-zino I like how Mantine handles this. Provide focus, then use left/right arrows:
https://mantine.dev/core/rating/#fractions

endigo9740:
@Mahmoud-zino hopefully you don't murder me for taking the initiative here, but I've completed a PR review and a bit of a refactor to the component. This streamlines a lot of the complexity and uses much less code for the same results.

Notable changes include:

- The local embedded snippet is gone; code has been moved within the each block
- The `<svelte:element>` in gone in favor of a hardcoded button
- The button hover state is set by `pointer-events-none` when `interactive` is false.
- I've refactored the styles to confirm to the base/dynamic/classes pattern ([see dynamic reference](https://next.skeleton.dev/docs/resources/contribute/components#dynamic-style-props)). There should be less redundancy.
- I've change the semantics for a few of the prop names
- Fixed an issue where snippets names internally/externally did not match

My hope here is that by slimming down the complexity, it should make it easier to focus on adding keyboard interactions. That would be the final piece before this ready to create documentation and port to React.

Let me know what you think!

Mahmoud-zino:
@endigo9740 I just reviewed your changes and I am really happy with what you did :+1: 
also, I just implemented the keyboard interaction part, and here are some points to notice about that:
1. I had to change the `onclick` event to `onmousedown` event so pressing enter won't miss with the calculation done on the mouse position when clicking for fractions.
2. for some reason I couldn't get the tabIndex to update re-actively so I've gone the ref route here.

Let me know what you think about the changes and if you wish to go another route for the keyboard interaction.

endigo9740:
@Mahmoud-zino sounds good. I'd say go ahead and setup a doc page for this and that'll make it easier for me to preview over the weekend (if you've got time!). Otherwise I'll review on Monday.

endigo9740:
@Mahmoud-zino thank you for adding the doc page. Things are looking great! With a quick glance, I'm good with how you're handling the keyboard events. The only suggestion I might have would be to default to `ltr` instead of `rtl`. The majority of the project, maintainers, and likely contributors will be able to make more sense of that.

endigo9740:
@Mahmoud-zino I saw this was still marked "ready for review" so I've bumped this back to the "draft" state after a quick review. I modified a couple of the doc pages, but otherwise everything in the component looks fine. You're good to roll with the React version asap!

endigo9740:
@Mahmoud-zino just saw this marked as ready for review. I'll aim for a PR review first thing tomorrow. Thanks!

packages/skeleton-svelte/src/lib/components/Rating/Rating.svelte
```diff
@@ -0,0 +1,134 @@
...
+
+	let {
+		value = $bindable(0),
+		max = 5,
+		interactive = false,
+		fraction = 1,
+		// Root
+		base = 'flex',
+		width = 'w-full',
+		text = 'text-token',
```
endigo9740, line 12: The `-token` classes no longer exist in v3. We should also avoid hardcoding values whenever possible.

If you were going automatically set these I'd opt for something like:

> Current means inherit from the nearest parent, aka "use the default"

```ts
text = 'text-current',
fill = 'fill-current',
```

Optionally, when providing set values, always use a pairing like: `fill-surface-950-50`.

However, users are supplying their own icons, so perhaps they should be responsible for these values? If so, drop these two props. Honestly this would be my recommendation. Less is more.

sites/next.skeleton.dev/src/examples/components/ratings/Example.svelte
```diff
@@ -0,0 +1,13 @@
+<script lang="ts">
+    import { Rating } from '@skeletonlabs/skeleton-svelte';
+	import Star from 'lucide-svelte/icons/star';
+</script>
+
+<Rating value={3} width="w-44">
+    {#snippet iconFull()}
+        <Star size={24} fill="white" />
```
endigo9740, line 8: Remove the hardcoded white fill. Either use the default (per my other comment) or use a color pairing like `class="fill-surface-950-50"`. Currently this is impossible to see in light mode:

Screenshot 2024-06-11 at 2 06 13 PM

Screenshot 2024-06-11 at 2 06 19 PM

sites/next.skeleton.dev/src/examples/components/ratings/Example.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { Star } from 'lucide-react';
+
+export const Page: React.FC<any> = () => {
+	return (
+		<Rating value={3} width="w-44">
+			<Rating.IconEmpty>
+				<Star size={24} />
+			</Rating.IconEmpty>
+			<Rating.IconFull>
+				<Star size={24} fill="white" />
```
endigo9740, line 12: Another hardcoded fill, switch to: `class="fill-surface-950-50"`

sites/next.skeleton.dev/src/examples/components/ratings/FractionExample.svelte
```diff
@@ -0,0 +1,13 @@
+<script lang="ts">
+    import { Rating } from '@skeletonlabs/skeleton-svelte';
+	import Star from 'lucide-svelte/icons/star';
+</script>
+
+<Rating value={2.5} fraction={2} width="w-44" interactive>
+    {#snippet iconFull()}
+        <Star size={24} fill="white" />
```
endigo9740, line 8: Another hardcoded fill, switch to: `class="fill-surface-950-50"`

sites/next.skeleton.dev/src/examples/components/ratings/FractionExample.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+export const Page: React.FC<any> = () => {
+	const [value, setValue] = useState(2.5);
+	return (
+		<Rating value={value} onValueChange={setValue} fraction={2} width="w-44" interactive>
+			<Rating.IconEmpty>
+				<Star size={24} />
+			</Rating.IconEmpty>
+			<Rating.IconFull>
+				<Star size={24} fill="white" />
```
endigo9740, line 13: Another hardcoded fill, switch to: `class="fill-surface-950-50"`

sites/next.skeleton.dev/src/examples/components/ratings/RTLExample.svelte
```diff
@@ -0,0 +1,17 @@
...
+    import { Rating } from '@skeletonlabs/skeleton-svelte';
+	import Star from 'lucide-svelte/icons/star';
+
+    let value = $state(2.5);
+</script>
+
+<div dir="rtl">
+    <Rating bind:value={value} fraction={2} width="w-44" interactive>
+        {#snippet iconFull()}
+            <Star size={24} fill="white" />
```
endigo9740, line 11: Another hardcoded fill, switch to: `class="fill-surface-950-50"`

sites/next.skeleton.dev/src/examples/components/ratings/RTLExample.tsx
```diff
@@ -0,0 +1,19 @@
...
+export const Page: React.FC<any> = () => {
+	const [value, setValue] = useState(2.5);
+	return (
+		<div dir="rtl">
+			<Rating value={value} onValueChange={setValue} fraction={2} width="w-44" interactive>
+				<Rating.IconEmpty>
+					<Star size={24} />
+				</Rating.IconEmpty>
+				<Rating.IconFull>
+					<Star size={24} fill="white" />
```
endigo9740, line 14: Another hardcoded fill, switch to: `class="fill-surface-950-50"`

sites/next.skeleton.dev/src/examples/components/ratings/UserInteractiveExample.svelte
```diff
@@ -0,0 +1,13 @@
+<script lang="ts">
+    import { Rating } from '@skeletonlabs/skeleton-svelte';
+	import Star from 'lucide-svelte/icons/star';
+</script>
+
+<Rating value={3} width="w-44" interactive>
+    {#snippet iconFull()}
+        <Star size={24} fill="white" />
```
endigo9740, line 8: Another hardcoded fill, switch to: `class="fill-surface-950-50"`

sites/next.skeleton.dev/src/examples/components/ratings/UserInteractiveExample.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+export const Page: React.FC<any> = () => {
+	const [value, setValue] = useState(3);
+	return (
+		<Rating value={value} onValueChange={setValue} width="w-44" interactive>
+			<Rating.IconEmpty>
+				<Star size={24} />
+			</Rating.IconEmpty>
+			<Rating.IconFull>
+				<Star size={24} fill="white" />
```
endigo9740, line 13: Another hardcoded fill, switch to: `class="fill-surface-950-50"`

packages/skeleton-react/src/lib/components/Rating/Rating.tsx
```diff
@@ -0,0 +1,196 @@
...
+    // Events
+    onMouseDown = () => {},
+    onKeyDown = () => {},
+    onValueChange = () => {},
+	// Children
+	children,
+}) => {
+    const figureRef = useRef<HTMLElement>(null);
+
+    const [focusedButtonIndex, setFocusedButtonIndex] = useState(0);
```
endigo9740, line 41: Keep your `useState` definitions grouped at the top, similar to `$state()` runes or vars in general.

packages/skeleton-react/src/lib/components/Rating/Rating.tsx
```diff
@@ -0,0 +1,196 @@
...
+        }
+
+        value = order + selectedFraction / fraction;
+        onValueChange(value);
+        onMouseDown(event, value);
+    }, [fraction]);
+
+    const onRatingKeyDown = useCallback((event: React.KeyboardEvent<HTMLButtonElement>) => {
+        if(!figureRef.current) return;
+
```
endigo9740, line 66: Remove the whitespace here (the line return). It'll help readability of the code.

packages/skeleton-react/src/lib/components/Rating/Rating.tsx
```diff
@@ -0,0 +1,196 @@
...
+    function decreaseValue() {
+        value = Math.max(0, value - 1 / fraction);
+        onValueChange(value);
+        refreshFocus();
+    }
+
+    return(
+        <>
+            <style>{`
```
endigo9740, line 100: Unless React has a specific implementation, I'd recommend keeping `<style>` blocks at the bottom of the template to match Svelte, Vue, etc

packages/skeleton-react/src/lib/components/Rating/Rating.tsx
```diff
@@ -0,0 +1,196 @@
...
+                                {children}
+                            </RatingContext.Provider>
+                    </button>
+                ))}
+            </figure>
+        </>
+    );
+};
+
+const IconEmpty: React.FC<IconProps> = ({
```
endigo9740, line 127: I think the dedicated `IconEmpty` and `IconFull` components may be overkill. For simple "slot-like" content (such as icons) React allows you to pass a `ReactNode` (read: React components themselves) as if they are props...

Types:
https://github.com/skeletonlabs/skeleton/blob/next/packages/skeleton-react/src/lib/components/Switch/types.ts#L63

Props:
https://github.com/skeletonlabs/skeleton/blob/next/packages/skeleton-react/src/lib/components/Switch/Switch.tsx#L42

Template Usage:
https://github.com/skeletonlabs/skeleton/blob/next/packages/skeleton-react/src/lib/components/Switch/Switch.tsx#L96

Example:
https://github.com/skeletonlabs/skeleton/blob/next/sites/next.skeleton.dev/src/examples/components/switch/ExampleIcons.tsx#L15

The end result will be a singular component that's much similar to the Svelte implementation! I realize this will take a non-trivial refactor, so let me know if you need help with this!

endigo9740 (changes requested):
Here's my review comments @Mahmoud-zino. This is a mix of small, medium, and large changes. Please review each and let me know if I can clarify anything.

Given the scope of some of these, I may be able to make myself available the next few days to help if you need it. Just let me know!

packages/skeleton-svelte/src/lib/components/Rating/Rating.svelte
```diff
@@ -0,0 +1,134 @@
...
+		<button
+			class="{buttonBase} {buttonPosition} {buttonAspect} {buttonClasses}"
+			class:pointer-events-none={!interactive}
+			tabindex={interactive && i === focusedButtonIndex ? 0 : -1}
+			onmousedown={(event: MouseEvent) => interactive ? onRatingMousedown(event, i) : undefined}
+			onkeydown={(event: KeyboardEvent) => interactive ? onRatingKeydown(event) : undefined}
+			type="button"
+		>
+			<!-- Icon: Empty -->
+			<span class="{emptyBase} {rxEmptyInteractive} {emptyClasses}" style="--clip_value: {(value - i) * 100}%">
```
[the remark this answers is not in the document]
Mahmoud-zino: yeah unfortunately this doesn't work on DOM elements :cry:

packages/skeleton-svelte/src/lib/components/Rating/Rating.svelte
```diff
@@ -0,0 +1,134 @@
...
+	let figureElement: HTMLElement;
+
+	// Dynamic Classes
+	const rxEmptyInteractive = $derived(interactive ? emptyInteractive : emptyStatic);
+	const rxFullInteractive = $derived(interactive ? fullInteractive : fullStatic);
+
+	// Use the latest rating or 0 if not available, ignoring fractions
+	let focusedButtonIndex = $derived(Math.max(0, Math.ceil(value - 1)));
+
+	$effect(() => {
```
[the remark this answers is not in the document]
Mahmoud-zino: Option 1 is so nice! I had to make some adjustments but it is working as expected, thank you for the cool tip :+1:

packages/skeleton-react/src/lib/components/Rating/Rating.tsx
Mahmoud-zino: Per the solution suggested above, there is no `style` block anymore :partying_face:

Mahmoud-zino:
@endigo9740 I applied all of your suggestions, thank you for your patience and always coming with cool suggestions :+1:

endigo9740:
@Mahmoud-zino and thanks for your help! I'm going to *try* to find some time to squeeze in a follow up review between now and our release on Tuesday, but no promises. Ideally I'd love to launch our next v3 pre-release with this, my upcoming nav components, and the recent progress ring component.

So feel free to polish anything anything you can in the meantime, and I'll ping you before I begin my review (likely Monday).

endigo9740:
@Mahmoud-zino I noticed the most recent deployment also failed. Make sure to do the following:

1. Delete the local `node_modules` in each project you made changes
2. Also delete the `package-lock.json` in the root of the monorepo
2. Upgrade your PNPM version to `9.3.0` (specifically)
3. Finally, run `pnpm i` to reinstall the local

This will do two things:

1. Ensure you're running the current version of PNPM, which @AdrianGonz97 recently made mandatory
2. Regenerate the lock file to ensure it's reflecting in all changes.

If you have any questions about this, ping me or Adrian this weekend and we can assist.

endigo9740:
@Mahmoud-zino I went ahead and implemented some last minute changes to this component. The only one worth mentioning is a minor refactor the keyboard navigation to support up/down in addition to left/right, per:

https://www.w3.org/WAI/ARIA/apg/patterns/radio/examples/radio-rating/#kbd_label

There's a few more things I'd like to revisit in the future, such as dynamically swapping between button/div based on whether there's interaction or not. But this is not worth holding the component back for now. We'll go ahead and merge and revisit in the future.

Great job on this one!

PR #2624, opened 2024-04-20
use variant-filled-primary class for active nav item to also use corr…

vhochstein:
currently example to highlight active nav item is using a bg background color class, however, that might lead to "incompatible" textcolor usage.

Therefore I would suggest to use a variant-filled-primary class instead, which will automatically select correct textcolor as well.

This is my first PR, so not sure what I have to comply to, but since it is an quite easy simple change.. 

## Linked Issue
None

## Description
currently example to highlight active nav item is using a bg background color class, however, that might lead to "incompatible" textcolor usage.

Therefore I would suggest to use a variant-filled-primary class instead, which will automatically select correct textcolor as well.

## Checklist

### DIFF
--- a/sites/skeleton.dev/src/routes/(inner)/elements/lists/+page.svelte
+++ b/sites/skeleton.dev/src/routes/(inner)/elements/lists/+page.svelte
@@ -195,7 +195,7 @@
 				</svelte:fragment>
 			</DocsPreview>
 			<p>To highlight active state, we recommend conditionally applying a background color to the anchor tag.</p>
-			<CodeBlock language="ts" code={`$: classesActive = (href: string) => (href === $page.url.pathname ? '!bg-primary-500' : '');`} />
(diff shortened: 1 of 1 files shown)


Sarenor:
That's not a bad improvement!
Usually, we like some discussion in an issue or Discord first and here's the contribution guide: https://www.skeleton.dev/docs/contributing
Only thing that's not quite correct is that the branch isn't prefixed with docs/ as far as I can see (and the pr out of nowhere no discussion).
Since it's a good suggestion I'd accept it though, will wait for @endigo9740 s final approval though.

Sarenor (approved):

endigo9740:
@vhochstein seems like a good change. Like @Sarenor said, we typically expect some upfront communication, but given the small scale of this we'll let it through. Please also make sure to use the checklist provided in the PR template as well. I've handled this for you this time. Thanks for the contribution.

PR #2673, opened 2024-05-18
feat: optionally prioritize new toasts

HenryMBaldwin:
## Description

Added prioritizeNew prop to Toast.svelte to allow enabling/disabling prioritizing newer toasts when the total number of toasts exceed the maximum. Currently older toasts are prioritized, and this is still the stock behavior as the prop defaults to false. 

## Changsets

## Checklist

### DIFF
--- a/.changeset/eleven-poets-rest.md
+++ b/.changeset/eleven-poets-rest.md
@@ -1,0 +1,5 @@
+---
+"@skeletonlabs/skeleton": minor
+---
+
+feat: Added prioritizeNew boolean prop to Toast.svelte that allows enabling/disabling newer toasts being prioritized when the total number of toasts is greater than the maximum defined by the max prop. Prop is false by default so default behavior is unchanged.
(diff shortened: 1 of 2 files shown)


endigo9740:
@HenryMBaldwin first off sorry for the long delay on this one. We've been snowed in with our pre-release prep and launches of Skeleton v3 and Floating UI Svelte.

I actually think this is a very interesting feature, and you've done a great job of following our contribution guidelines, so thanks for that. I can definitely see the appear of this feature, however, I'm cautious of adding new features this late in v2's lifecycle. As such I'm going to recommend we do the following:

1. I'm going to close out this PR without moving forward for now
2. I've already noted for consideration in the [pending v3 milestone issue](https://github.com/skeletonlabs/skeleton/pull/2673) for overlay systems
3. Then we will evaluate and approach a means to implement this in a systematic manner for v3

The v3 overlay system will see us combining toasts/drawers/modals under a singular system so they can have a consistent set of features. This has a number of positive implications for users. The result will be we can add your feature once, and it'll be available for all three!

Again, I appreciate the contribution and please do look forward to a version of this feature coming in the future.

Thanks again,
Chris

PR #2690, opened 2024-05-27
Increase default transition duration to 300ms to avoid a white flash

### LINKED ISSUE #2665: NEXT Focus ring/border flashes white when switching focus
### Current Behavior

Focus ring/border flashes white, see video

https://github.com/skeletonlabs/skeleton/assets/3662507/46bf0159-948d-4271-8bf0-9016fbc61528

### Expected Behavior

Focus ring/border does not flash

### Steps To Reproduce

1. go to [next skeleton docs](https://next.skeleton.dev/docs/tailwind/forms)
2. have cerberus theme selected (there its the most obvious)
3. switch focus on form elements

_No response_

### More Information

Chrome Version 125.0.6422.61 (64-Bit)
Windows 11 23H2

jaakkonakaza:
## Linked Issue

Closes #2665 

## Description

The default transition duration was 100ms, making it too fast to see properly. See the issue for before and after videos

## Changsets

## Checklist

- [ ] Ensure Svelte and Typescript linting is current - run `pnpm ci:check`

### DIFF
--- a/.changeset/mean-schools-double.md
+++ b/.changeset/mean-schools-double.md
@@ -1,0 +1,5 @@
+---
+"@skeletonlabs/skeleton": patch
+---
+
+bugfix: Increased the default animation duration of inputs to avoid a white flash
--- a/.changeset/mean-schools-double.md
+++ b/.changeset/mean-schools-double.md
@@ -2,4 +2,4 @@
 "@skeletonlabs/skeleton": patch
 ---
 
-bugfix: Increased the default animation duration of inputs to avoid a white flash
+bugfix: Increased the default animation duration of inputs and changed the transition to avoid a white flash
--- a/packages/skeleton/src/plugin/components/forms.css
+++ b/packages/skeleton/src/plugin/components/forms.css
@@ -45,7 +45,7 @@
 .select {
 	@apply field-outter field-inner;
 	@apply border-none;
-	@apply transition-[all] duration-100;
+	@apply transition-[all] duration-300;
 }
 
 /* Select --- */
--- a/packages/skeleton/src/plugin/components/forms.css
+++ b/packages/skeleton/src/plugin/components/forms.css
@@ -45,7 +45,7 @@
 .select {
 	@apply field-outter field-inner;
 	@apply border-none;
-	@apply transition-shadow duration-200;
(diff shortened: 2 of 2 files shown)


endigo9740:
@jaakkonakaza I solved this issue for Skeleton v2 a long while ago, but it's been so long I don't remember the specifics. So just a few things to consider here...

A good rule of thumb is to keep animations at around 200ms by default. I'm not sure why I opted for 100ms in the next version - probably just a typo, so let's aim for that. I don't believe the duration alone will be enough to solve the issue though.

Comparing to v2, we use this the generic `transition` styles. I believe this may be the culprit. We're currently opted for `transition-[all]`:

- v2: https://github.com/skeletonlabs/skeleton/blob/master/packages/plugin/src/styles/components/forms.css#L20
- v3: https://github.com/skeletonlabs/skeleton/blob/next/packages/skeleton/src/plugin/components/forms.css#L48

Beyond that, the only thing that sticks out to me is the move from `border` -> `ring`. I'd prefer to stick with ring, but if it's the culprit then that may need to be adjusted as well.

If you can try these few small adjustments, I'm happy to retest. Vercel will generate a preview URL directly in this thread if that helps.

jaakkonakaza:
https://github.com/skeletonlabs/skeleton/assets/34682147/1ebf8cca-d726-4e11-80ab-2d86ef1de72b

I actually noticed some problems on mobile in v2, this new one doesn't have that same problem. I also think the animation on this branch looks nicer on desktop than the v2 one

jaakkonakaza:
https://github.com/skeletonlabs/skeleton/assets/34682147/6fe5de88-8d7c-4148-898a-0e6919787bde

https://github.com/skeletonlabs/skeleton/assets/34682147/e7a9e37c-cfda-4f9f-92b3-fbda6eae9bf6

Here you can compare the old and new, which one do you prefer?

endigo9740:
@jaakkonakaza I've swapped to the generic `transition` class. While yes, the `transition-shadow` is more isolated, I think in many cases folks are going to implement their own styles on top of this - such as a red background for invalid states. We don't want folks to have to consult the source every time they do this. So this is likely a happy medium between all and pure isolation.

PR #2716, opened 2024-06-07
Resolve A11y errors for AppRailAnchor

### LINKED ISSUE #2662: Additional a11y warnings to resolve
### Current Behavior

We recently updated a number of components to silence a11y warnings provided in newer versions of Svelte. Most errors have been resolved, but it looks like we missed on for the AppRailAnchor

Screenshot 2024-05-15 at 9 22 00 AM

### Expected Behavior

The error is due to the slew of events forwarded here:

https://github.com/skeletonlabs/skeleton/blob/master/packages/skeleton/src/lib/components/AppRail/AppRailAnchor.svelte#L53

We should either:

1. Remove the extraneous events
2. Set a proper aria role (per the error) - best guess would be `button` off-hand

### Steps To Reproduce

The errors appear when starting a Skeleton app. Regardless if you use the AppRailAnchor or not.

_No response_

### More Information

Please consult with me before making this change. We need to decided between the two options provided above.

endigo9740:
## Linked Issue

Closes #2662

## Description

Removes the extra role and forward events from the `<AppRailAnchor>` component. Svelte was throwing a11y errors for these because they are invalid on an anchor element.

## Changsets

## Checklist

### DIFF
--- a/.changeset/unlucky-keys-brake.md
+++ b/.changeset/unlucky-keys-brake.md
@@ -1,0 +1,5 @@
+---
+"@skeletonlabs/skeleton": patch
+---
+
+bugfix: Removed invalid role and forward events on AppRailAnchor component
--- a/packages/skeleton/src/lib/components/AppRail/AppRailAnchor.svelte
+++ b/packages/skeleton/src/lib/components/AppRail/AppRailAnchor.svelte
@@ -45,21 +45,7 @@
 	}
 </script>
 
-<a
-	class="app-rail-anchor {classesBase}"
-	href={$$props.href}
-	{...prunedRestProps()}
-	role="button"
-	on:click
-	on:keydown
-	on:keyup
-	on:keypress
-	on:mouseover
-	on:mouseleave
-	on:focus
-	on:blur
-	data-testid="app-rail-anchor"
->
+<a class="app-rail-anchor {classesBase}" href={$$props.href} {...prunedRestProps()} data-testid="app-rail-anchor">
 	<div class="app-rail-wrapper {classesWrapper}">
 		{#if $$slots.lead}<div class="app-rail-lead {classesLead}"><slot name="lead" /></div>{/if}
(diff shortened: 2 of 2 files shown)


JonnyBGod:
This introduced a breaking change and should not have been included as a minor change.

Also svelte was only complaining about the role attribute not the events.

Please consider reverting the event related changes and try to be more carefull when introducing breaking changes in the future.

JonnyBGod:
Cool, I understand and agrees that most of those events are probably overkill.

My usecase is quite simple I have a drawer nenu with Anchors that use the onclick to triguer a function that manages some state and closes the drawer.

I understand there are multiple ways of doing it but this is the straightforwars and less hacky one.
Also not easy to put that "on:click" event someplace else.
And the button based version with goto would work but would lose all the svelete goodies around the anchor such as prerendering without again more hacking.

IMO the on click event should be mostly a defacto for everything at this point, since we are talking about SPA enabling frameworks where it is very common to require side effects for user actions.
The other events can usualy be deal with with css apart from some edge cases that I think it is very accepatble to neglect.

endigo9740:
@JonnyBGod I've created an issue for this. Please move any further discussion on this topic to here:
https://github.com/skeletonlabs/skeleton/issues/2742

Note that I'm just returning from a week-long vacation, so it may take a couple days to get to this. And it won't be in place until the next release (two weeks from tomorrow). In the meantime I might suggest rolling back to a prior release until this is sorted. I also welcome a PR if you can get to it sooner.

I went ahead and checked the status for our upcoming Skeleton v3 version of this feature, which you can see here:

- https://next.skeleton.dev/docs/components/navigation/svelte/#rail
- Source: https://github.com/skeletonlabs/skeleton/blob/next/packages/skeleton-svelte/src/lib/components/Navigation/NavTile.svelte#L70

There's now a single universal "tile" component that updates the tag dynamically. It defaults to `<button>` but will switch to `<a>` when `href` is present. Both include the `onclick` event by default.

Issue #2733, opened 2024-06-22
Refactor React components to fall more inline with existing React libraries


eagerestwolf:
### Describe the feature in detail (code, mocks, or screenshots encouraged)

So, the time has come. I am going to open this issue as a place to track whats going on, and to communicate changes and whatnot before opening a full fledged pull request. I am by no means a React expert (though I follow one on YouTube lol), with that said the goal of this issue/pull request is going to be manyfold:

- First, optimize the React components. Running the React app right now, it's using a pretty healthy chunk of my CPU (and I have a Ryzen 5 5600X); so there are probably some unnecessary re-renders happening throughout the code base (and by that I mean there are, I have been starting on the rewrite already).
- Second, bring the library more inline with what the React community is largely already doing, thus making the library feel easy to use (and debug if needed) for React developers.
- Third, try to optimize compatibility with as many frameworks as possible, including vanilla React, NextJS, Remix, Quik, React Router v7 aka Remix v3 (once I can start testing that), Preact, SolidJS (don't get your hopes up, I just want to make as many components as possible work for you guys), and (of course) Astro.

With that said, I have made one change already (that I feel is 100% necessary, fight me) and that is adding a path alias to the `tsconfig.json` for the `./src/*` path (of course it's `@/*`, fight me again). However, I do have a few proposed changes that need to be run by the Skeleton team before I implement them:

[...]

PR #2734, opened 2024-06-23
Remove `reactCompose` utility

eagerestwolf:
## Linked Issue

Addresses one issue raised in #2733  (side note: I will from now on open smaller issues, this was just an easy target)

## Description

This PR makes 2 notable changes. 

1. Adds a path alias to `vite.config.ts`, `tsconfig.json`, and `tsconfig.package.json` to redirect the standard alias `@` to `__dirname/src`.
2. All components that rely on composition and previously used the `reactCompose` function, now instead implement `Object.assign` directly.

It's nice to make a small, easy change this time.

## Changsets

Changesets have been added. I'll just need to push one more commit to update my message (my mistake, I forgot to follow conventional commit standards).

## Checklist

- [x] This PR targets the `dev` branch (NEVER `master`) **this PR actually targets `next`**

### DIFF
--- a/packages/skeleton-react/src/lib/components/Tabs/Tabs.tsx
+++ b/packages/skeleton-react/src/lib/components/Tabs/Tabs.tsx
@@ -1,4 +1,9 @@
+'use client';
+
 import React, { useEffect, useRef, useState } from 'react';
+
+import { reactCompose } from '@/lib/utils/react-compose.js';
+
 import { TabsProps, TabsListProps, TabsControlProps, TabsControlItemProps, TabsPanelItemProps } from './types.js';
 
 // Components ---
@@ -160,7 +165,7 @@ const TabsControl: React.FC<TabsControlProps> = ({
 				onKeyUp={onKeyup}
 			>
 				{/* Keep these classes on wrapping element */}
-				<div className="h-0 w-0 flex-none overflow-hidden">
+				<div className="flex-none w-0 h-0 overflow-hidden">
 					<input
 						ref={elemInputRef}
 						type="radio"
@@ -215,7 +220,7 @@ const TabsPanelItem: React.FC<TabsPanelItemProps> = ({
 	);
 };
 
-export const Tabs = Object.assign(TabsRoot, {
+export const Tabs = reactCompose(TabsRoot, {
 	List: TabsList,
 	Control: TabsControl,
 	Panel: TabsPanelItem,
--- a/packages/skeleton-react/src/lib/components/Tabs/Tabs.tsx
+++ b/packages/skeleton-react/src/lib/components/Tabs/Tabs.tsx
@@ -163,7 +163,7 @@ const TabsControl: React.FC<TabsControlProps> = ({
 				onKeyUp={onKeyup}
 			>
 				{/* Keep these classes on wrapping element */}
-				<div className="flex-none w-0 h-0 overflow-hidden">
+				<div className="h-0 w-0 flex-none overflow-hidden">
 					<input
 						ref={elemInputRef}
 						type="radio"
--- a/packages/skeleton-react/src/lib/components/Tabs/Tabs.tsx
+++ b/packages/skeleton-react/src/lib/components/Tabs/Tabs.tsx
@@ -2,8 +2,6 @@
 
 import React, { useEffect, useRef, useState } from 'react';
 
-import { reactCompose } from '@/lib/utils/react-compose.js';
-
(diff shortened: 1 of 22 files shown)


Hugos68:
`reactCompose` was created a while ago by me but we can actually remove `reactCompose` as it's a dirty wrapper for `Object.assign`.

So instead of moving everything to `reactCompose` you may to the opposite and move everything to `Object.assign`. It removes a unneccesary wrapper. Object.assign is built in and more clear.

Hugos68 (changes requested):
`reactCompose` is a old wrapper that essentially just does `Object.assign` I suggest we get rid of the `reactCompose` function all together as there is less overhead and easier to understand.

packages/skeleton-react/vite.config.ts
```diff
@@ -13,6 +13,11 @@ export default defineConfig({
 		RemixRouter(),
 		skeletonPluginWatcher(path.resolve(path.join('..', '..', 'packages', 'skeleton', 'src', 'plugin')))
 	],
+	resolve: {
+		alias: {
+			'@': path.resolve(__dirname, './src')
```
Hugos68, line 18: In Svelte land we always use `$lib` as the alias, if it's possible could we keep the same alias across our component libraries? I'm not sure if `@` is a react convention but if not we should probably aim to keep as much parity between the codebases.

eagerestwolf: I can definitely do that either before I go to work this evening, or after I get home tomorrow morning. Should be a quick, easy commit.

Hugos68: Yeah, I don't care what the alias exactly is but consistency between projects would be nice both for us and future contributors

Hugos68:
Is this ready to be merged?

endigo9740:
@Hugos68 I'm not merging anything this week while I'm out. I'll review all pending PRs when I return next week.

eagerestwolf:
Yup, everything is done on my end. Whenever Chris comes back from vacation everything is ready to go.

Hugos68:
I noticed in a couple of places tailwind classes are being shifted around, this should not be happening, please run format before pushing, I assume this comes from your workspace formatter.

packages/skeleton-react/src/lib/components/Tabs/Tabs.tsx
```diff
@@ -160,7 +163,7 @@ const TabsControl: React.FC<TabsControlProps> = ({
 				onKeyUp={onKeyup}
 			>
 				{/* Keep these classes on wrapping element */}
-				<div className="h-0 w-0 flex-none overflow-hidden">
+				<div className="flex-none w-0 h-0 overflow-hidden">
```
endigo9740, line 166: @eagerestwolf pointing out an example of what Hugo was referring to.

endigo9740 (changes requested):
I agree with @Hugos68's suggestion above. If you can revert these class adjustments, then everything else looks good. I'm happy to merge.

endigo9740:
@eagerestwolf sorry for multiple pings. But wanted to give you a heads up about these incoming updates. I did a bit of an overhaul to the React Tabs component, and I may adjust further next week. We the implementation didn't quite match Svelte, but it should be much closer now. Feedback is welcome though.

https://github.com/skeletonlabs/skeleton/pull/2748

Also FYI it's a holiday weekend, so I'll be in and out the next few days, but ping me if you need anything. I typically respond most quickly on Discord!

Thanks!

eagerestwolf:
Fixed the Tailwind class order issues, and I added an option to the `.vscode/settings.json` to not allow Headwind to sort on save (this won't cause any problems if the user doesn't have Headwind installed or it's disabled), as well as adding Headwind to the `unwantedExtensions` in `.vscode/extensions.json`. If I missed any classes let me know. I also apologize of the radio silence. I have been incredibly busy with work (we tend to get busy closer to major holidays because I work in the package delivery industry).

endigo9740:
@eagerestwolf thanks for the quick turn around. And no worries on the response time. We just wanted to communicate the info, and figured you would respond when possible. If you're ever pressed for time and can't implement changes in a timely manner, just let us know. Communication is key ;)

As per the PR itself, all looks well. Just two minor items of note:

1. I went ahead ahead and modified the Changeset to better describe both changes in the PR.
2. The exception for Headwind should be fine, but I know you were previously trying to configure Prettier Tailwind as well. So jus for your own purposes - you likely want to commit to one or the other. Mixing the two seems like a bad idea!

That aside, I'm merging this now. Thanks for your contribution! I'll note this PR in the original thread to track overall progress.

PR #2754, opened 2024-07-11
Feat: Listbox

### LINKED ISSUE #2387: NEXT: Listboxes

## Zag Integration

NOTE: the [Zag team has confirmed](https://github.com/chakra-ui/zag/discussions/1807#discussioncomment-10533452) there's no 1:1 equivalent of Listboxes available, nor does it sound like this is something they will consider in the future. Given this, we'll keep this issue pending, with one of the two possible outcomes:

1. We build it ourselves, which would mean it's an outlier compared to our Zag-powered components
2. We retire this component and it does not carry forward to v3

Right now we're leaning towards option number 2 for launch. Though we can definitely revisit in the future.

## Maintainer Requests

- [ ] Adheres to the [ARIA Guidelines](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/)
- [ ] Design should be based on the Figma mocks
- [ ] Implementation can mostly mirror that of the [Segment Control](https://next.skeleton.dev/docs/components/segment-control/svelte) ([Svelte](https://github.com/skeletonlabs/skeleton/tree/next/packages/skeleton-svelte/src/lib/components/Segment) / [React](https://github.com/skeletonlabs/skeleton/tree/next/packages/skeleton-react/src/lib/components/Segment))
- [ ] Supports a [multiple selection mode](https://www.skeleton.dev/components/listboxes#multiple-selection). In v2 we did this by [toggling between checkbox and radio inputs](https://github.com/skeletonlabs/skeleton/blob/master/packages/skeleton/src/lib/components/ListBox/ListBoxItem.svelte#L124).
  - Reference the [React prototypes](https://github.com/skeletonlabs/react-form-components/tree/main/src/components) we created for form control components if relevant.
- [ ] Should support max height with vertical scrolling
- [ ] Optional: Consider adding support for [group options](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/examples/listbox-grouped/) (not mocked)

## Community Requests

* n/a

## Bugs and Issues

* n/a

## Feedback

Hugos68:
## Linked Issue

Closes #2387 

### Svelte

- [x] Implementation
- [ ] Tests
- [ ] Docs

### React

- [x] Implementation
- [ ] Tests
- [ ] Docs

### Figma
Design: https://www.figma.com/design/KiJ0E0sQk9ouMY6ProdZPJ/Skeleton-v3---Design-System?node-id=3828-62174&t=0FCCZG6rRSldYZYu-0

## Changsets

## Checklist

- [x] This PR targets the `next` branch (NEVER `master`)


Hugos68:
@endigo9740 A review for the Svelte version would be appreciated. I tried to match the Figma design and I haven't split up the classes into multiple props simply because I don't know which props should be configurable. Hopefully this is made clear in the review.

packages/skeleton-svelte/src/lib/components/Listbox/ListboxItem.svelte
```diff
@@ -0,0 +1,92 @@
...
+		lead,
+		trail,
+		...attributes
+	}: ListboxItemProps = $props();
+
+	// Context
+	const context = getListboxContext();
+
+	// Local
+	const id = createId();
```
endigo9740, line 29: I've followed up regarding the ID system in Discord. I think this might be a good point to talk through our approach to keyboard interaction for all components. Start implementing our shared conventions and paradigms around this, if not dedicated tooling (in-house, Zag, etc)

packages/skeleton-svelte/src/lib/components/Listbox/ListboxItem.svelte
```diff
@@ -0,0 +1,92 @@
...
+		trailBase,
+		trailClasses,
+		children,
+		lead,
+		trail,
+		...attributes
+	}: ListboxItemProps = $props();
+
+	// Context
+	const context = getListboxContext();
```
endigo9740, line 26: I've used `ctx` in favor of `context` in other components. It's a common pattern I picked up from working with Canvas elements. We should converge on a standard here.

packages/skeleton-svelte/src/lib/components/Listbox/ListboxItem.svelte
```diff
@@ -0,0 +1,92 @@
...
+
+	// Local
+	const id = createId();
+	const listboxItemSelector = `[data-skeleton-id="${context.id}"] > [data-skeleton-part="listbox-item"]`;
+
+	// Events
+	const onclick: (typeof attributes)['onclick'] = (event) => {
```
endigo9740, line 33: I've noticed it can feel a bit weird mixing passed event props (ex: `onclick`) and implementing local functions (ex: also called `onclick` here). It's not immediately clear the source of these, and could lead to some potential naming conflicts.

For other components I've opted for the following conventions...

Passed event props adhere to the framework preference:

- `onclick` for Svelte; `onClick` for React
- `onkeydown` for Svelte; `onKeyDown` for React

Then for the local functions we can set our own standard since this is all internal:

- `function onClickHandler() { ... }`
- `function onKeyDownHandler() { ... }`

> Camel case + "handler" at the end.

This is just one idea for the convention, but I'm open to other suggestions.

endigo9740: I'd also welcome some insight as why everyone is moving from:

`function doSomething() {}`

To:

`const doSomething = () => {}`

Wasn't sure if it was just personal preference, but last I looked into things scope works different between standard and arrow functions.

packages/skeleton-svelte/src/lib/components/Listbox/types.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export interface ListboxProps extends HTMLAttributes<HTMLDivElement> {
+	/** Set the item name. */
+	name: string;
+	/** The value of the selected item(s). */
+	value?: string | string[];
+	/** Whether multiple items can be selected. */
+	multiple?: boolean;
+	/** Set the base classes. */
+	base?: string;
```
endigo9740, line 20: Take a look at how we divide the various types of prop types. Just a line break + leading comment.
https://github.com/skeletonlabs/skeleton/blob/next/packages/skeleton-svelte/src/lib/components/Segment/types.ts#L19

 It helps readability a bit, or at least my old man brain :D

sites/next.skeleton.dev/src/content/docs/components/listbox/svelte.mdx
```diff
@@ -0,0 +1,3 @@
+---
+layout: '@layouts/LayoutDoc.astro'
```
endigo9740, line 2: I might recommend setting up at least one Svelte demo before switching over to React. You never know when we might run into an Astro quirk! Better to catch that sooner than later.

endigo9740 (commented):
@Hugos68 overall this looks great. There's some easy adjustments here, and of course we'll circle back around keyboard interaction asap.

packages/skeleton-svelte/src/lib/components/Listbox/Listbox.svelte
```diff
@@ -0,0 +1,57 @@
...
+	import { setListboxContext } from './context.js';
+	import type { ListboxProps } from './types.js';
+	import { createId } from '$lib/internal/create-id.js';
+
+	// Props
+	let {
+		name = '',
+		value = $bindable(),
+		multiple = false,
+		base = 'grid gap-2 overflow-y-auto',
```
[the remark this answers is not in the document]
Hugos68: Done

packages/skeleton-svelte/src/lib/components/Listbox/Listbox.svelte
```diff
@@ -0,0 +1,57 @@
...
+		padding = 'p-4',
+		border = 'border border-surface-200-800',
+		radii = 'rounded-container',
```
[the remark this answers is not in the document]
Hugos68: Done

packages/skeleton-svelte/src/lib/components/Listbox/ListboxItem.svelte
Hugos68: Done, agreed.

packages/skeleton-svelte/src/lib/components/Listbox/ListboxItem.svelte
Hugos68: Yeah, I really think Zag is really great, I definitely want to consider it.

packages/skeleton-svelte/src/lib/components/Listbox/ListboxItem.svelte
Hugos68: First point:
The reason I named it `onclick` and not `onclickHandler` or whatever is so that when applying it to the component I can use Svelte's shorthand: `{onclick}` instead of doing: `onclick={onClickHandler}` which is much more verbose. I can change it though, that's not a problem at all, just explaining my reasoning, lmk how you feel about this.

Hugos68: Second point:
Using `const` for arrow functions makes code more maintainable in my experience for 3 reason:
1. Arrow functions don't hoist, which is confusing behaviour
2. Arrow functions don't inherit `this` which is also a source of confusion
3. Arrow functions are more concise and less verbose when overloading

So mainly a preference, but for good reasons IMHO.

packages/skeleton-svelte/src/lib/components/Listbox/Listbox.svelte
```diff
@@ -0,0 +1,57 @@
...
+	import type { ListboxProps } from './types.js';
+	import { createId } from '$lib/internal/create-id.js';
+
+	// Props
+	let {
+		name = '',
+		value = $bindable(),
+		multiple = false,
+		base = 'grid gap-2 overflow-y-auto',
+		padding = 'p-4',
```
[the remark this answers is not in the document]
Hugos68: Done

packages/skeleton-svelte/src/lib/components/Listbox/types.ts
Hugos68: Done!

packages/skeleton-svelte/src/lib/components/Listbox/ListboxItem.svelte
endigo9740: Yeah, to put it another way. Imagine I had a complete template that implemented:

```html
<!-- SomeComponent.svelte -->
<button {onclick}>...</button>
```

Without seeing more of the component you don't know if the source is internal or external. Even if it's more verbose, being able to quickly identify that would be handy. We can likely still opt for the shorthand on prop provided event handlers.

Plus if you need to trigger the prop provided event handler inside a local function, you might end up with:

```
function onclick() {
    // ...
    onclick?.();
}
```
Which would fail because the names conflict.

packages/skeleton-svelte/src/routes/components/listbox/+page.svelte
```diff
@@ -0,0 +1,57 @@
...
+
+	let data = $state({});
+
+	const onsubmit = (e: SubmitEvent) => {
+		e.preventDefault();
+		data = formDataToObject(new FormData(e.target as HTMLFormElement));
+	};
+</script>
+
+<h1 class="h1">Listbox</h1>
```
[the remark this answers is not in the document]
Hugos68: I've added some examples of various uses aswell as use uniform styling.

Hugos68:
FYI @endigo9740 I really vouch to remove the `multiple` prop in favor of deriving it off wether the value is and array or not. The multiple prop causes a lot of confusion internally and externally as it's basically defining behaviour twice, through multiple, and through `value`.

packages/skeleton-svelte/src/routes/components/listbox/+page.svelte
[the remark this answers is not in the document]
Hugos68: Done!.

sites/next.skeleton.dev/src/content/docs/components/listbox/svelte.mdx
Hugos68: Done! No quirks yet.

packages/skeleton-svelte/src/lib/components/Listbox/ListboxItem.svelte
Hugos68: I'm actually already doing this through `...attributes`:
```js
const onclick: (typeof attributes)['onclick'] = (event) => {
    ctx.isSelected(value) ? ctx.deselect(value) : ctx.select(value);
    attributes.onclick?.(event); // External 
};

endigo9740: First I've seen this. Is `attributes` similar to `$$props` from Svelte 4?

Hugos68: This is similar to $$restProps

Hugos68:
I've now utilized the `btn` class aswell as use `hover:` classes for the hover state and `focus:` classes for focus state.

endigo9740:
@Hugos68 now that we're past the hump of integrating Zag, we should review options for how we wish to handle this component. Sadly Zag does not appear to provide a prop Listbox solution. The closest I can find is [Menu](https://zagjs.com/components/react/menu), but even that is not quite right. Listboxes should not require a trigger element to be shown, they can be fixed on the page. I'm open to any other ideas you have here.

Hugos68:
@endigo9740 The Listbox is actually present under the Select machine: https://zagjs.com/components/react/select
It supports single and multiple selection.

Hugos68:
We can use the https://zagjs.com/components/react/select.
It supports a popup feature with a trigger but you can use it without too. It supports both multiple and single selection, so sounds like what we want!

endigo9740:
Cool, that makes sense. In that case I'll leave it up to you if you would prefer to try and update this PR or start anew. The latter might not be the worst option.

PR #2767, opened 2024-07-17
docs: make svelte switch example in sync with react

bartduisters:
## Sync for example apps

This brings svelte in sync with the example react app (and vue/angular).


endigo9740:
@bartduisters the local apps for the Svelte/React apps are dev sandboxes. They are not intended to be use facing, so no need for parity. Just gives us a place to build and test things.

Additionally Svelte has two-way data binding, so we would use `bind:checked={stateVar}`.

bartduisters:
Can be closed then!

PR #2768, opened 2024-07-17
feat: remove/sync base classes switch

bartduisters:
Svelte used `inline`, React used `flex`, removing both has the same result.


bartduisters:
Wrongly tagged commit. It's not a feature.

PR #2770, opened 2024-07-17
refactor: consistent progress API

bartduisters:
I've synced the API between svelte and react.
Also the examples had a non-working example for changing the background (it used meterClasses instead of meterBg, the default of meterBg takes precedent).

### DIFF
--- a/packages/skeleton-react/src/lib/components/Progress/Progress.tsx
+++ b/packages/skeleton-react/src/lib/components/Progress/Progress.tsx
@@ -5,7 +5,7 @@ import { ProgressProps } from './types.js';
 export const Progress: FC<ProgressProps> = ({
 	value,
 	max = 100,
-	labelledBy = '',
+	labelledby = '',
 	// Root
 	base = 'overflow-x-hidden',
 	bg = 'bg-surface-200-800',
@@ -35,7 +35,7 @@ export const Progress: FC<ProgressProps> = ({
 		<>
 			<div
 				role="progressbar"
-				aria-labelledby={labelledBy}
+				aria-labelledby={labelledby}
 				aria-valuenow={value}
 				aria-valuemin={0}
 				aria-valuemax={max}
--- a/packages/skeleton-react/src/lib/components/Progress/Progress.tsx
+++ b/packages/skeleton-react/src/lib/components/Progress/Progress.tsx
@@ -5,7 +5,7 @@ import { ProgressProps } from './types.js';
 export const Progress: FC<ProgressProps> = ({
 	value,
 	max = 100,
-	labelledby = '',
+	labelledBy = '',
 	// Root
 	base = 'overflow-x-hidden',
 	bg = 'bg-surface-200-800',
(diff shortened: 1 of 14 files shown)


packages/skeleton-react/src/lib/components/Progress/Progress.tsx
```diff
@@ -5,7 +5,7 @@ import { ProgressProps } from './types.js';
 export const Progress: FC<ProgressProps> = ({
 	value,
 	max = 100,
-	labelledBy = '',
+	labelledby = '',
```
endigo9740, line 8: @bartduisters we don't call it out anywhere (yet), but we opt for camel-case for multi-word props names that *we* control. I realize it's confusing when compared to the raw ARIA attribute which is all lowercase though.

To further confuse things, event handlers also differ per framework:

- Svelte 5: `onclick` (lowercase)
- React: `onClick` (camel-case)

For context, this adheres to a change in Svelte v5. In prior versions they used `on:click`.

In practice they look like this:

## Svelte

```html
<!-- handler -->
<button onclick={doSomething}>
<!-- forwarded -->
<button onclick>
```

## React

```jsx
<button onClick={doSomething}>
```

> NOTE: can't recall if React allows for forwarding or not off hand.

packages/skeleton-react/src/routes/components/progress.tsx
```diff
@@ -48,9 +48,9 @@ export function Component() {
 				<p>bg-primary-500 (default)</p>
 				<Progress value={value} max={max} />
 				<p>bg-secondary-500</p>
-				<Progress value={value} max={max} meterClasses="bg-secondary-500" />
+				<Progress value={value} max={max} meterBg="bg-secondary-500" />
```
endigo9740, line 51: FYI, no change need here, but we should be using Tailwind convention for style prop names. So `bg` is correct , but I believe we've strayed from this in other components and fell back to `background`. Just a heads up! This is something I hope to correct soon.

endigo9740 (changes requested):

packages/skeleton-react/src/lib/components/Progress/Progress.tsx
bartduisters: I initially changed it to camel case. But then noticed that on the Switch component it was all lowercase and thought that that was the convention.

I'll change the others then and add it to this PR

packages/skeleton-react/src/routes/components/progress.tsx
bartduisters: `meterClasses` does exist as a prop on Progress, but it is applied after `meterBg`

```html
<div class="{meterBase} {meterBg} {meterRounded} {meterTransition} {rxIndeterminate} {meterClasses}" style:width={fillPercentage}></div>
```

The default is e.g.  `bg-surface-500` if a custom bg e.g. `bg-secondary-500` is applied to `meterClasses`, then the resulting string would be:
```html
<div class="bg-surface-500 bg-secondary-500" style:width={fillPercentage}></div>
```

The end example then just shows `bg-surface-500`.

I just changed this to have a working example in the sandbox.

endigo9740: Sorry I was speaking specifically to the prop name - no change needed there. The change to where the color prop is provided is valuable and should remain!

PR #2778, opened 2024-07-23
NEXT: Update components to use Zag.js

Hugos68:
This PR integrates [Zag](https://zagjs.com/) into our existing suite of Svelte and React components. This also updates the respective documentation and examples for each component.

The following indicates the status of updates per components:

| Component | Svelte | React | Docs | Owner |
|:-:|:-:|:-:|:-:|:-:|
| Accordion | ✅ | ✅ | ✅ | Hugo |
| Avatar | ✅ | ✅ | ✅ | Chris |
| Progress | ✅ | ✅ | ✅ | Hugo |
| Progress Ring | ✅ | ✅ | ✅ | Chris |
| Rating | ✅ | ✅ | ✅ | Hugo |
| Seg. Control | ✅ | ✅ | ✅ | Chris |
| Switch | ✅ | ✅ | ✅ | Chris |
| Tabs | ✅ | ✅ | ✅ | Chris |

> NOTE: AppBar and Navigation components will not use Zag.

Also implements:

- [x] `useId()` local utility function for `skeleton-svelte`

## Remaining Tasks

- [x] Resolve Astro docs CI deployment issue
- [x] Ensure React `schema.ts` information is up to date
- [x] Run PR checklist tasks (format, lint, etc)

Documentation:

- [x] React Ratings: do not handle interaction the same as Svelte (needs `onValueChange`)
- [x] Segment Control: orientation example should not use item `disable`
- [x] Segment Control: import button group example at bottom
- [x] Switch: Improve List example (text on left, switch on right)
- [x] Tabs: examples need more variety; too same-y

Framework Components:

- [x] Verify all components have uniform use of `data-testid` attributes
- [x] Verify consistent use of Svelte `@component` component comment descriptions
- [x] Ensure Zag-related machine/api logic is co-located under a single `// Zag` comment (see Ratings)
- [x] Accordion: has hardcoded `flex-1` class in template - needs to be move to style prop
- [x] Svelte Accordion: bindable `value` needs to be provided to `context: {}` ala Segment Control, etc
- [x] React Accordion: animations not working as expected on panel close
  - [x] Remove the React Framer Motion dependency, which is now covered by Zag
- [x] React Accordion: the `disabled` prop is non-functional and not visually displayed
- [x] Svelte Switch: bindable props need to be passed to `context: {}` (ex: checked, disabled, compact)

Post-Release Follow Ups:

- [ ] Fix and update React component unit tests
- [ ] Decide on the use of `<figure>` for data visualization components
- [ ] Improve both global and component-specific focus states
  - React Ratings components cannot be tab focused when value is `0`
- [ ] Update Zag packages to utilize PNPM categories, which keeps versions in sync
- [ ] Improve Ratings component built-in star SVGs
- [ ] Zag component integration process needs to be added to the Contributor docs
- [ ] Update React Accordion to use Zag animations (reference: [presence](https://zagjs.com/components/react/presence))
- [ ] Improve consistency of JSDoc comments in `types.ts` for all components

## Checklist

## Changsets


packages/skeleton-react/package.json
```diff
@@ -40,6 +40,7 @@
 		"!./dist/schemas.d.ts"
 	],
 	"dependencies": {
+		"@zag-js/switch": "^0.62.1",
```
bartduisters, line 43: This should be moved to devDependencies as well

packages/skeleton-svelte/src/lib/components/Avatar/Avatar.svelte
```diff
@@ -17,17 +20,39 @@
...
 		imageClasses = '',
+		// Fallback
+		fallbackBase = 'w-full h-full flex justify-center items-center',
+		fallbackClasses = '',
 		// Snippets
 		children
 	}: AvatarProps = $props();
+
+	// Zag
+	const [snapshot, send] = useMachine(avatar.machine({ id: '1' }));
```
bartduisters, line 31: the `id` needs to be random

packages/skeleton-react/package.json
endigo9740: Svelte sadly doesn't have a random ID generator built in, so we'll have to implement our own.

endigo9740: A local `useId` util function has been implemented into the `skeleton-svelte` package.

packages/skeleton-svelte/src/lib/internal/uuid.ts
```diff
@@ -0,0 +1,7 @@
+// UUID
```
Hugos68, line 1: a UUID is a spec for a type of ID, we should not name this `uuid` or even mention it, this is an id, not anything else.

packages/skeleton-svelte/src/lib/components/Progress/Progress.svelte
```diff
@@ -18,30 +19,34 @@
 		meterRounded = 'rounded',
 		meterTransition = 'transition-[width]',
 		meterAnimate = 'animate-indeterminate',
-		meterClasses = ''
+		meterClasses = '',
+		// Snippets
+		children,
+		...zagProps
```
endigo9740, line 25: Add a `// Zag` comment above to denote this as it's own grouping

packages/skeleton-svelte/src/lib/components/Progress/Progress.svelte
```diff
@@ -18,30 +19,34 @@
...
-<div
-	role="progressbar"
-	aria-labelledby={labelledBy}
-	aria-valuenow={value}
-	aria-valuemin={0}
-	aria-valuemax={max}
-	class="{base} {bg} {width} {height} {rounded} {classes}"
->
-	<div class="{meterBase} {meterBg} {meterRounded} {meterTransition} {rxIndeterminate} {meterClasses}" style:width={fillPercentage}></div>
+<div {...api.getRootProps()}>
```
endigo9740, line 42: Since we're introducing a new element here, we should add a new set of style props. I'd recommend `lableX` for the prefix.

packages/skeleton-svelte/src/lib/components/Progress/Progress.svelte
```diff
@@ -18,30 +19,34 @@
...
+	{#if children}
+		<div {...api.getLabelProps()}>{@render children(api.value)}</div>
+	{/if}
+	<div class="{base} {bg} {width} {height} {rounded} {classes}" {...api.getTrackProps()}>
```
endigo9740, line 46: Go ahead and add an HTML comment above each respective element within the template. This should match then identity that Zag provides. Including:

* `<!-- Label -->`
* `<!-- Track -->`
* `<!-- Meter -->`

packages/skeleton-svelte/src/lib/components/Progress/Progress.svelte
```diff
@@ -18,30 +19,34 @@
...
+		<div
+			class="{meterBase} {meterBg} {meterRounded} {meterTransition} {api.indeterminate ? meterAnimate : ''} {meterClasses}"
+			{...api.getRangeProps()}
```
endigo9740, line 49: Unfortunately everyone refers to this part of the interface different. We should at least consider if we stick with `meter` or if we follow Zag's convention of `range`.

endigo9740 (changes requested):
Just a few things after a quick review...

packages/skeleton-svelte/src/lib/components/Rating/RatingItem.svelte
```diff
@@ -0,0 +1,23 @@
+<script lang="ts">
```
endigo9740, line 1: I think there's a case to be made against dropping the child `<RatingsItem>` component. While I know Ratings can have multiple children - they're basically always clones of the same few combinations of icon states (ex: all stars).

I think we can follow the prior Skeleton iteration and keep a `max` prop, which generates the X iterations within a loop. Much like [Zag's own doc example](https://zagjs.com/components/react/rating-group#usage):

- https://github.com/skeletonlabs/skeleton/blob/next/packages/skeleton-svelte/src/lib/components/Rating/Rating.svelte#L6
- https://github.com/skeletonlabs/skeleton/blob/next/packages/skeleton-svelte/src/lib/components/Rating/Rating.svelte#L93

The edge case of someone wanting to implement custom icons per each and every child seems highly unlikely. And that's all the child components really buy us here.

TL;DR: K.I.S.S.

packages/skeleton-svelte/src/routes/components/ratings/+page.svelte
```diff
@@ -1,163 +1,122 @@
...
-			</div>
-			<h3 class="h3">5.0</h3>
-		</div>
-		<div class="flex items-center gap-8">
-			<div class="w-48">
-				<Rating value={0}>
-					{#snippet iconFull()}
-						<Star size={24} class="fill-surface-950-50" />
+					{#snippet iconHalf()}
+						<StarHalf size={24} class="fill-surface-950-50" />
```
endigo9740, line 22: You did a great job with the half stars, but I see the SVG fill isn't quite cutting it here for Lucide. We need a full outline and half fill. Sadly Lucide doesn't seem to a means to do that. [Maybe we can request that](https://github.com/lucide-icons/lucide/issues/new?assignees=&labels=%F0%9F%99%8C+icon+request&projects=&template=01_icon_request.yml)? Seems like an easy add.

Unless you can find some SVG trickery to only fill half of outlined star, we may need to implement our own custom SVGs here.

On that note, we did have a request on Discord to provide some generic star SVGs as built-in placeholders. Maybe we can find a generic enough set of icons to fulfill this purpose. I'm thinking two SVGs:

1. Full star - can customize the stroke/fill; can be used for full/empty states
2. Half Star - full star in shape for stroke, but fill only ever covers half

That should take care of all three states: full/empty/half. See me if I can clarify this at all!

endigo9740: Mock:

Screenshot 2024-08-05 at 3 20 31 PM

endigo9740 (changes requested):
@Hugos68 in addition to the items below, I've noted the individual star items are lacking a `cursor-pointer` style to indicate interaction. Zag doesn't implement this, but we should.

Make sure to set this conditionally only when not in a disable or read-only state.

packages/skeleton-react/src/lib/components/Segment/Segment.tsx
```diff
@@ -1,113 +1,102 @@
 'use client';
 
-import React, { createContext, useContext, useEffect, useState } from 'react';
+import React, { createContext, useContext, useId } from 'react';
```
Hugos68, line 3: Lets try to avoid using the `React` default import. So `React.FC` becomes `FC`.

packages/skeleton-react/src/lib/components/Segment/Segment.tsx
```diff
@@ -1,113 +1,102 @@
...
 	width = '',
 	classes = '',
+	// Indicator
+	indicatorBase = 'top-[var(--top)] left-[var(--left)] w-[var(--width)] h-[var(--height)]',
+	indicatorBg = 'preset-filled',
+	indicatorRounded = 'rounded',
+	indicatorClasses = '',
 	// Events
-	onChange,
+	onValueChange = noop,
```
Hugos68, line 33: Do we really want to explicitely define events like this when zag already does this for you in `...zagProps`? The only benefit we have is that it will be `onValueChange(value)` instead of `onValueChange({ value })`. Doesn't seem worth the trouble IMHO.

packages/skeleton-react/src/lib/components/Segment/Segment.tsx
```diff
@@ -1,113 +1,102 @@
...
 	// Set Context
-	const ctx = {
-		value,
-		name,
-		onSelectionHandler
-	};
+	const ctx = { api };
+
+	// Reactive
+	// FIXME: always returns vertical
```
Hugos68, line 54: The `orientation` prop does work, but `vertical` is the default value. So if you want a `horizontal` control you need to pass that value. `segmented controls` are a type of radio group and on the web the radio groups are by default `vertical`. So this is work as expected.

packages/skeleton-react/src/lib/components/Segment/Segment.tsx
```diff
@@ -1,113 +1,102 @@
...
-			onClick={onClickHandler}
-			type="button"
-			className={`${base} ${rxActive} ${classes}`}
-			title={title}
-			disabled={disabled}
-			data-testid="segment-item"
-		>
-			{/* Input */}
-			{selected ? <input type="hidden" name={ctx.name} id={id} value={value} /> : null}
+		<label key={value} {...ctx.api.getItemProps({ value: value })} className={`${base} ${classes}`} data-testid="segment-item">
```
Hugos68, line 92: `key` is only required for looping in react, I assume you copied this from the Zag example. You can remove the `key` property here.

packages/skeleton-react/src/lib/components/Segment/Segment.tsx
```diff
@@ -1,113 +1,102 @@
...
-	useEffect(() => {
-		setRxActive(selected ? active : hover);
-	}, [selected, active, hover]);
-
-	function onClickHandler() {
-		ctx.onSelectionHandler(value);
-		onClick?.(value);
-	}
+	const rxActive = ctx.api.value === value;
+	const rxActiveText = rxActive ? labelActiveText : '';
```
Hugos68, line 89: You can do:
```suggestion
	// Reactive
    const state = ctx.api.getItemState(zagProps);  
    const activeClasses = state.active ? labelActiveText : ''; 
```

packages/skeleton-react/src/lib/components/Segment/Segment.tsx
```diff
@@ -1,113 +1,102 @@
...
 	classes = '',
 	// Label
 	labelBase = 'pointer-events-none',
+	labelActiveText = 'preset-filled',
 	labelClasses = '',
-	// Events
-	onClick,
 	// Children
 	children
 }) => {
```
Hugos68, line 83: Let zag handle it by spreading rest props. This includes `value` and `disabled` props. Something you were wondering (disabling one specific item, instead of the whole segmented group).
```suggestion
const SegmentItem: React.FC<SegmentItemsProps> = ({
	// Root
	base = 'btn cursor-pointer z-[1]',
	classes = '',
	// Label
	labelBase = 'pointer-events-none',
	labelActiveText = 'preset-filled',
	labelClasses = '',
	// Children
	children,
    // Zag
    ...zagProps
}) => {
```

Hugos68 (commented):
Functionally this works great, just some small things internally. I also noticed there were no focus styles when using keyboard nav to navigate the segmented control. This should really be present because I was lost on where my focus was.

packages/skeleton-svelte/src/routes/components/ratings/+page.svelte
Hugos68: I've done 3 SVG's for now because I simply couldn't get it working how I wanted. I've done the rest afaik. You may review and change if needed.

packages/skeleton-react/src/lib/components/Segment/Segment.tsx
endigo9740: I was using the Accordion as reference, which I think still utilizes a custom `onValueChange`. Maybe double check that?

packages/skeleton-react/src/lib/components/Segment/Segment.tsx
endigo9740: Oh boy, that's confusing since the default state of this component is horizontal. Would that mean we need to create our own custom prop and pass that to the machine?

packages/skeleton-react/src/lib/components/Segment/Segment.tsx
endigo9740: ...this is already done, right?

packages/skeleton-react/src/lib/components/Segment/Segment.tsx
endigo9740: I think we've bounced back and forth on this between all components, so we should standardize. I'm fine with that approach.

packages/skeleton-react/src/lib/components/Segment/Segment.tsx
Hugos68: You have `value` defined, and not `...zagProps`

endigo9740: Oh on the `SegmentItem`. Yeah I wasn't sure if we wanted to handle that with the "overflow" zag components, as it's the only prop we're using in the items. What do you think?

endigo9740: I don't think Zag supports `disabled` state for individual items. Only for the entire group, which is already handled in `<Segment>`:

- https://zagjs.com/components/react/segmented-control#disabling-the-segmented-control
- https://zagjs.com/components/react/segmented-control#making-the-segmented-control-readonly

This is set on the `machine`, which is only initialized per the overall "group". Unless I'm misunderstanding the docs.

Hugos68: It does support it but it's undocumented, I tried it out when I was reviewing your code.
image

Hugos68: `value` is required, `disabled` is what you wanted, `invalid` is something I'm not familiar with.

endigo9740: How do I apply this within the `<SegmentItem>` though? There's no `machine` per item? The only data we're passing down from parent -> child component is the `api`. Can I provide these values to `api` somehow?

Hugos68: Whenever you do `{ value }` just replace that with `zagProps`

packages/skeleton-react/src/lib/components/Segment/Segment.tsx
endigo9740: We really need `horizontal` to be the default, so I'm going to opt to a custom `orientation` prop the sets this as the default.

packages/skeleton-react/src/lib/components/Segment/Segment.tsx
endigo9740: Per Discord - by moving this to the overflow `zagProps` end users run into two issues:

1. They can no longer use the `useState` setter directly (since this returns `{ value }`
2. They would be required to set the return value to `ValueChangeDetails` - which comes from Zag

I've kept the custom prop for now unless you have a better idea.

packages/skeleton-react/src/lib/components/Segment/Segment.tsx
endigo9740: Per Discord, this introduces type errors when trying to use `zagProps`. I tried a few options but no luck. Would welcome your suggestions for how to resolve this.

packages/skeleton-react/src/lib/components/Segment/Segment.tsx
endigo9740: Same type errors described here: https://github.com/skeletonlabs/skeleton/pull/2778#discussion_r1707673468

packages/skeleton-svelte/src/routes/components/ratings/+page.svelte
endigo9740: Looks good. I'll generate new artwork for these asap.

packages/skeleton-react/src/lib/components/Segment/Segment.tsx
Hugos68: Yeah this was my concern aswell, we can just use a custom callback indeed. I'll resolve this, was just wondering about this.

endigo9740:
@hugos68 I've pushed my changes. The `disabled` and `readme` modes should work as you described above. Just with three Svelte-specific bugs:

1. Using `ctx.api.getItemState(zagProps)` has a type error not present in React (see the FIXME comment)
2. Using `state.checked` doesn't work (see the FIXME comment)
3. Multiple components that bind to the same `$state()` rune do not stay in sync. Try the `sm/md/lg` example for React and Svelte to see what I mean.

endigo9740:
@Hugos68 logging this so we don't miss it - but we'll need one of the following changes for Astro to return to a working state:

1. Roll back to [these dependency versions](https://github.com/skeletonlabs/skeleton/pull/2783#issuecomment-2258972851), which will break support for `createRawSnippet()`. Used for Rating SVG icons.
2. Update to the latest version of Astro once they [merge this PR fix](https://github.com/withastro/astro/pull/11624#issuecomment-2273987633c).

packages/skeleton-svelte/src/lib/components/Rating/types.ts
```diff
@@ -1,70 +1,15 @@
...
-	/** Sets the maximum rating value. */
-	max?: number;
-	/** Sets interactive mode. */
-	interactive?: boolean;
-	/** Sets the rating fractional granularity. */
-	step?: number;
-
-	// Root ---
-	/** Sets base styles. */
+export interface RatingProps extends Omit<rating.Context, 'id'> {
```
endigo9740, line 4: Just a reminder to implement the comments and JSDocs documentation for both framework `types.ts` files.

packages/skeleton-svelte/src/lib/components/Rating/Rating.svelte
```diff
@@ -1,116 +1,73 @@
...
-		// Snippets ---
-		iconEmpty,
-		iconFull
+		base = 'flex gap-1',
+		classes,
+		// Label
+		labelBase,
+		labelClasses,
+		// Item
+		itemBase = 'cursor-pointer data-[disabled]:cursor-not-allowed data-[readonly]:cursor-default',
```
endigo9740, line 18: I might suggest reviewing how I handle `disabled` and `readonly` states styles for the Segment Control. I break them to individual style props so users can easily overwrite them. I also use the `.disabled` class which is a Skeleton-provided utility class.

There should be a visible difference between each state:

1. Default state - fully visible; fully interactive
2. Disabled state - half transparent; non-interactive
3. Read-Only state - fully visible; non-interactive

packages/skeleton-svelte/src/lib/components/Rating/Rating.svelte
```diff
@@ -1,116 +1,73 @@
...
-			class:pointer-events-none={!interactive}
-			tabindex={interactive && i === focusedButtonIndex ? 0 : -1}
-			onmousedown={(event: MouseEvent) => interactive ? onRatingMousedown(event, i) : undefined}
-			onkeydown={(event: KeyboardEvent) => interactive ? onRatingKeydown(event) : undefined}
-			type="button"
-		>
-			<!-- Icon: Empty -->
-			<span class="{emptyBase} {emptyClip} {rxEmptyInteractive} {emptyClasses}" style="--clipValue: {(value - i) * 100}%">
-				{#if iconEmpty}
+<div {...api.getRootProps()}>
```
endigo9740, line 52: Make sure you're adding HTML comments within the template to denote each section. This helps match which style props goes to which portion of the template.

packages/skeleton-react/src/lib/components/Rating/Rating.tsx
```diff
@@ -1,155 +1,72 @@
...
+				<label className={`${labelBase} ${labelClasses}`} {...api.getLabelProps()}>
+					{label}
+				</label>
+			)}
+			<div className={`${base} ${classes}`} {...api.getControlProps()}>
+				{api.items.map((index) => {
+					const itemState = api.getItemState({ index });
+
+					const icon = (() => {
+						if (!itemState.highlighted) {
```
endigo9740, line 53: Consider a switch/case here if JSX allows it.

packages/skeleton-react/src/lib/components/Rating/Rating.test.tsx
```diff
@@ -1,204 +0,0 @@
-import { render, fireEvent } from '@testing-library/react';
```
endigo9740, line 1: Make sure to reimplement the test spec file, even if it's just a single test case for now. I'd understand if the changes are dramatic enough to warrant us starting from scratch. But I want to ensure we have the test spec in place when we need it.

endigo9740 (changes requested):
@Hugos68 I submitted some changes to to the sandbox doc pages and minor comment changes to the components directly. But otherwise these should be the only items remaining.

packages/skeleton-svelte/src/lib/components/Rating/Rating.svelte
Hugos68: That's a wondeful idea yeah, let me actually do that.

packages/skeleton-react/src/lib/components/Rating/Rating.test.tsx
Hugos68: What do we test? Since zag already tests all the logic. Ensuring class x is applied and such?

endigo9740: * Barebones "does it work" with minimal props
* Anything Skeleton specific
* I'd recommend some light tests for style props like `base/classes`

Hugos68: Thanks, I'll be sure to test this 👍

packages/skeleton-react/src/lib/components/Rating/Rating.tsx
Hugos68: JSX allows it but a switch isn't applicable here, since we're checking different types of states, not the same state that could have different values.

Hugos68:
> @Hugos68 I've pushed my changes. The `disabled` and `readme` modes should discussed.
> 
> Note I'm running into three Svelte-specific bugs in `Segment.svelte`:
> 
> 1. Using `ctx.api.getItemState(zagProps)` has a type error not present in React (see the FIXME comment)
> 2. Using `state.checked` doesn't work (see the FIXME comment)
> 3. ~Multiple components that bind to the same `$state()` rune do not stay in sync. Try the `sm/md/lg` example for React and Svelte to see what I mean.~
> 
> > #3 was fixed by moving `value` to the Zag `context: {}`

All of these are solved now:

The first issue had a type error because there was an extra type `labelStatusText` that wasn't used and thus was added to the `...zagProps` since it wasn't explicitely destructured. I didn' see this prop anywhere in the React implementation so I removed it entirely.

The second issue was because we weren't using `$derived` so state got out of sync. In React you ofcourse don't have to do this because in React the entire component reruns on *every* rerender.

bartduisters:
I've done some clicking through, here are some things I've noticed:

Accordion (React):
Screenshot 2024-08-11 at 14 39 26

Avatars (React and Svelte) filter example on Safari (I know it's an example, but it might also be used as an example on the docs page - something to keep an eye out for)
Screenshot 2024-08-11 at 14 31 28

Progess Rings: no comment

Segment Control (React and Svelte):
No indication which segment control is selected when pressing `tab`.
When you have tab focus and navigate with arrow up/down or arrow left/right, keep pressing left or right, it will not jump over from the last one to the first one (or from frist to last) on the first control. It will work on the second and third one.

Switch (React and Svelte)
No tab indication. When using keyboard navigation, there is no indication as to which one is focused.

packages/skeleton-react/src/lib/components/Progress/Progress.tsx
```diff
@@ -1,51 +1,59 @@
...
-	});
+	// Zag
+	const [state, send] = useMachine(progress.machine({ id: useId() }), { context: zagProps });
+	const api = progress.connect(state, send, normalizeProps);
 
-	const indeterminate = value === undefined;
-	const fillPercentage = `${indeterminate ? 50 : ((value! - 0) / (max - 0)) * 100}%`;
-	const rxIndeterminate = indeterminate ? meterAnimate : '';
+	// Reactive
+	const rxIndeterminate = api.indeterminate ? meterAnimate : '';
```
endigo9740, line 39: The default indeterminate mode is not working for either React or Svelte. Custom animations are though. Let me know if you need me to help circle back and help with this tomorrow.

sites/next.skeleton.dev/src/examples/components/rating/AllowHalf.svelte
```diff
@@ -0,0 +1,5 @@
+<script lang="ts">
```
endigo9740, line 1: Ratings examples do not follow the file naming convention. Should always be:

```
Example + {Description} + {extension}
```

Such as:

* Example.svelte | Example.tsx (the default)
* ExampleFoo.svelte | ExampleFoo.tsx
* ExampleBar.svelte | ExampleBar.tsx

This helps ensure don't have overlaps with custom components generated for the Svelte app. I know the path will differ, but this helps keep things explicit.

endigo9740 (changes requested):
@Hugos68 Per Discord, I push a number of changes on my own the Progress/Ratings components and docs. But here's the things I haven't go to yet. If you could please follow up asap.

NOTE: we should follow up on our discussion of the use `<figure>` tomorrow. This would impact a number of components if we wish to change it.

packages/skeleton-react/src/lib/components/Progress/Progress.tsx
Hugos68: Done.

sites/next.skeleton.dev/src/examples/components/rating/AllowHalf.svelte
Hugos68: Done.

packages/skeleton-react/src/lib/components/Accordion/Accordion.tsx
```diff
@@ -1,39 +1,29 @@
...
-	AccordionPanelProps,
-	AccordionProps
-} from './types.js';
+import type { AccordionContextState, AccordionControlProps, AccordionItemProps, AccordionPanelProps, AccordionProps } from './types.js';
+import { noop } from '../../internal/noop.js';
 
 // Contexts ---
 
 export const AccordionContext = createContext<AccordionContextState>({
 	animDuration: 0.2,
```
bartduisters, line 14: AccordionContextState has a comment that `animDuration` should be in milliseconds. The Svelte implementation passes `200`.

packages/skeleton-react/src/lib/components/Accordion/Accordion.tsx
```diff
@@ -1,39 +1,29 @@
...
 	iconOpen: '-',
```
bartduisters, line 15: iconOpen and iconClose are not options in the Svelte API, is this intentional?

packages/skeleton-react/src/lib/components/Accordion/Accordion.tsx
```diff
@@ -46,111 +36,104 @@ const AccordionRoot: React.FC<AccordionProps> = ({
...
-		animDuration,
-		iconOpen,
-		iconClosed,
-		open,
-		close,
-		toggle,
-		isOpen
-	};
+	// Zag
+	const [state, send] = useMachine(
```
bartduisters, line 46: Just for consistency, I noticed that sometimes the first variable is named `state` and sometimes `snapshot`. I'd opt for one term for all useMachine() variables

packages/skeleton-react/src/lib/components/Rating/Rating.tsx
```diff
@@ -1,155 +1,93 @@
-'use client';
+import * as rating from '@zag-js/rating-group';
+import { useMachine, normalizeProps } from '@zag-js/react';
+import { useId, type FC } from 'react';
+import type { RatingProps } from './types';
+import { starEmpty, starHalf, starFull } from '../../internal/nodes';
+import { noop } from '../../internal/noop';
 
-import React, { useCallback, useEffect, useRef, useState } from 'react';
```
bartduisters, line 3: I've noticed two things while using keyboard navigation.

The first thing is: when no stars are selected, it is impossible to use tab to focus the rating stars (to then add a value with the keyboard).

The second thing is: there is an inconsistency between React and Svelte. Using the keyboard to add values, it will add half a star in React, and a whole star in Svelte.

packages/skeleton-react/src/lib/components/Accordion/Accordion.tsx
endigo9740: This matches Zag's implementation instructions. I noticed it too. We'll need to investigate if this is intentional.

packages/skeleton-react/src/lib/components/Tabs/Tabs.tsx
```diff
@@ -1,236 +1,157 @@
 'use client';
```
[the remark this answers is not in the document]
endigo9740: Yeah the Tabs and Accordion component structure differ because the implementation differs. This is due to the fact React doesn't have a direct equivalent for named slots. They have `React.Node` (passing template content via props) but that's not a great solution IMO. So we would have to worsen the experience with all non-JSX based frameworks in order to make these match.

Hugo and I have talked about this a bunch, and we want symmetry, but unfortunately I don't think the cost is worth it.

packages/skeleton-react/src/lib/components/Rating/Rating.tsx
Hugos68: That differs based on the `allowHalf` prop

packages/skeleton-react/src/lib/components/Accordion/Accordion.tsx
endigo9740: The default value supplied is correct, it's the JSdoc comment that's wrong. In Svelte we use the Svelte transitions system which takes values as milliseconds (ex: `200`), but in React we're leaning into Zag to control animations, which takes them as float values (ex: `0.2`). So I've updated the React JSdoc comment to make this more clear.

packages/skeleton-react/src/lib/components/Accordion/Accordion.tsx
endigo9740: They are options, they're just implemented with different tech per framework:

1. In Svelte we use snippets
2. In React we use React.Node (props that accept template data)

Both are accounted for and use the same `iconX` naming scheme in the data and list of props.

You may have missed it for Svelte because the props are implemented on the parent, the passed down to the items via the Context API. That way you only have to set them once.

packages/skeleton-react/src/lib/components/Rating/Rating.tsx
endigo9740: As Hugo said, this is working per my expectation when I test on the React doc examples.

endigo9740: Per the two issues:

1. Confirmed - I can replicate this. It may be related to focus states which we plan to return to post-release, but I'll log this specifically so it's not lost in the shuffle.
2. What Hugo said, the stars work per my expectations based on the `allowHalf` setting on the doc examples.

endigo9740: Per 1, I've added this to the list in the original post above. I'll be making dedicated tickets for these on GitHub soon.

PR #2786, opened 2024-08-01
NEXT: Cookbook Image Layouts

kmalloy24:
## Description

Adds image layouts from v2 Blocks to v3 Cookbook. Includes two new layouts,
- Quad
- Featured

The page seems laggy on my local dev, it might be all the image URLs loading at full size. I can implement some image optimization features from Astro, but was trying to minimize any maintainable image assets in the codebase for now. It would also complicate the interaction between example components and raw code if we are maintaining a different component for the docs than what the example raw code is showing.

### DIFF
--- a/sites/next.skeleton.dev/src/examples/resources/cookbook/image-layouts/ExampleFeatured.astro
+++ b/sites/next.skeleton.dev/src/examples/resources/cookbook/image-layouts/ExampleFeatured.astro
@@ -1,0 +1,28 @@
+<!-- Flex -->
+<section class="flex justify-center">
+	<!-- Grid -->
+	<div class="grid gap-4">
+		<!-- Featured -->
+		<div>
+			<img class="h-auto max-w-full rounded-lg" src="https://images.unsplash.com/photo-1597077962467-be16edcc6a43" alt="" />
+		</div>
+		<!-- Row -->
+		<div class="grid grid-cols-5 gap-4">
+			<div class="w-86 h-48">
+				<img class="h-full w-full rounded-lg object-cover" src="https://images.unsplash.com/photo-1620005839871-7ac4aed5ddbc" alt="" />
+			</div>
+			<div class="w-86 h-48">
+				<img class="h-full w-full rounded-lg object-cover" src="https://images.unsplash.com/photo-1553184570-557b84a3a308" alt="" />
+			</div>
+			<div class="w-86 h-48">
+				<img class="h-full w-full rounded-lg object-cover" src="https://images.unsplash.com/photo-1509130446053-899ae7358ce6" alt="" />
+			</div>
+			<div class="w-86 h-48">
+				<img class="h-full w-full rounded-lg object-cover" src="https://images.unsplash.com/photo-1510111652602-195fc654aa83" alt="" />
+			</div>
+			<div class="w-86 h-48">
+				<img class="h-full w-full rounded-lg object-cover" src="https://images.unsplash.com/photo-1597077917598-97ca3922a317" alt="" />
+			</div>
+		</div>
+	</div>
+</section>
--- a/sites/next.skeleton.dev/src/examples/resources/cookbook/image-layouts/ExampleFeatured.astro
+++ b/sites/next.skeleton.dev/src/examples/resources/cookbook/image-layouts/ExampleFeatured.astro
@@ -1,28 +1,14 @@
-<!-- Flex -->
-<section class="flex justify-center">
-	<!-- Grid -->
-	<div class="grid gap-4">
-		<!-- Featured -->
-		<div>
-			<img class="h-auto max-w-full rounded-lg" src="https://images.unsplash.com/photo-1597077962467-be16edcc6a43" alt="" />
-		</div>
-		<!-- Row -->
-		<div class="grid grid-cols-5 gap-4">
-			<div class="w-86 h-48">
-				<img class="h-full w-full rounded-lg object-cover" src="https://images.unsplash.com/photo-1620005839871-7ac4aed5ddbc" alt="" />
-			</div>
-			<div class="w-86 h-48">
-				<img class="h-full w-full rounded-lg object-cover" src="https://images.unsplash.com/photo-1553184570-557b84a3a308" alt="" />
-			</div>
-			<div class="w-86 h-48">
-				<img class="h-full w-full rounded-lg object-cover" src="https://images.unsplash.com/photo-1509130446053-899ae7358ce6" alt="" />
-			</div>
-			<div class="w-86 h-48">
-				<img class="h-full w-full rounded-lg object-cover" src="https://images.unsplash.com/photo-1510111652602-195fc654aa83" alt="" />
-			</div>
-			<div class="w-86 h-48">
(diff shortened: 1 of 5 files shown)


endigo9740:
@kmalloy24 this is looking great! Just FYI you might can use Unsplash's CDN features to resize the images:

```
https://images.unsplash.com/photo-1617296538902-887900d9b592?ixid=M3w0Njc5ODF8MHwxfGFsbHx8fHx8fHx8fDE2ODc5NzExMDB8&ixlib=rb-4.0.3&w=128&h=128&auto=format&fit=crop
```

You'll note several URL params added at the end here:

```
&w=128
&h=128
&auto=format
&fit=crop
```

* w = width
* h = height

You might tailor those *per* example to ensure they are no larger than the need to be.

We'll want to try to avoid Astro-only optimization if possible, to keep this as generic as possible.

kmalloy24:
I had removed the Unsplash params to make the image sizing dependent on Tailwind classes instead. This way folks can just swap in their image sources without the need for URL params to maintain the layouts. I can go back to the previous approach - it was just a little confusing to me when I copied out the code block and lost the formatting when I put in different image urls. 

I'll check out place-hold.it and see if it helps make the page more lightweight.

sites/next.skeleton.dev/src/examples/resources/cookbook/image-layouts/ExampleGrid.astro
```diff
@@ -0,0 +1,36 @@
+<!-- Flex -->
+<section class="flex justify-center">
+	<!-- Grid -->
+	<div class="grid grid-cols-2 gap-6 md:grid-cols-3">
+		<!-- Row -->
+		<div class="h-48 w-48">
+			<img class="h-full w-full rounded-lg object-cover" src="https://images.unsplash.com/photo-1617296538902-887900d9b592" alt="" />
```
endigo9740, line 7: Per my previous note, let's reintroduce the Unsplash URL parameters to resize the images. Think of it like this:

- The params control the size of the image delivered from the CDN - which affects the file size - which affects the speed at which these load
- The Tailwind classes affect the display size.

sites/next.skeleton.dev/src/examples/resources/cookbook/image-layouts/ExampleGrid.astro
```diff
@@ -0,0 +1,36 @@
...
+		</div>
+		<div class="h-48 w-48">
```
endigo9740, line 9: Note, Tailwind now has a `size` property as a shorthand when height/width match. This is relatively new:
https://tailwindcss.com/docs/size

Feel free to move this to the `` itself to remove the extra wrapping `<div>`.

sites/next.skeleton.dev/src/examples/resources/cookbook/image-layouts/ExampleGrid.astro
```diff
@@ -0,0 +1,36 @@
...
+			<img class="h-full w-full rounded-lg object-cover" src="https://images.unsplash.com/photo-1597077962467-be16edcc6a43" alt="" />
```
endigo9740, line 10: Let's make sure we're populating `alt` tags to follow best practices for accessibility.

sites/next.skeleton.dev/src/examples/resources/cookbook/image-layouts/ExampleFeatured.astro
```diff
@@ -0,0 +1,28 @@
+<!-- Flex -->
+<section class="flex justify-center">
```
endigo9740, line 2: The use of `section > div` seems redundant. I'd recommend we keep the `section` as the parent, and move the grid style up to here.

sites/next.skeleton.dev/src/examples/resources/cookbook/image-layouts/ExampleFeatured.astro
```diff
@@ -0,0 +1,28 @@
...
+	<!-- Grid -->
+	<div class="grid gap-4">
+		<!-- Featured -->
+		<div>
```
endigo9740, line 6: Optional, but recommended: this `<div>` seems like a prime candidate to use `<header>`

sites/next.skeleton.dev/src/examples/resources/cookbook/image-layouts/ExampleQuad.astro
```diff
@@ -0,0 +1,18 @@
+<!-- Flex -->
+<section class="flex justify-center">
```
endigo9740, line 2: Similar to Featured, let's move the grid styles up to the wrapping `section` and remove this extra dive. It's redundant.

endigo9740 (changes requested):
@kmalloy24 I had a chance to do a quick review and everything is looking excellent! Here's some small items when you get a chance.

--

Note that while we're not using using `<Image>` components in our examples, it wouldn't hurt to mention that these exist. Perhaps as part of a small blurb near the bottom of the page. Note that we do NOT need to teach folks how to use these tools, they're outside the scope of Skeleton.

You can reference how we're handling this for Avatars in an upcoming update:

- Next.js: https://skeleton-next-8ur332lry-skeleton-labs.vercel.app/docs/components/avatar/react#fallback
- SvelteKit: https://skeleton-next-8ur332lry-skeleton-labs.vercel.app/docs/components/avatar/svelte#fallback

I might suggest something like this:

```
## Image Components

Some meta-frameworks implement components or tooling to improve the performance of images. In many cases these can be using instead of raw `<img>` tags to improve performance. This includes:

* Next.js Image Component (as a link)
* SvelteKit Enhanced:Image (as a link)
```

> NOTE: in the future we expand with Nuxt/Astro/etc...

endigo9740:
Note I've bumped this back to draft state until the above work is complete. When ready for another review, feel feel to select that option again please!

endigo9740:
@kmalloy24 just checking on the status of this. No rush, I just want to confirm you saw the requested changes above.

kmalloy24:
@endigo9740 I did, thanks for taking the time to review! Was hoping to get to it this week, but realistically I probably won't get a chance until next week.

endigo9740:
@kmalloy24 I finally had a moment free so I did a quick pass through this section. I made three notable changes:

1. Made a couple minor edits to the text copy on the main guide page
2. Updated each example to use [Lorem Picsum](https://picsum.photos/)
3. Cleaned up the styles within each example to only use what we need
4. Swapped the fixed rounded class to Skeleton's version which will adhere to your theme

The guide should now be complete, so I'll merge right away. Thanks for your contribution as always!

kmalloy24:
@endigo9740 Sweet! TIL about Lorem Picsum...will definitely be using that in the future. 

This was a good experience getting my feet wet with v3 so hopefully will be more efficient on cookbook contributions in the future.

endigo9740:
@kmalloy24 oh yeah it's cool. If you jump on another PR (no rush) just a heads up that we're about to push a change that will make PNPM v9.50 the minimum version (9.7.x is the latest and that's fine too). I'm trying to give all the regular contributors a heads up about this one:

- https://github.com/skeletonlabs/skeleton/pull/2812

PR #2799, opened 2024-08-11
use current svelte-add syntax to install tailwind

### LINKED ISSUE #2798: installation instructions for next recommend a deprecated package
### Link to the Page

https://next.skeleton.dev/docs/get-started/installation/sveltekit#install-tailwind

The next instructions for SvelteKit recommend using `npx @svelte-add/tailwindcss@latest --typography false` but the https://github.com/svelte-add/tailwindcss package is deprecated:

```
% npx @svelte-add/tailwindcss@latest --typography false

Need to install the following packages:
@svelte-add/tailwindcss@1.0.17
Ok to proceed? (y) 

npm warn deprecated @svelte-add/tailwindcss@1.0.17: Package no longer supported.
Please use npx svelte-add tailwindcss
```

`npx svelte-add@latest tailwindcss` works as expected.

The [live instructions](https://www.skeleton.dev/docs/get-started#install-tailwind) look correct, it's just the `next` branch that's using the deprecated package.

iain3000:
## Linked Issue

Closes #2798 

## Description

The @svelte-add/tailwindcss package is deprecated and recommends using `svelte-add tailwindcss`: https://github.com/svelte-add/tailwindcss

### DIFF
--- a/sites/next.skeleton.dev/src/content/docs/get-started/installation/sveltekit.mdx
+++ b/sites/next.skeleton.dev/src/content/docs/get-started/installation/sveltekit.mdx
@@ -41,8 +41,8 @@ import ProcessStep from '@components/docs/ProcessStep.astro';
         ## Install Tailwind
         Install Tailwind using [Svelte-Add](https://svelte-add.com/) to automate the process.
         ```console
-        npx @svelte-add/tailwindcss@latest --typography false
-        npm install
+        npx svelte-add@latest tailwindcss
+        # choose to not install typography
         ```
         > NOTE: if you see "something went wrong", everything will work as expected. Please proceed to the next step.
     </ProcessStep>
--- a/sites/next.skeleton.dev/src/content/docs/get-started/installation/sveltekit.mdx
+++ b/sites/next.skeleton.dev/src/content/docs/get-started/installation/sveltekit.mdx
@@ -41,8 +41,7 @@ import ProcessStep from '@components/docs/ProcessStep.astro';
         ## Install Tailwind
         Install Tailwind using [Svelte-Add](https://svelte-add.com/) to automate the process.
         ```console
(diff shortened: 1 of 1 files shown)


sites/next.skeleton.dev/src/content/docs/get-started/installation/sveltekit.mdx
```diff
@@ -41,8 +41,8 @@ import ProcessStep from '@components/docs/ProcessStep.astro';
         ## Install Tailwind
         Install Tailwind using [Svelte-Add](https://svelte-add.com/) to automate the process.
         ```console
-        npx @svelte-add/tailwindcss@latest --typography false
-        npm install
+        npx svelte-add@latest tailwindcss
+        # choose to not install typography
```
endigo9740, line 45: Thanks for the submission @iain3000. We need one of two things here:

1. We need the flag to automatically skip this.
2. Otherwise we should skip the comment. This might make it sound like this incompatible, which is not the case.

A lot of folks just have a misconception about what this plugin is and does so we typically try to avoid it.

endigo9740 (changes requested):

sites/next.skeleton.dev/src/content/docs/get-started/installation/sveltekit.mdx
iain3000: so I think the svelte-add docs are a bit inconsistent on the syntax for passing options to adders; the website still lists `--typography false` while the cli command itself lists `--tailwindcss-typography false`.

The former doesn't currently work, while the latter does so that's what I've amended it to.

endigo9740: Perfect, thanks for the quick turn around!

PR #2912, opened 2024-10-25
adjusted order of head tags

LukeHagar:


endigo9740:
@LukeHagar as discussed on Discord. I'd normally go for this change, but when it comes to Analytics specifically, we typically want those to be the last items loaded. That helps ensure page state is "finalized" before new metrics are recorded. We'll keep this as is for now. But thanks for the PR!

PR #2941, opened 2024-11-09
Fix: Build error on Vercel

Hugos68:
Fixes build issue on Vercel

### DIFF
--- a/sites/next.skeleton.dev/tailwind.config.js
+++ b/sites/next.skeleton.dev/tailwind.config.js
@@ -1,3 +1,6 @@
+// import { skeleton } from '../../packages/skeleton/dist/plugin/index.cjs'
+// NOTE: Do not delete the above comment. It's required for local HMR on plugin changes.
+
 import { join, dirname } from 'node:path';
 import { fileURLToPath } from 'node:url';
 import { skeleton } from '@skeletonlabs/skeleton/plugin';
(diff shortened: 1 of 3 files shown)


sites/next.skeleton.dev/tailwind.config.js
```diff
@@ -1,27 +1,26 @@
-// import { skeleton } from '../../packages/skeleton/dist/plugin/index.cjs'
```
endigo9740, line 1: We'll need to keep these comments or it may break HRM

sites/next.skeleton.dev/tailwind.config.js
```diff
@@ -1,27 +1,26 @@
...
+		join(__dirname, 'node_modules/@skeletonlabs/skeleton-svelte/**/*.{html,js,ts,svelte}')
 	],
 	theme: {
 		extend: {}
 	},
 	plugins: [
 		forms,
-		// typography,
 		skeleton({
-			themes: Object.values(themes) // all available themes
```
endigo9740, line 24: We provide our apps as reference for others. Would you mind keeping this comment to explain what this is doing.

endigo9740 (changes requested):

sites/next.skeleton.dev/tailwind.config.js
Hugos68: Added back.

sites/next.skeleton.dev/tailwind.config.js
Hugos68: Added it back.

Hugos68:
We should test if the commented out import statement is still required as that looks to only be needed for the `.cjs` file and in #2931 we started outputtin both ESM and CJS.

PR #2967, opened 2024-11-14
Make Svelte's Switch component in sync with React version

### LINKED ISSUE #2882: Switch `onchange` errors
### Current Behavior

Documentation states that [Switch](https://next.skeleton.dev/docs/components/switch/svelte) can be used with `onchange` property. However, currently there are two issues regarding it:

1. The property does not exist in the elements types.
2. Adding the property causes a slew of the following warnings: 
image

The function instead of running returns the same warning whenever the switch state changes. `on:change` does not throw a type error, but does not do anything, not even the aforementioned warning

### Expected Behavior

`onchange` working.

### Steps To Reproduce

Copy paste what the documentation says in the **Lightswitch** example.

https://github.com/MrVauxs/skeletonrepro

### More Information

Code_CcSGg3Kz1L
chrome_NGdFYeAQ5r

phamduylong:
## Linked Issue

Closes #2882

## Description

- Updates the way `onChangeChecked` is implemented in the Svelte Switch component
- Updates the Svelte Switch components docs - which has an incorrect event listed in the examples

## Checklist

## Changsets


endigo9740:
Thanks @phamduylong, but can you elaborate on this point:

> The Svelte version of the Switch component seems to not do anything previously

The component seems functional on the documentation, so I'm not sure I follow:
https://next.skeleton.dev/docs/components/switch/svelte

Plus the issue you've linked to was referring to console log warning. Can you confirm the goal of this PR and that this does resolve those errors please.

Additionally, since this issue touches a project in `/packages`, you are required to submit a Changeset per the instructions provided in the original message above. Make sure to include that please.

phamduylong:
@endigo9740 Yeah I completely forgot that. The console warnings I was able to do nothing about those, the attempt was just made to have onCheckedChange to be available as props. Sorry for the confusion.

phamduylong:
For some reason I don't encounter the console warnings anymore after testing this again with the [repo provided in the linked issue](https://github.com/MrVauxs/skeletonrepro). And a changeset is now provided.

packages/skeleton-svelte/src/lib/components/Switch/types.ts
```diff
@@ -82,4 +82,7 @@ export interface SwitchProps extends Omit<zagSwitch.Context, 'id'> {
 	inactiveChild?: Snippet;
 	/** The active state snippet. */
 	activeChild?: Snippet;
+	// Events ---
+	/** Triggers when the switch is toggled. */
+	onCheckedChange?: (value: boolean) => void;
```
Hugos68, line 87: Why are we manually handling `onCheckedChange` instead of letting zag handle it? This would make the handler inconsistent with the rest of the components as they all delegate events to zag.

packages/skeleton-svelte/src/lib/components/Switch/Switch.svelte
```diff
@@ -54,9 +56,7 @@
 		zagSwitch.machine({
 			id: useId(),
 			name,
-			onCheckedChange(details) {
-				checked = details.checked;
```
Hugos68, line 58: Removing this line means the `bind:checked` no longer will work as it's never updated. Please revert.

Hugos68 (changes requested):

packages/skeleton-svelte/src/lib/components/Switch/types.ts
phamduylong: Not super certain how Zag works but it seems that when passing a customer handler it doesn't work without triggering it like this.

endigo9740: @Hugos68 this is why this one has been pending for a bit - I need to run through and verify if perhaps Svelte was doing it right and React doing it wrong. As much as possible we want to get out of the way and let Zag do it's thing.

Hugos68: @phamduylong I will test it out, but other components don't have these issues so my thought is either it *does* work and you had something misconfigured or there is a bug in `zag` which has happened before as the zag Svelte package is relatively new.

Hugos68: Ok after doing some tests I've discovered 2 things:
1. `onCheckedChange` works fine, so this is not something we need to touch.
2. `bind:checked` however currently does *not* update whenever the switch is checked or unchecked, so this will need to be fixed.

phamduylong: @Hugos68 I'll give it a look

Hugos68: @phamduylong FYI I've just fixed a similar issue in #2970, if you want to use that as a reference.

endigo9740: @phamduylong @Hugos68 I've merged Hugo's change. If you want to update your branch with the latest changes, this is relevant in two ways:

1. You can implement Hugo's suggested change
2. We merged #2968 today, which enables the ability to add the test case Hugo suggested.

I'd welcome both. I'll also keep Hugo as a reviewer. He might be able to review before I can (which will be Monday).

phamduylong: I rewrote the whole thing and added a test case for it. Hopefully behaves as it should now.

packages/skeleton-svelte/src/lib/components/Switch/Switch.svelte
```diff
@@ -53,14 +52,15 @@
...
-			}
+			name
 		}),
 		{
 			context: {
 				...zagProps,
+				onCheckedChange: (details) => {
+					checked = details.checked;
```
Hugos68, line 61: Please swap the order of operations around, first invoke the callback, then sync the binding. This needs to be consistent across all the components. For the rest this looks good!

Hugos68 (changes requested):

Hugos68 (approved):

PR #2994, opened 2024-11-21
Solved the feature request to hide first/last page button if page length is less then or 2 

### LINKED ISSUE #2993: Option to hide first/last buttons for Pagination's alternative UI

The first and last page buttons here should be optional. 
image

Also, maybe consider automatically hiding them when there are only two pages.

None

_No response_

rxhul18:
Added the feature to hide first/last page button if page length is less then or 2

Closes #2993 

## TestList
If there is only 2 or less page like in example, First/Last button is hidden
image
If there is above then 2 page like in example, First/Last button is active
image

### DIFF
--- a/packages/skeleton/src/lib/components/Paginator/Paginator.svelte
+++ b/packages/skeleton/src/lib/components/Paginator/Paginator.svelte
@@ -193,7 +193,7 @@
 			<button
 				type="button"
 				aria-label={labelFirst}
-				class={buttonClasses}
+				class="{buttonClasses} {lastPage <= 1 ? '!hidden' : ''}"
 				on:click={() => {
 					gotoPage(0);
 				}}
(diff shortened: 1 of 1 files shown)


endigo9740:
@rxhul18 so this great, but this feature was already implemented and available for Skeleton v2.

Screenshot 2024-11-21 at 9 33 49 AM

Source: https://www.skeleton.dev/components/paginators?tab=properties

The issue you were responded to was for Skeleton v3, where the components have been reimplemented from the ground up and don't always have v2 parity yet. Even the screenshots from the reference issue were from the v3 documentation.

I'm sorry but I'll have to deny this. If you would like to submit a similar change for v3, let me know here or in the ticket. Otherwise I will remove you as the owner of the ticket.

endigo9740:
@rxhul18 see my message above, I already did.

rxhul18:
Yess, My bad @endigo9740 . It was my early stage in Open-source. I will solve the v3 issue!

PR #2998, opened 2024-11-23
Fix/combobox-options-button-type

### LINKED ISSUE #2962: Combobox Submits Forms
### Current Behavior

Combobox uses buttons (in comparison, [Zag seems to use list elements with listeners on them](https://zagjs.com/components/react/combobox#methods-and-properties)) which do not have the role of `button`, causing form submissions.

### Expected Behavior

Expected not to submit forms.

### Steps To Reproduce

_No response_

_No response_

### More Information

Hugo confirming the issue on Discord: https://discord.com/channels/1003691521280856084/1191790107867488316/1303465018653937685

phamduylong:
## Linked Issue

Closes #2962

## Description

Set Svelte's Combobox options to button type. [The default type was submit](https://www.w3schools.com/jsref/prop_pushbutton_type.asp) and it caused undesired behavior with forms (see linked issue) 

## Changsets

## Checklist

### DIFF
--- a/packages/skeleton-svelte/src/lib/components/Combobox/Combobox.svelte
+++ b/packages/skeleton-svelte/src/lib/components/Combobox/Combobox.svelte
@@ -93,7 +93,7 @@
 			<!-- Input -->
 			<input {...api.getInputProps()} class={inputGroupInput} />
 			<!-- Arrow -->
-			<button {...api.getTriggerProps()} class={inputGroupButton} type="button">
+			<button {...api.getTriggerProps()} class={inputGroupButton}>
 				{#if arrow}
 					{@render arrow()}
 				{:else}
@@ -126,7 +126,8 @@
 						{@const isChecked = api.getItemProps({ item })['data-state'] === 'checked'}
 						{@const displayClass = isChecked ? optionActive : optionHover}
 						<!-- Option -->
-						<button {...api.getItemProps({ item })} class="{optionBase} {displayClass} {optionClasses}">
+						<!-- ZagJs should have set button type to "button" here. See https://github.com/skeletonlabs/skeleton/pull/2998#discussion_r1855511385 -->
+						<button {...api.getItemProps({ item })} class="{optionBase} {displayClass} {optionClasses}" type="button">
 							{item.label}
 						</button>
 					{/each}
--- a/packages/skeleton-svelte/src/lib/components/Combobox/Combobox.svelte
+++ b/packages/skeleton-svelte/src/lib/components/Combobox/Combobox.svelte
@@ -93,7 +93,7 @@
 			<!-- Input -->
 			<input {...api.getInputProps()} class={inputGroupInput} />
 			<!-- Arrow -->
-			<button {...api.getTriggerProps()} class={inputGroupButton}>
+			<button {...api.getTriggerProps()} class={inputGroupButton} type="button">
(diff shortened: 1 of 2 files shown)


phamduylong:
Seems like the React version of Combobox was done without much abstraction and it used Floating React (idk?). I assume it's the temporary version of the component so this much is enough.

packages/skeleton-svelte/src/lib/components/Combobox/Combobox.svelte
```diff
@@ -93,7 +93,7 @@
 			<!-- Input -->
 			<input {...api.getInputProps()} class={inputGroupInput} />
 			<!-- Arrow -->
-			<button {...api.getTriggerProps()} class={inputGroupButton}>
+			<button {...api.getTriggerProps()} class={inputGroupButton} type="button">
```
endigo9740, line 96: Long, actually after seeing this I was reminded Zag typically appends all the required attributes. So I looked and sure enough this type is already set, so this is redudant:

image

There much be another cause for why this is submitting a form. Feel free to research on your own, otherwise I'll need to circle back to this.

phamduylong: I'll look into it.

phamduylong: I actually don't have that. On docs site:

image

Could this be browser behavior? I'm using Chrome 131.0.6778.86 (Official Build) (64-bit) (cohort: Stable)

endigo9740: Ok I realized what's going on - I was inspecting the target input's button. You're reviewing the drop-down menu. What you're showing is populated in the "Content" section of the markup:

https://github.com/skeletonlabs/skeleton/blob/d0ee088343f8b59ffb6c29487c5a8c73d367048f/packages/skeleton-svelte/src/lib/components/Combobox/Combobox.svelte#L129

So i can confirm the type attribute is missing and will need to be appended by us for now. Really though, this issue should reported upstream and patch on Zag's end. Because this goes outside their standard behavior here. They apply the type for all other button elements.

phamduylong: Could you do the reporting upstream to Zag? My understanding to their product is quite limited so I will most likely not know what I'm talking about had I filed an issue. I'll put a comment to the mark that links to your comment.

endigo9740: Sure

Hugos68:
@endigo9740 We should not use a button here, Zag is correct to use an li here, not sure why we are diverting from Zag's recommendation and then be surprised when it doesn't work.

endigo9740:
@Hugos68 Li are not interactive elements, which means they are violating a11y best practices. I'm not surprised in the least that we had this issue, we just missed adding the type to prevent form submission. They way we've done it should be technically correct and this is one of a few things I'm going to report upstream to Zag when I get a chance.

Hugos68:
> Li are not interactive elements, which means they are violating a11y best practices.

This is not true because Zag sets both: `role="option"` and `tabindex="-1"`. Which follows the WACG combobox spec. I can assure you, you're wrong here.

PR #3008, opened 2024-11-28
Bugfix/combobox multiple display

### LINKED ISSUE #2963: Combobox does not show results with `multiple`
### Current Behavior

Combobox does not fill the input with the selected choices if `multiple` is enabled true. 

This also inadvertently causes `required` to always cause the input to fail the check.

### Expected Behavior

Combobox filling the input with the given array, would be nice to be able to provide a function for the separator used (so you can have something like `x, y, and z` instead of plain `.join(", ")`.

### Steps To Reproduce

_No response_

_No response_

### More Information

Earlier Discord Report for History: https://discord.com/channels/1003691521280856084/1191790107867488316/1303464712981184584

phamduylong:
## Linked Issue

Closes #2963 

## Description

In Zag's documentation:

image

This leave values to be handled on our own. `selectedLabels` does that. I provided a transform function as a prop in case user wanted to customize it.

I think this is an OK solution since the component is temporary. There are limitations when multiple is set now, such as manually wiping the input value (not using the dropdown) doesn't reset the value (if it's supposed to do so even).

## Changsets

## Checklist

### DIFF
--- a/.changeset/kind-bananas-perform.md
+++ b/.changeset/kind-bananas-perform.md
@@ -1,0 +1,5 @@
+---
+'@skeletonlabs/skeleton-svelte': patch
+---
+
+bugfix: display Svelte's Combobox value when `multiple` is enabled
--- a/packages/skeleton-svelte/src/lib/components/Combobox/Combobox.svelte
+++ b/packages/skeleton-svelte/src/lib/components/Combobox/Combobox.svelte
@@ -38,6 +38,8 @@
 		optionClasses = '',
 		// Snippets
 		arrow,
+		// Customization
+		transformFunction = (values) => (values ? values.join(', ') : ''),
 		// Zag ---
 		...zagProps
 	}: ComboboxProps = $props();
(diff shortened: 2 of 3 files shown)


endigo9740:
@Hugos68 would you mind doing a quick audit of this when you can. I want to verify this is the most Zag-friendly way to go about this. Feedback or suggestions welcome.

Hugos68:
Will do! 👍

endigo9740:
@phamduylong after thinking about this I don't think we'll move forward with this change. There's nothing wrong with your implementation, I just think from a UX perspective I don't think multi-selection is a great experience for our Svelte-only Combobox for now. There's essentially two ways to handle the display:

Example 1: using a comma separated list like so (which I believe you opted for):

image

Example: 2: Using an input with tags:

image

I really don't like the first option as it is extremely limited on space, even if the input expands. And the second should look familiar because we already have a primitive component for this interface - the [Tags Input component](https://next.skeleton.dev/docs/components/tags-input/svelte)!

I understand that it's a common user pattern to select multiple options from a canned list of options, but I'd rather we wait until we have Floating UI Svelte available to pair with the Tags Input, rather than re-inventing the Tags Input component within Combobox. Especially since it's only intended as a temporary solution. Basically the simplest solution is probably the best when it comes to any Svelte-specific Popover components.

Plus I'd rather not have a feature, then have it implement in a way I don't think it useful.

So for now I'm going to do do the following:

1. Close your PR
2. Update the ticket with the decision to remove the `multiple` option from Combobox
3. Then submit a follow up PR that omits the `multiple` prop from the component's type to disable this feature

phamduylong:
Understandable. Thanks for the heads up 😄

endigo9740:
@phamduylong yeah, thanks for your effort on this all the same. We always have this as reference if we change our mind in the future. Much appreciated sir!

PR #3024, opened 2024-12-05
NEXT fix empty switch padding (fixes #3010)

### LINKED ISSUE #3010: Next: switch has empty padding to the right
### Current Behavior

The `<Switch/>` component has a slight bit of padding to the right.

inspector
full element

### Expected Behavior

No ghost margin to the right.

### Steps To Reproduce

1. Navigate to https://next.skeleton.dev/docs/components/switch/svelte
2. Open the HTML inspector
3. target the switch element - it is off-center

https://next.skeleton.dev/docs/components/switch/svelte

### More Information

Seems to be caused by the `gap-4` tailwind class + the empty switch label `<span>` element.
Can be fixed by manually overriding `base="inline-flex items-center gap-2"` - `classes` does not work as there is no override handling done in the `Switch` component, leading to the `base` `gap-4` overriding.

SIMULATAN:
## Linked Issue

Fixes #3010 

## Description

Fixes the Switch component having some extra padding to the right when there is no child (they're shown as a label)

**before**
before, there is some extra padding that off-centers the switch and leaves some padding when aligned to the right

**after**
the switch is now centered or doesn't have empty space to the right

### DIFF
--- a/.changeset/weak-items-destroy.md
+++ b/.changeset/weak-items-destroy.md
@@ -1,0 +1,6 @@
+---
+'@skeletonlabs/skeleton-svelte': patch
+'@skeletonlabs/skeleton-react': patch
+---
+
+bugfix: empty padding when switch has no children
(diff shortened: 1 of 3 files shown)


endigo9740:
Awesome, thanks @SIMULATAN

Issue #2880, opened 2024-10-18
SEO and Meta Tags Updates


endigo9740:
### Link to the Page

https://next.skeleton.dev/

### Describe the Issue (screenshots encouraged!)

We'll need to make sure we update the doc to cover al required meta data, including:

- General meta tags for browser specific features
- SEO optimization
- Social Media sharing (Discord thumbnails, etc)
- etc...

endigo9740:
Dupe: https://github.com/skeletonlabs/skeleton/issues/3069

PR #3032, opened 2024-12-08
NEXT fix opengraph

SIMULATAN:
## Description

Fixes next.skeleton.dev previews on Discord just showing "Astro description"

discord preview

Additionally, I set the `og:image` property to the same value as in the v2 site - which is a hard-coded GitHub banner. Let me know in case we should revamp this (for example by passing the `LayoutDoc#docSite` to `LayoutRoot` to construct an absolute URL to the `/favicon.png`).

discord preview w/ image

## Help wanted
Checking the inspector, it appears the `LayoutDoc` doesn't work as expected. There's two `LayoutRoot` instances nested within each other, leading to the default page title being left in the `<head>`, with the correct one all the way down in the `body > div.container`.
Despite some debugging, I couldn't find out how to fix this - maybe someone knows?

HTML containing the default and correct <title>

### DIFF
--- a/sites/next.skeleton.dev/src/layouts/LayoutDoc.astro
+++ b/sites/next.skeleton.dev/src/layouts/LayoutDoc.astro
@@ -44,6 +44,7 @@ const { frontmatter, headings } = Astro.props satisfies Props;
 // Layout Props
 const layoutProps = {
 	title: frontmatter.title,
+	description: frontmatter.description,
 	classList: 'grid grid-rows-[auto_1fr] xl:grid-rows-[auto_1fr_auto]'
 };
 // GitHub Settings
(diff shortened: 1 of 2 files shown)


endigo9740:
Thanks @SIMULATAN, I'm aware of the issue and plan to fix soon via this issue. This will come with a number of other related changes to SEO and meta tags that should enable not just better text, but actual image previews as well. Potentially dynamic images via generated SVGs.

-https://github.com/skeletonlabs/skeleton/issues/2880

I may go ahead and merge this when ready to provide a temporary solution, but just know a more flushed out solution is coming. It's just the kind of thing I typically leave until near the end of the process.

endigo9740:
@SIMULATAN for the sake of time, and since we know a lot of this will be revised soon anyways, I'm going to go ahead and table this PR. I'll link this in the associated ticket I linked before as reference though:

I will personally handle this change and more in this upcoming ticket:

- https://github.com/skeletonlabs/skeleton/issues/2880

Thanks for the submission though!

PR #3074, opened 2024-12-21
Remove extra height class from NavBar

mrh-is:
## Linked Issue

None

## Description

The NavBar currently has both `h-full` & `h-20` applied by default. This means that its height is determined by which class ends up earlier in the CSS. Seems like this was probably an oversight to have both?

## Checklist

## Changsets

### DIFF
--- a/.changeset/sweet-pens-compare.md
+++ b/.changeset/sweet-pens-compare.md
@@ -3,4 +3,4 @@
 '@skeletonlabs/skeleton-react': patch
 ---
 
-Remove extra height class from NavBar
+chore: Remove extra height class from NavBar
--- a/.changeset/sweet-pens-compare.md
+++ b/.changeset/sweet-pens-compare.md
@@ -1,0 +1,6 @@
+---
+'@skeletonlabs/skeleton-svelte': patch
+'@skeletonlabs/skeleton-react': patch
+---
+
+Remove extra height class from NavBar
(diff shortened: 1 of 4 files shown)


endigo9740:
Good catch @mrh-is two things though:

1. We have to ensure we keep parity between both the Svelte and React components, so we need the same change to the React version here:
https://github.com/skeletonlabs/skeleton/blob/next/packages/skeleton-react/src/lib/components/Navigation/Navigation.tsx#L21
2. When reviewing the PR I noted our quick links to the source code for these components if off. We went through a process where we renamed from `<Nav>` -> `<Navigation>` but missed these links. I'll detail them below. It's a quick fix if you don't mind helping us out.
3. Per our PR templates, anytime you change the source in `/packages` you need to submit a Changeset. I've appended the original post above with the latest instruction.

Here's the location of the file, which GitHub tries to parse as markdown:
https://github.com/skeletonlabs/skeleton/blob/next/sites/next.skeleton.dev/src/content/docs/components/navigation/meta.mdx

Here's the raw values so you can see what I'm referring to:
https://raw.githubusercontent.com/skeletonlabs/skeleton/refs/heads/next/sites/next.skeleton.dev/src/content/docs/components/navigation/meta.mdx

We need to change:

```ts
srcSvelte: '/src/lib/components/Nav' <--- rename Navigation
srcReact: '/src/lib/components/Nav' <--- rename Navigation
```

mrh-is:
@endigo9740 Got it! Think I've handled all those. :)

endigo9740:
Thank you for that @mrh-is it's a huge help. Note I push one more quick change to append the Changeset description with `chore:`. This describes the type of change. These descriptions show verbatim in our generated changelog so we need that in place to match.

Thanks for your help though!

PR #3097, opened 2025-01-09
Make Nav.Tile id mandatory and update examples

### LINKED ISSUE #3085: NavTileProps.id is typed as optional, but is required
### Current Behavior

Using a `Navigation.Tile` component without an `id` is allowed by the type settings

### Expected Behavior

Using a `Navigation.Tile` component enforces the `id` property to be set

### Steps To Reproduce

```svelte
<Navigation.Rail expanded background="bg-transparent">
    {#snippet tiles()}
          <Navigation.Tile
              labelExpanded="Hello there!"
          >
              <SomeIcon />
          </Navigation.Tile>
    {/snippet}
</Navigation.Rail>
```

_No response_

### More Information

https://github.com/skeletonlabs/skeleton/blob/next/packages/skeleton-svelte/src/lib/components/Navigation/types.ts#L102

phamduylong:
## Linked Issue

Closes #3085 

## Description

Make `id` mandatory on Nav.Tile instances

## Changsets

## Checklist

### DIFF
--- a/.changeset/tall-pigs-perform.md
+++ b/.changeset/tall-pigs-perform.md
@@ -1,0 +1,6 @@
+---
+'@skeletonlabs/skeleton-svelte': patch
+'@skeletonlabs/skeleton-react': patch
+---
+
+bugfix: Make use of `useId` on Navigation component to set default `id` for Tiles if user had not defined it explicitly
--- a/packages/skeleton-react/src/lib/components/Navigation/Navigation.tsx
+++ b/packages/skeleton-react/src/lib/components/Navigation/Navigation.tsx
@@ -1,6 +1,6 @@
 'use client';
 
-import React, { createContext, useContext } from 'react';
+import React, { createContext, useContext, useId } from 'react';
 import { NavContextState, NavRailProps, NavBarProps, NavTileProps } from './types.js';
 
 // Contexts ---
@@ -186,6 +186,11 @@ export const NavTile: React.FC<NavTileProps> = ({
 }) => {
 	// Get Context
 	const ctx = useContext<NavContextState>(NavContext);
+
+	const generatedId = useId();
+	if (!id) {
+		id = generatedId;
+	}
 
 	// Local
 	const TileElement = href ? 'a' : 'button';
--- a/packages/skeleton-svelte/src/lib/components/Navigation/NavTile.svelte
+++ b/packages/skeleton-svelte/src/lib/components/Navigation/NavTile.svelte
@@ -1,9 +1,10 @@
 <script lang="ts">
+	import { useId } from '$lib/internal/use-id.js';
 	import { getNavigationContext } from './context.js';
 	import type { NavTileProps } from './types.js';
 
 	let {
-		id,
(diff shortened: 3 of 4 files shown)


sites/next.skeleton.dev/src/examples/components/navigation/ExampleExpanded.svelte
```diff
@@ -16,27 +16,27 @@
 	<!-- Component -->
 	<Navigation.Rail expanded>
 		{#snippet header()}
-			<Navigation.Tile href="/" labelExpanded="Menu"><IconMenu /></Navigation.Tile>
+			<Navigation.Tile id="0" href="/" labelExpanded="Menu"><IconMenu /></Navigation.Tile>
 		{/snippet}
 		{#snippet tiles()}
-			<Navigation.Tile id="0" labelExpanded="Browse Files" href={hrefExample}>
+			<Navigation.Tile id="1" labelExpanded="Browse Files" href={hrefExample}>
```
endigo9740, line 22: @phamduylong this is one of the only unique components we introduced that doesn't use Zag (intentionally). Now that I look at this, I wonder if rather than making ID mandatory, we utilize `useId()` to set a default automatically.

We didn't have access to `useId()` in Svelte when this component was built - that was introduced only when we integrated Zag. React has it built in as part of the library.

I think that'll provide a better experience - the ID will always be set, but users can still overwrite with a custom ID as they wish. Which would mean `id` would be optional again. Once less setting users have to deal with.

What do you think?

endigo9740 (changes requested):

phamduylong:
Not sure why the test fails but seems irrelevant for this PR.

endigo9740:
@phamduylong the CI tests are for specific apps, to `skeleton-docs` is the v2 doc site. Since there's no changes to that, and that we're working on v3 project, it can be ignored. The `skeleton-next` project is the one to keep an eye on here.

As per the PR, this looks great. This should be ready to merge, but I'm going to hold it while we continue to release CLI updates. Once those settle down I'll merge and this can be part of the next proper release.

phamduylong:
Updated changeset.

packages/skeleton-react/src/lib/components/Navigation/Navigation.test.tsx
```diff
@@ -128,7 +128,7 @@ describe('Navigation', () => {
 		it('Renders the component', () => {
 			const { getByTestId } = render(
 				<Navigation.Rail>
-					<Navigation.Tile></Navigation.Tile>
+					<Navigation.Tile id="0"></Navigation.Tile>
```
endigo9740, line 131: @phamduylong the change to use `useId()` is great Long, but we're missing the benefit to this change if we still show all our examples including an `id`. This should now be an optional field, so let's review these from the examples. The API reference will indicate this is available if folks wish to add it.

To put it another way, we try to keep examples a minimal as possible so folks can focus more on what is there, than what isn't.

endigo9740 (changes requested):

PR #3156, opened 2025-01-24
Docs: Updated News Feed

### LINKED ISSUE #3070: v3 Docs: News section is not updating
https://next.skeleton.dev/news

This section does not appear to be fetching the latest posts as they occur. Each post seems to be fetched during Astro's build-time, which means it only updates each release. We should review our options to resolve this. If need be, we might consider porting from a Astro component to Svelte or React.

Hugos68:
## Linked Issue

Closes #3070

## Description

Uses the BlueSky API to fetch the feed (which is now typesafe 🎉) and display it in a Svelte component as guided by: https://docs.astro.build/en/guides/data-fetching/#fetch-in-framework-components.

@endigo9740 Since you will see this, I've added placeholder blank cards with `animate-pulse`, feel free to edit this to be more "Skeleton", I'm no designer 😄.

This is what it looks like:
image

This is what it looks like while loading the posts:
image


Hugos68:
`pnpm check` fails due to errors not related to this PR, I will ignore for the sake of pressing release deadline.

PR #3170, opened 2025-01-25
Bugfix: Pin TW version in CSA

### LINKED ISSUE #3166: The Skeleton v2 CLI is installing Tailwind v4, which is not currently compatible
### Current Behavior

A user on Discord has confirmed that the current v2 CLI instructions are installing Skeleton v2 + Tailwind v4. This is not compatible and will lead to errors for any users trying to use this.

### Expected Behavior

The CSA app should be pinned to the latest v3.x release

The version should be updated from `latest` to `v3.4.17`

> NOTE: I'm assuming this will match package.json formatting for version
> For example `tailwindcss@3.4.17`, so we would only add `3.4.17`

https://github.com/skeletonlabs/skeleton/blob/dev/packages/create-skeleton-app/src/creator.js#L56

### Steps To Reproduce

https://www.skeleton.dev/docs/get-started

_No response_

### More Information

Additionally I recommend we pin other dependencies, especially those related to Skeleton. Otherwise similar issues with arise when Skeleton v3 drops.

Hugos68:
## Linked Issue

Closes #3166

## Description

Pins the TW version to the latest 3.

I wasn't able to generate a changeset because it doesn't show, not sure why?
image

### DIFF
--- a/packages/create-skeleton-app/src/creator.js
+++ b/packages/create-skeleton-app/src/creator.js
@@ -53,7 +53,7 @@ export class SkeletonOptions {
 		this.devDependencies = new Map([
 			['postcss', 'latest'],
 			['autoprefixer', 'latest'],
-			['tailwindcss', 'latest'],
+			['tailwindcss', '^3.4.17'],
 			['@skeletonlabs/skeleton', 'latest'],
 			['@skeletonlabs/tw-plugin', 'latest'],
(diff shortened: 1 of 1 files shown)


TazorDE (changes requested):
change looks good, but I think this should be merged into `dev`. That should also resolve the Changeset issue I think

PR #3223, opened 2025-02-16
Conditional rendering for trigger button

### LINKED ISSUE #3206: [maybe bug] Modal without trigger still creates trigger top left in the layout
### Current Behavior

I am migrating to V3 and some of my actions, which call modal/dialog, are in popover. I initially put Modal with trigger and content. But when I select an action and close the popover, I get a warning "something with window/frame is null" and my modal/dialog does not appear. So I implemented my own modal stack/overlay in a similar way as it was with the V2, and my modals don't have "trigger" defined.
+layout.svelte (root)
```js
<!-- Overlays -->
<ModalStack />
````
I also need this to show an error if it happens on form action.
When such 'trigger'-less modals appear, there also appears the trigger element in the html, which is empty but it does has size and shifts the whole content (under the dim).

If I show more modals (e.g. form, and then the error, more such elements appear).

```html
<span data-testid="modal" class=" "><button data-scope="dialog" data-part="trigger" id="dialog:hy0vnfbgdu50:trigger" aria-haspopup="dialog" type="button" aria-expanded="true" data-state="open" aria-controls="dialog:hy0vnfbgdu50:content" class="modal-trigger  "><!----></button>  <!----></span>

<span data-testid="modal" class=" "><button data-scope="dialog" data-part="trigger" id="dialog:d8kcn0zvgri3:trigger" aria-haspopup="dialog" type="button" aria-expanded="true" data-state="open" aria-controls="dialog:d8kcn0zvgri3:content" class="modal-trigger  "><!----></button>  <!----></span>
```

Image

Image

So my request is if 'trigger' is not provided, then this html elements should not appear.

Thank you

### Expected Behavior

Trigger html elements should not appear for the 'trigger'-less modals

### Steps To Reproduce

```js
<Modal open={true}>
  {#snippet content()}
    whatever
  {/snippet}
</Modal>
````

_No response_

### More Information

_No response_

phamduylong:
## Linked Issue

Closes #3206 

## Description

Only render button wrapping around `trigger` if the snippet was provided

## Changsets

## Checklist

### DIFF
--- a/.changeset/stale-coats-build.md
+++ b/.changeset/stale-coats-build.md
@@ -1,0 +1,5 @@
+---
+'@skeletonlabs/skeleton-svelte': patch
+---
+
+bugfix: only render button wrapping around `trigger` if the snippet was provided
--- a/packages/skeleton-svelte/src/lib/components/Modal/Modal.svelte
+++ b/packages/skeleton-svelte/src/lib/components/Modal/Modal.svelte
@@ -70,9 +70,11 @@
 
 <span class="{base} {classes}" data-testid="modal">
 	<!-- Trigger -->
-	<button {...triggerProps} class="{triggerBase} {triggerBackground} {triggerClasses}" {disabled} type="button">
-		{@render trigger?.()}
-	</button>
+	{#if trigger}
+		<button {...triggerProps} class="{triggerBase} {triggerBackground} {triggerClasses}" {disabled} type="button">
+			{@render trigger?.()}
+		</button>
+	{/if}
 
 	{#if api.open}
 		<!-- Backdrop -->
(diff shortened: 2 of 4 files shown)


Hugos68:
I know this has been closed, but just an observation, now that we wrap the `snippet` in a `#if` block, we can change: `{@render trigger?.()}` to `{@render trigger()}` because `trigger` will always be defined.

phamduylong:
> I know this has been closed, but just an observation, now that we wrap the `snippet` in a `#if` block, we can change: `{@render trigger?.()}` to `{@render trigger()}` because `trigger` will always be defined.

I had that in mind, but then missed it during the refactoring. Thanks for the catch mate.

PR #3233, opened 2025-02-18
Chore: Reenable prettier on `mdx` files.

Hugos68:
This PR re enables prettier for .mdx files, this is an experimental PR, we need to confirm the issues back then are resolved (mainly codeblocks in .mdx breaking prettier, I couldn't see any issues, so if you find any please report!).


Hugos68:
> @Hugos68 can we have prettier ignore the Changeset markdown files? Those are generated by Changeset's CLI, so I'd assume it's already formatted in the manner they prefer. It makes me a bit nervous to have a lintering touch those at all.

Why would we not want those formatted? It's replacing the quotes with our formatting standard, it doesn't affect changesets in any way.

endigo9740:
@Hugos68 We've already been through a scenario where Prettier was providing malformed adjustments to documentation. Messed up documentation is annoying but we can fix it. However, there may be nothing we can do if it breaks a changeset files and those records don't get included in the final changelog. I just don't think that's something we should risk here.

The Changeset CLI does a good job of keeping changesets uniform and to the specification they expect. There's no substantial benefit to it modifying quotes here. Changesets are an administrative tool, no user-facing. And almost never maintained by contributors where formatting is useful.

sites/next.skeleton.dev/src/content/docs/resources/contribute/documentation.mdx
```diff
@@ -138,9 +138,9 @@ export const tableProps: [
 ```
 
 ```html
-<Table label="Props" data="{tableData}" tonal>
+<table label="Props" data="{tableData}" tonal>
```
endigo9740, line 141: This is incorrect - it should remain as `<Table>` (capital T) as this is a component, not an HTML element.

endigo9740 (changes requested):

PR #3297, opened 2025-03-01
chore: replace eslint and prettier with biome

Hugos68:
Replaces ESLint and Prettier in favor of Biome.

Linting before took around 10-15 seconds, now it takes 2 seconds
Formatting before took around 10-15 seconds now it takes 2 seconds

Note: Biome doesn't support HTML/Markdown yet, so this will not be merged until then.


endigo9740:
@Hugos68 do we have any window or when MDX support might arrive? If not I'd say we table this PR until then. Otherwise it's just going sit around getting stale.

Hugos68:
Yes let's create an issue instead and close this out, I was just experimenting but it's indeed unnecessary to keep this stale until then. I can do this too if you want.

PR #3313, opened 2025-03-04
chore: update deps

### LINKED ISSUE #3310: Update all monorepo dependencies before launch
Prior to launch, we need to do another round of dependency updates for all projects.

Hugos68:
Closes #3310

Update all monorepo dependencies


.changeset/chilled-comics-hear.md
```diff
@@ -1,6 +1,6 @@
 ---
-"@skeletonlabs/skeleton-svelte": minor
-"@skeletonlabs/skeleton-react": minor
+'@skeletonlabs/skeleton-svelte': minor
```
endigo9740, line 2: All changeset quotes are being affected by Prettier again.

packages/skeleton-react/src/components/Avatar/Avatar.tsx
```diff
@@ -34,7 +34,7 @@ export const Avatar: FC<AvatarProps> = ({
 	// Zag
 	const service = useMachine(avatar.machine, {
 		id: useId(),
-		...zagProps
+		...zagProps,
```
endigo9740, line 37: Not a fan of commas for the last value in an array.

endigo9740 (commented):
Unfortunately I won't be able to provide a better review for now as the linting changes are vastly outweighing any dependency updates. I'm not sure I agree with all changes being made here. I might suggest we roll back or undo as many Linting-specific changes for now. Then proactively undo/re-enable the changes we want to keep in a dedicated linting PR following this? That will eliminate a lot o the noise here.

packages/skeleton-react/src/components/Avatar/Avatar.tsx
Hugos68: This is used for diffing better in git, but I think I'll close this PR, open a new one without improvements or anything, just only updating deps.

PR #3324, opened 2025-03-05
Add generic code slot example

phamduylong:
image


endigo9740:
@phamduylong great meme lol.

Per the change, rather than adding a third snippet, can you extend the second to append the generic fragment. That way it shows:

- example
- generic
- react
- svelte

All in one.

PR #3348, opened 2025-03-09
Combobox option template

### LINKED ISSUE #3341: Add snippet and generic types to Combobox component

I’d like to propose two related enhancements to the Combobox component. In the previous v2 implementation of the Autocomplete component, HTML was allowed to be passed as the option label, giving users the option to render custom content.

My proposal is to reintroduce similar behavior through the use of an optional snippet (in the case of the Svelte implementation):

```js
<button {...api.getItemProps({ item })} class="{optionBase} {displayClass} {optionClasses}" type="button">
  {#if itemSnippet}
    {@render itemSnippet(item)}
  {:else}
    {item.label}
  {/if}
</button>
```

Additionally, I think it would make sense to either add an optional meta property to the data object to allow users to pass any extra contextual data that might be useful when rendering the option, or as @Hugos68 suggested, make the Combobox data a generic type and let the user decide what the data should be (keeping the internal `value` mandatory).

One example use case of this is a simple timezone picker with custom formatting, where the timezone's offset is displayed next to the value:

Image

New Feature

_No response_

phamduylong:
## Linked Issue

Closes #3341 

## Description

Enable user to have custom template for combobox options

## Changsets

## Checklist

### DIFF
--- a/.changeset/poor-dolls-arrive.md
+++ b/.changeset/poor-dolls-arrive.md
@@ -1,0 +1,6 @@
+---
+"@skeletonlabs/skeleton-svelte": patch
+---
+
+feat: provide `itemTemplate` snippet for Combobox
+  
--- a/.changeset/poor-dolls-arrive.md
+++ b/.changeset/poor-dolls-arrive.md
@@ -2,5 +2,5 @@
 "@skeletonlabs/skeleton-svelte": patch
 ---
 
-feat: provide `itemTemplate` snippet for Combobox
+feat: provide `item` snippet for Combobox
   
--- a/packages/skeleton-svelte/src/components/Combobox/types.ts
+++ b/packages/skeleton-svelte/src/components/Combobox/types.ts
@@ -1,3 +1,4 @@
+/* eslint-disable prettier/prettier */
 import type { Snippet } from 'svelte';
 import * as combobox from '@zag-js/combobox';
 
@@ -69,7 +70,7 @@ export interface ComboboxProps extends Omit<combobox.Props, 'id' | 'collection'>
 	/** Provide a custom arrow icon. */
 	arrow?: Snippet;
 	/** Provide a custom template for the option. */
-	itemTemplate?: Snippet<[{ label: string; value: string }]>;
+	item?: Snippet<[{ label: string; value: string }]>;
 
 	// Events ---
 	/** Handle the combobox dropdown button click event. */
--- a/packages/skeleton-svelte/src/components/Combobox/types.ts
+++ b/packages/skeleton-svelte/src/components/Combobox/types.ts
@@ -68,6 +68,8 @@ export interface ComboboxProps extends Omit<combobox.Props, 'id' | 'collection'>
 	// Snippets ---
 	/** Provide a custom arrow icon. */
 	arrow?: Snippet;
+	/** Provide a custom template for the option. */
+	itemTemplate?: Snippet<[{ label: string; value: string }]>;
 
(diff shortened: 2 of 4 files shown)


phamduylong:
Thought about removing `label` from item, can't bring myself to do it. Seems unworthy.

sites/next.skeleton.dev/src/examples/components/popover/ExampleCombobox.svelte
```diff
@@ -21,4 +22,12 @@
 	onValueChange={(e) => (selectedCountry = e.value)}
 	label="Select Country"
 	placeholder="Select..."
-/>
+>
+	<!-- This is optional. Combobox will render label by default -->
+	{#snippet itemTemplate(item)}
+		<div class="flex w-full justify-between space-x-2">
+			<span>{item.label}</span>
+			<span>{(item as ComboboxData).emoji}</span>
```
Hugos68, line 30: We should probably make the extension of `ComboboxItem` generic so you don't have to do these type of assertions. This would mean utilizing Svelte's generics: https://svelte.dev/docs/svelte/typescript#Generic-$props (React would be similar).

Hugos68 (changes requested):
Looks good, couple things here and there and also missing the React implementation.

packages/skeleton-svelte/src/components/Combobox/types.ts
```diff
@@ -68,6 +68,8 @@ export interface ComboboxProps extends Omit<combobox.Props, 'id' | 'collection'>
 	// Snippets ---
 	/** Provide a custom arrow icon. */
 	arrow?: Snippet;
+	/** Provide a custom template for the option. */
+	itemTemplate?: Snippet<[{ label: string; value: string }]>;
```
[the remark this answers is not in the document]
phamduylong: `itemTemplate` express that it's a template. It avoids polluting the common naming space in the file. I'd say we go and make `arrow` -> `arrowTemplate` rather than the other way around. `item` and `arrow` can be very confusing as props names.

sites/next.skeleton.dev/src/examples/components/popover/ExampleCombobox.svelte
phamduylong: Looks nice, I didn't know the generic props existed. Will check it out.

packages/skeleton-svelte/src/components/Combobox/types.ts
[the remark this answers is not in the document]
Hugos68: That's fine too, I just think it should be consistent between the too, I agree that `arrow` could mean anything, but then again, that is kind of typescripts job. I don't name my variables `countNumber`, I name them `count` with type `number` for example. I didn't see this `Template` convention anywhere else, so was just wondering.

endigo9740:
@phamduylong we'll await @Hugos68 final review, but otherwise this looks good to me.

If all is well, feel free to merge right away tomorrow Hugo. Thanks!

.changeset/poor-dolls-arrive.md
```diff
@@ -0,0 +1,6 @@
+---
+"@skeletonlabs/skeleton-svelte": patch
+---
+
+feat: provide `itemTemplate` snippet for Combobox
```
Hugos68, line 5: Please edit the changeset to `item` instead of `itemTemplate`.

Hugos68 (changes requested):

Issue #2991, opened 2024-11-20
Progress ring's indeterminate animation doesn't work well at smaller sizes


knd775:
### Current Behavior

The ring disappears for some time each loop. This example is with `size-4`

https://github.com/user-attachments/assets/af177305-d7de-4fef-b69e-5e826b416cbc

endigo9740:
@knd775 Ben just FYI I did take a look at this yesterday. I agree the animation doesn't work well at the small size, so I've gotta figure out how to address this. The issue is our default animation defined here:

https://github.com/skeletonlabs/skeleton/blob/next/packages/skeleton/src/plugin/extends/animations.ts#L15

The way these components work, is by using an SVG circle and animating the stroke like so:

https://css-tricks.com/svg-line-animation-works/

So it's not a fill like a progress bar, but actually a single portion of a dashed line stretched for the entire stroke's length. As such, it's not dynamic. Meaning as the size of the circle grows and shrinks it should be a relatively larger/smaller value. Where as right now we have a fixed value.

This does scale somewhat - just not to the extremes, so what I might do is create a few variations of this and use use responsive breakpoints to swap between. That is unless I can come up with a better way to accomplish this and still allow users to replace the animation if desired.

nullpointerexceptionkek:
Adding on to this,
The animation also breaks if the size is too large

endigo9740:
@knd775 @nullpointerexceptionkek FYI we think this PR will improve things.

https://github.com/skeletonlabs/skeleton/pull/3360

This will be a part of this weekend v3 launch release. We'll want to check before we modify the CSS animation itself. But if need be we can and will follow up on this.

Please feel free to report back after migrating to the final release Otherwise we'll circle back asap.

nullpointerexceptionkek:
@Hugos68 @endigo9740 

I am unable to reproduce this issue, perhaps its an issue with zag and its fixed?


In addition, should it be added to the documentation that strokeWidth is relative and not absolute because of the PR #3360

knd775:
> feel free to update and confirm.

We just finished our migration to the final release and the progress ring works much better at all sizes we're currently using. Thanks!

Hugos68:
@knd775 That's great to hear!

PR #3350, opened 2025-03-10
fix(progress-ring): Fix progress ring being 100% on Safari

### LINKED ISSUE #3222: V3 ProgressRing always displays 100% on iOS for values greater than 0
### Current Behavior

On iOS (Safari), the ProgressRing component does not display progress correctly:

- If value = 0, it displays correctly as 0% progress.

- If value > 0 (even 1), it immediately jumps to 100% progress, ignoring the actual value.

- The issue occurs only in iOS browsers, the component works correctly on desktop browsers.

### Expected Behavior

- ProgressRing should display the correct progress value on iOS just like it does on desktop.

- The ring should scale between 0% and 100%, rather than jumping to fully completed for any nonzero value.

- It should maintain smooth transitions on iOS devices.

### Steps To Reproduce

1. Use ProgressRing on an iOS device with the following setup:
```svelte
<script>
let value= $state(0)
</script>

<ProgressRing value={value} size="size-14" />
<button onclick={() => value = 0}>set 0</button>
<button onclick={() => value = 25}>set 25</button>
<button onclick={() => value = 75}>set 75</button>
```
2. Open the page on Safari on iOS.
3. Click the buttons to change value.
4. Notice that any value greater than 0 causes the ring to display 100% progress, instead of the expected percentage.

_No response_

### Environment Information

_No response_

### More Information

_No response_

Hugos68:
## Linked Issue

Closes #3222
Partially improves #2991

## Description

Fixes the progress by adding @nullpointerexceptionkek's suggestion of applying the `viewBox` prop.

### DIFF
--- a/.changeset/stupid-rockets-see.md
+++ b/.changeset/stupid-rockets-see.md
@@ -1,0 +1,7 @@
+---
+"@skeletonlabs/skeleton-svelte": patch
+"@skeletonlabs/skeleton-react": patch
+---
+
+bugfix: Fix safari showing 100% on progress rings
+  
--- a/packages/skeleton-react/src/components/ProgressRing/ProgressRing.tsx
+++ b/packages/skeleton-react/src/components/ProgressRing/ProgressRing.tsx
@@ -63,6 +63,7 @@ export const ProgressRing: FC<ProgressRingProps> = ({
 			{/* SVG */}
 			<svg
 				{...api.getCircleProps()}
+				viewBox="0 0 100 100"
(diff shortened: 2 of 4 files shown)


endigo9740:
Unfortunately the issue persists in the desktop Safari, even though it appears resolved in iOS.

image

Given the time crunch to get v3 shipped I'm going to close this PR, and we'll reopen and revisit the issue post-release.

Thanks for testing Hugo!

PR #3411, opened 2025-03-14
Substitute raw RGB with OKLCH

### LINKED ISSUE #3402: malformed theme file contrast values when using themes.skeleton.dev
### Current Behavior

When setting `white` or `black` as the color contrast values they are added as malformed color values to the theme file.

```css
--color-primary-contrast-dark: 0 0 0;
--color-primary-contrast-light: 255 255 255;
```

Image

### Expected Behavior

They should be added as valid CSS colors

```css
--color-primary-contrast-dark: rgb(0 0 0);
--color-primary-contrast-light: rgb(255 255 255);
```
or

```css
--color-primary-contrast-dark: black;
--color-primary-contrast-light: white;
```

### Steps To Reproduce

1. Using the theme generator
2. Set the contrast colors to either `white` or `black`
3. They are added to the theme-file in a malformed manner

themes.skeleton.dev

### Environment Information

_No response_

### More Information

_No response_

phamduylong:
## Linked Issue

Closes #3402 

## Description

Substitute raw RGB with OKLCH

## Checklist

## Changsets


endigo9740:
This is looking good @phamduylong. Can you do me a favor and scan through the preset themes - I think we had some of the old color values sneak through when converting to the CSS format. For example:

https://github.com/skeletonlabs/skeleton/blob/main/packages/skeleton/src/themes/catppuccin.css#L37

I am confident the main colors are converted, but contrast values and background colors were likely not converted properly due to the bugs you're addressing here.

endigo9740 (changes requested):
This is looking good @phamduylong. Can you do me a favor and scan through the preset themes - I think we had some of the old color values sneak through when converting to the CSS format. For example:

https://github.com/skeletonlabs/skeleton/blob/main/packages/skeleton/src/themes/catppuccin.css#L37

I am confident the main colors are converted, but contrast values and background colors were likely not converted properly due to the bugs you're addressing here. I [shared some tips](https://github.com/skeletonlabs/skeleton/issues/3402#issuecomment-2724983387) in my comment in the issue to address this.

phamduylong:
> This is looking good @phamduylong. Can you do me a favor and scan through the preset themes - I think we had some of the old color values sneak through when converting to the CSS format. For example:
> 
> https://github.com/skeletonlabs/skeleton/blob/main/packages/skeleton/src/themes/catppuccin.css#L37
> 
> I am confident the main colors are converted, but contrast values and background colors were likely not converted properly due to the bugs you're addressing here. I [shared some tips](https://github.com/skeletonlabs/skeleton/issues/3402#issuecomment-2724983387) in my comment in the issue to address this.

So this should be done even for valid values in themes CSS files, you're saying?

endigo9740:
@phamduylong if the format is still `XX XX XX` that is *not* valid, that's the problem. In v2 we would use those like this:

```
rgb(var(XX XX XX))
```

But that's no longer the case in v3, we assume the color if fully formed before we use it:

```
var(XX XX XX)
```

So very much not valid. There's no `rgb()` function and there's no commas.

So yes, if we add those missing details they will be valid and work, but we're trying to standardize on OKLCH, so we need that extra conversion step as well.

endigo9740 (approved):

PR #3418, opened 2025-03-16
Bugfix: ensure toast unique

### LINKED ISSUE #3415: Svelte: Toast timeout appears to be global
### Current Behavior

When testing out the [Toast in the v3 documentation](https://www.skeleton.dev/docs/integrations/toasts/svelte) for Svelte, I noticed that when creating multiple toasts, they all will disappear at the same time, even though they should be disappearing one at a time. This behavior is working correctly for the React component.

### Expected Behavior

When multiple toasts are on screen, each one is expected to have its own timeout before disappearing and toasts that were created 3 seconds apart should disappear 3 seconds apart, rather than both at the same time.

### Steps To Reproduce

1. Go to the [Toast docs for Svelte](https://www.skeleton.dev/docs/integrations/toasts/svelte)
2. Click on the toast preview for "Error"
3. Wait a second or two
4. Click on the toast preview for "Error" again
5. Observe two toasts on screen
6. In a few seconds, observe both toasts disappear at the same time
    a. Expected: Toast 1 disappears, then a few seconds later Toast 2 disappears

_No response_

### Environment Information

_No response_

### More Information

_No response_

nullpointerexceptionkek:
## Linked Issue

Closes #3415

## Description

- Hotfixed the original Svelte toast function to ensure all toasts dismiss simultaneously.

- Migrated the integration to svelte-french-toast as the new default and deprecated the original Svelte toast

## Checklist

## Changsets

### DIFF
--- a/packages/skeleton-svelte/src/components/Toast/ToastProvider.svelte
+++ b/packages/skeleton-svelte/src/components/Toast/ToastProvider.svelte
@@ -58,11 +58,14 @@
 	const id = $props.id();
 	setContext('toast', {
 		create: (toast: Toast) => {
+			// Unique ID
+			const unid = id + '-' + crypto.randomUUID();
+			console.log(unid)
 			// Set default settings
 			toast = {
 				...defaults[String(toast.type || 'info')],
 				...toast,
-				id: id
+				id: unid
 			};
 			// Push to Queue
 			toastQueue.push(toast);
--- a/packages/skeleton-svelte/src/components/Toast/ToastProvider.svelte
+++ b/packages/skeleton-svelte/src/components/Toast/ToastProvider.svelte
@@ -54,24 +54,28 @@
 	// State
 	let toastQueue: Toast[] = $state([]);
 
+	// Unique toast count
+	let count = 0;
+
+	function getId(prefix: string) {
+		return prefix + String(count++);
+	}
 	// Context
-	const id = $props.id();
+	const prefix = $props.id();
 	setContext('toast', {
 		create: (toast: Toast) => {
-			// Unique ID
-			const unid = id + '-' + crypto.randomUUID();
 			// Set default settings
-			toast = {
+			const defaultToast = {
 				...defaults[String(toast.type || 'info')],
 				...toast,
-				id: unid
+				id: getId(prefix)
 			};
 			// Push to Queue
-			toastQueue.push(toast);
+			toastQueue.push(defaultToast);
 			// Set duration timeout
 			setTimeout(() => {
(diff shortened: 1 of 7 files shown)


packages/skeleton-svelte/src/components/Toast/ToastProvider.svelte
```diff
@@ -58,11 +58,13 @@
 	const id = $props.id();
 	setContext('toast', {
 		create: (toast: Toast) => {
+			// Unique ID
+			const unid = id + '-' + crypto.randomUUID();
```
Hugos68, line 62: Let's not tightly depend on the crypto API, we don't need that, please move to an incrementer:
```html
<script lang="ts" module>
    let count = 0;
    
    function getId(prefix: string) {
        return prefix + String(count++);
    }
</script>
```

packages/skeleton-svelte/src/components/Toast/ToastProvider.svelte
```diff
@@ -58,11 +58,13 @@
...
 			// Set default settings
 			toast = {
```
Hugos68, line 64: Don't mutate existing objects, it's confusing, I know you didn't write this code but just create a new object instead.

Hugos68 (changes requested):

Hugos68 (approved):

endigo9740:
@nullpointerexceptionkek @Hugos68

See my message here to ensure we're all on the same page:

- https://github.com/skeletonlabs/skeleton/issues/3415#issuecomment-2730160009

We essentially have two paths forward here:

1. We finalize this PR and merge it, so the fix and new Svelte integration guide are immediately available for users.
2. We roll back and remove the Svelte integration and replace it with the original "homegrown" component docs, then push merge and release as an immediate fix for folks that need it.

In either scenario we will replace this integration with the Zag-based Toast component if/when ready. Though I don't have an ETA for that - that'll need to come from Hugo.

Any preference here? I'd say if the Toast component won't be ready for a bit, #1 will be the better short term solution.

nullpointerexceptionkek:
I think either way, the hot fix for the global timeout is needed. It sort of depends on how long the new toast component will take. I think the second solution might be better. I also have time to work on this if you need me.

endigo9740:
Yeah that's my goal - get the fix out asap.

> In addition, the last time I checked, Zag still have the display issue if you spam the notification, like what I showed on discord. They don't seem to fixed this yet(looking at changelog).

@Hugos68 verbally confirmed all original issues were fixed, so that'll be something to check in his PR/proof-of-concept/etc.

PR #3434, opened 2025-03-19
fix: add missing spacing in table of contents

### LINKED ISSUE #3431: Table of contents broken spacing
### Current Behavior

This is how ToCs look right now (you can check on any skeleton doc page):

Image

This is how it should look like:

Image

Notice that in the current version there is missing spacing between the title and content of the ToC. This is caused I believe by the title spacing being done with a margin, but the margin is applied to the title, which is an inline element (`<span>`).  I believe this is caused by the tailwind [`space-y` breaking change](https://tailwindcss.com/docs/upgrade-guide#space-between-selector), it now applies `mb-2` to the top element, instead of `mt-2` to the bottom element (which is block).

rChaoz:
## Linked Issue

Closes #3431

## Description

Adds back missing spacing in table of contents components (both on site and cookbook) as it was removed by a Tailwind v4 breaking change (see issue).

If the title should keep being a `<span>`, then `space-y` should be removed and `mt-2` added to the content instead.

## Checklist

- [x] (N/A) Documentation should be updated to describe all relevant changes
- [x] (N/A) If you modify `/package` projects, please supply a Changeset

## Changsets

### DIFF
--- a/sites/skeleton.dev/src/components/docs/TableOfContents.astro
+++ b/sites/skeleton.dev/src/components/docs/TableOfContents.astro
@@ -23,7 +23,7 @@ function setDepthClass(depth: number) {
 {
 	headings.length > 0 && (
 		<nav class="text-sm space-y-2">
-			<span class="font-bold">On This Page</span>
+			<div class="font-bold">On This Page</div>
 			<ul class="space-y-2">
 				<li>
 					<a href={`#_top`} class="anchor block">
--- a/.changeset/cool-beers-call.md
+++ b/.changeset/cool-beers-call.md
@@ -1,0 +1,5 @@
+---
+"@skeletonlabs/skeleton.dev": patch
+---
+
+fix: add missing spacing in table of contents
--- a/.changeset/cool-beers-call.md
+++ b/.changeset/cool-beers-call.md
@@ -1,5 +1,0 @@
----
-"@skeletonlabs/skeleton.dev": patch
----
-
-fix: add missing spacing in table of contents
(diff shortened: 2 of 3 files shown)


sites/skeleton.dev/src/components/docs/TableOfContents.astro
```diff
@@ -23,7 +23,7 @@ function setDepthClass(depth: number) {
 {
 	headings.length > 0 && (
 		<nav class="text-sm space-y-2">
-			<span class="font-bold">On This Page</span>
+			<div class="font-bold">On This Page</div>
```
Hugos68, line 26: Instead of doing this, why don't we remove `space-y-2` and replace it with `grid gap-2`. Much better layout tool.

Hugos68 (changes requested):

sites/skeleton.dev/src/components/docs/TableOfContents.astro
rChaoz: Is there a reason why the title needs to be a `<span>`? Also, personally I didn't want to switch to flex/grid as `space-y` is already in use in a bunch of placed and I like consistency, this PR merely fixes the breaking change from Tailwind.

Another solution is to add `mt-2` to the second element. Regardless, let me know what you want to go with.

endigo9740: > Instead of doing this, why don't we remove space-y-2 and replace it with grid gap-2. Much better layout tool.

This is only every used in a vertical layout so it shouldn't matter too much. Grid is great when things flip between vertical/horizontal.

> Is there a reason why the title needs to be a <span>?

If there's a good reason to change it I'm open to suggestions.

rChaoz: > If there's a good reason to change it I'm open to suggestions.

I mean, this PR changes it to a div to work with the new `space-y` tailwind changes. There's multiple ways to change it but I choose the simplest one/with the least changes.

endigo9740: Yeah that should be fine then.

endigo9740 (approved):

PR #3442, opened 2025-03-20
fix: add missing core utilities

### LINKED ISSUE #3417: Documented classes don't actually exist
### Current Behavior

For example, `base-font-color`, `base-font-size` etc., they are [documented](https://www.skeleton.dev/docs/get-started/core-api#base-1), and the migration script will use them (e.g. `text-token` becomes `base-font-color`, but they don't actually exist.

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

_No response_

### Environment Information

_No response_

### More Information

_No response_

rChaoz:
## Linked Issue

#3417 

## Description

Implements the missing Core API classes as documented here:

- Body Background: https://www.skeleton.dev/docs/get-started/core-api#colors
- Typography: https://www.skeleton.dev/docs/get-started/core-api#typography

## Checklist

- [x] N/A Documentation should be updated to describe all relevant changes

## Changsets


packages/skeleton/src/base/theme.scss
```diff
@@ -134,3 +134,65 @@
 	--text-9xl: calc(8rem * var(--text-scaling));
 	--text-9xl--line-height: calc(1 * var(--text-scaling));
 }
+
+/* Add missing theme utilities - solution until variables can be renamed to follow Tailwind theme namespaces in v4 */
+
+// Colors
+@utility body-color {
```
endigo9740, line 141: I'm a little confused at why `body-color` were not included within `$theme-variables` and the loop.

packages/skeleton/src/base/theme.scss
```diff
@@ -134,3 +134,65 @@
 	--text-9xl: calc(8rem * var(--text-scaling));
 	--text-9xl--line-height: calc(1 * var(--text-scaling));
 }
+
+/* Add missing theme utilities - solution until variables can be renamed to follow Tailwind theme namespaces in v4 */
```
endigo9740, line 138: If these are `@utility` classes, they should live in the `/utilities` section of the package. Not in `/base/theme.css`

packages/skeleton/src/base/theme.scss
```diff
@@ -134,3 +134,65 @@
...
+	(anchor-line-height, line-height),
+	(anchor-font-weight, font-weight),
+	(anchor-font-style, font-style),
+	(anchor-letter-spacing, letter-spacing),
+	(anchor-text-decoration, text-decoration),
+	(anchor-text-decoration-active, text-decoration),
+	(anchor-text-decoration-focus, text-decoration),
+	(anchor-text-decoration-hover, text-decoration)
+);
+@each $variable in $theme-variables {
```
endigo9740, line 180: Good use of Sass here.

packages/skeleton/src/base/theme.scss
```diff
@@ -134,3 +134,65 @@
...
+	$name: list.nth($variable, 1);
+	$property: list.nth($variable, 2);
+	@utility #{$name} {
+		#{$property}: var(--#{$name});
+	}
+}
+// Edges
```
endigo9740, line 187: Default edge properties like border/ring/divide are already available because they use the correct namespace. You're correct, these were undocumented, which is why I contacted the Tailwind team to gather these:

https://github.com/tailwindlabs/tailwindcss.com/issues/2073

So implementing these as utility classes should be redundant and unnecessary.

endigo9740 (changes requested):

packages/skeleton/src/base/theme.scss
rChaoz: Because the class name is not equal to the variable name, which is true for every other variable. Basically, in the loop I do

```css
@utility NAME {
    property: var(--NAME);
}
```

If I were to add it to the loop, I'd have to specify 3 things for each entry instead of 2, which would make the code even longer.

packages/skeleton/src/base/theme.scss
rChaoz: Ah, my bad... I had no clue. I really spent time time searching for `--default-*` variables and couldn't find anything, so I decided to add this to be safe.

packages/skeleton/src/utilities/core.scss
```diff
@@ -0,0 +1,52 @@
+@use 'sass:list';
+
+// Colors (can't use the same sass loop as class name != var name)
```
endigo9740, line 3: I'm tying to decide if it makes sense to implement this utility using `light-dark()`. So it would be:

```css
@utility body-color {
	background-color: light-dark(var(--body-background-color), var(--body-background-color-dark));
}
```

There should be very few use cases where users don't want this to act as a pairing IMO.

endigo9740: I suppose wouldn't get that functionality unless we generate a utility for them once the theme properties have been renamed though. So perhaps this would set a bad precedence for future support.

rChaoz: That applies to everything, though. Variables will be named `color-*` and be accessed with `bg-`, not with any of the existing methods, so I don't think I understand the exact issue here. 

From what I can tell, when the renaming will come, there will likely be a migration guide + script with the old/new names, and regardless of what names are chosen, 100% of them will have to be changed, as currently none of them have the tailwind prefix.

endigo9740: Thinking about it, let's table or this now and plan to revisit when we make the breaking theme property changes in the next major release. I'd really prefer something more like this to be explicit and balanced:

```css
@theme {
    --color-background-light: ...;
    --color-background-dark: ...;
}

@theme inline {
    --color-background: light-dark(var(--color-body-background-light), var(--color-body-background-dark));
}
```

packages/skeleton/src/utilities/core.scss
```diff
@@ -0,0 +1,52 @@
...
+@utility body-color {
```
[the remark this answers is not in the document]
rChaoz: Yeah, that's exactly what I was following.

@endigo9740 I just made the changes. I also updated the docs page with the correct class name and removed a random usage of the wrong/non-existing utility in the theme generator.

Apparently the background color in the theme generator is set by the actual `<body>` element, not the preview, but if that should be changed I think that's another PR, I just removed a bit of dead code.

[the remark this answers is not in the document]
endigo9740: Awesome, looks good. Note I've removed the `closes` from the PR description above, as this represents only the first wave of fixes. I'll move that ticket into our v4.0 (Next) milestone so we can revisit.

As for the theme generator - that's due for a bit of an overhaul, so we'll revisit that in the near future as well. We'll likely end up rewriting most of that app so any changes can be made along the way.

endigo9740 (approved):

endigo9740 (approved):

Issue #2383, opened 2024-01-03
Cookbook: Layouts (replaces App Shell)


endigo9740:
This will act as a hub to centralize this information.

## Maintainer Requests

The following requests are coming straight from the Skeleton team. These are highly likely be implemented:

> [!NOTE] 
> We're aiming to retire the AppShell in Skeleton v3. The number of issues introduced by it far outweigh it's potential gains. Instead, we'll focus instead on providing a set of [Cookbook](https://next.skeleton.dev/docs/resources/cookbook) recipes for common layouts, that will help teach the core concepts for building layouts with Tailwind.

* ~~https://github.com/skeletonlabs/skeleton/issues/1993~~

## Community Requests

The following requests have come from the community and are under consideration:
 
* https://github.com/skeletonlabs/skeleton/issues/2383#issuecomment-1913361001

## Bugs and Known Issues

* Mobile Safari bottom panel doesn't auto-hide if the `<body>` doesn't scroll ([reference](https://github.com/skeletonlabs/skeleton/issues/2560))
* Pull to refresh fails is inaccessible for mobile browsers
* Provides some a11y concerns around scrolling with assistive devices
* Can negatively affect CSS print styles
* Is not immediately clear on how to implement sticky headers and other common layouts
* Pigeonholes app design to a few common layouts
* The way layouts are implement can vary wildly between meta frameworks (ex: Astro vs SvelteKit)

## Feedback

If you have additional updates or requests for this feature, please do so in the comments section below.

xmlking:
`svelte:window bind:scrollY` not working due to 
```css
html,
body {
  @apply h-full overflow-hidden;
}
```

for example  I added `<GotoTop showAtPixel={1} />` at the end to `+layout.svete`  in freshly created app with `npm create svelte@latest my-skeleton-app` 
following example only works if comment above CSS from `app.postcss` 

**GoToTop** example 
https://svelte.dev/repl/1c49e964ab144dcf82bb607a9aff74b5?version=3.58.0

endigo9740:
@xmlking when you use the App Shell (which is purely optional), the shell scrolls, not the body. So we provide a way to bind and listen to the scroll positions. This is described here:

https://www.skeleton.dev/components/app-shell#tracking-scroll-position

PR #3444, opened 2025-03-20
feat: AppShell

### LINKED ISSUE #3428: Provide AppShell cookbook recipe

I understand exactly why AppShell was removed, but I think Skeleton could provide a _new_ AppShell, with similar features to the old one, but with none of the drawbacks.

For example, this is what I created as a replacement in my project as I migrated to Skeleton v3:

<details><summary>AppShell.svelte</summary>

```svelte
<script lang="ts">
    import type { Snippet } from "svelte"
    import type { ClassValue, SvelteHTMLElements } from "svelte/elements"

    interface Props {
        /**
         * Main page content.
         */
        children?: Snippet
        /**
         * Classes applied to the `<main>` element.
         */
        class?: ClassValue
        /**
         * Custom props applied to the `<main>` element.
         */
        props?: Omit<SvelteHTMLElements["main"], "class" | "children">

        /**
         * Header content.
         */
        header?: Snippet
        /**
         * Classes applied to the `<header>` element.
         */
        headerClass?: ClassValue
        /**
         * Custom props applied to the `<header>` element.
         */
        headerProps?: Omit<SvelteHTMLElements["header"], "class" | "children">
        /**
         * If non-null, the `scroll-padding-top` CSS property of the `<html>` element will be set to this value,
         * with `{headerHeight}` replaced with the height of the header, in pixels.
         * @default calc({headerHeight}px + 1rem)
         */
        headerScrollPadding?: string
        /**
         * If true, the header element will be sticky, remaining fixed as the page is scrolled.
         * @default true
         */
        stickyHeader?: boolean

        /**
         * Footer content.
         */
        footer?: Snippet
        /**
         * Classes applied to the `<footer>` element.
         */
        footerClass?: ClassValue
        /**
         * Custom props applied to the `<footer>` element.
         */
        footerProps?: Omit<SvelteHTMLElements["header"], "class" | "children">
    }

    const {
        // Main content
        children,
        class: classes,
        props,
        // Header
        header,
        headerClass,
        headerProps,
        headerScrollPadding = "calc({headerHeight}px + 1rem)",
        stickyHeader = true,
        // Footer
        footer,
        footerClass,
        footerProps,
        // TODO sidebars
    }: Props = $props()

    let headerHeight = $state(0)

    // Set scroll padding to prevent content from being hidden behind the header
    $effect(() => {
        if (stickyHeader && headerElement && headerScrollPadding != null)
            document.documentElement.style.scrollPaddingTop = headerScrollPadding.replace("{headerHeight}", headerHeight)
    })
</script>

<svelte:body onscroll={onBodyScroll} />

{#if header}
    <header
        bind:this={headerElement}
        bind:offsetHeight={headerHeight}
        data-hide={stickyHeader && hideHeaderOnScroll && shouldHideHeader ? "" : undefined}
        {...headerProps}
        class={[
            headerClass,
            stickyHeader && "sticky top-0 z-[99]",
            stickyHeader && hideHeaderOnScroll && "data-hide:-translate-y-full translate-y-0 transition-[translate]",
        ]}>
        {@render header()}
    </header>
{/if}

<main {...props} class={classes}>{@render children?.()}</main>

{#if footer}
    <footer {...footerProps} class={footerClass}>{@render footer()}</footer>
{/if}
```

</details>

It solves the basic header/main/footer layout, with the option to make the header sticky easily, taking care of stuff like scroll padding.

New Feature

I'm happy to make a base PR for this. I also would like to address a few possible issues with this, that I think aren't really issues:
1. It doesn't provide a main/scrolling div, instead relying on the `<body>` to scroll, which fixes accessability issues, framework integration issues (e.g. SvelteKit resets body scroll on navigation, but this didn't work with previous AppShell), pull to refresh/hiding the browser bar on mobile etc.
2. It won't have the same features, so it's not a 1:1 replacement. The documentation on layouts will still exist, and will be recommended for more complex use-cases.
3. It will cover basic use cases (e.g. header/main/footer, optionally sticky header, possibly sidebars), and it's purpose will be to be a great starting point for new websites, similar to the old AppShell. It doesn't have to be 100% customisable as there will be documentation on replacing it with a custom layouts (but users will have the option to copy/paste the AppShell code as a starting point), and the layouts page should remain.
4. I understand that Skeleton is all about customizability and encouraging users to create their own, but I believe providing a basic layout, with options to easily move to a custom layout, is fine.

rChaoz:
## Linked Issue

Closes #3428 

## Description

Provide AppShell component.

**Known issues**, to be fixed:
- sticky sidebars don't have their height limited and glitch out if content is too large, should have a `max-height` + `overflow: hidden` set

## Checklist


packages/skeleton-svelte/src/components/AppShell/AppShell.svelte
```diff
@@ -0,0 +1,84 @@
...
+<!-- @component Base application layout component. -->
+
+<div id="appShell" class="{base} {classes}">
+	{@render children()}
+</div>
+
+<style>
+	:global(body, html) {
+		min-height: 100%;
+	}
```
rChaoz, line 77: The old AppShell didn't have this included, instead the documentation simply mentioned to add this to the global CSS file. I'm fine with either approach here.

packages/skeleton-svelte/src/components/AppShell/AppShell.svelte
```diff
@@ -0,0 +1,84 @@
...
+
+	#appShell {
+		display: grid;
+		grid-template-areas: 'header header header' 'sidebar-left page sidebar-right' 'footer footer footer';
+		grid-template-columns: auto 1fr auto;
+	}
```
rChaoz, line 83: When porting to React, I haven't seen any components using raw CSS like this. I believe `grid-template` cannot be achieved with pure tailwind. So there are two options:
- use inline styles, but this is harder to override, although at that point users should opt for a custom layout instead
- use a separate `.css` file imported from `AppShell.tsx`, this wouldn't cause any conflict/scoping issues as all of these styles use IDs

packages/skeleton-svelte/src/components/AppShell/AppShellFooter.svelte
```diff
@@ -0,0 +1,40 @@
...
+<!-- @component AppShell footer subcomponent. -->
+
+<footer id="appShell-footer" bind:offsetHeight={ctx.footerHeight} class="{rxSticky} {rxHideOnScroll} {rxHidden} {classes}">
+	{@render children()}
+</footer>
+
+<style>
+	footer {
+		grid-area: footer;
+	}
```
rChaoz, line 39: The selector could be `#appShell-footer`, in Svelte this doesn't matter since it's auto-scoped anyways

packages/skeleton-svelte/src/components/AppShell/AppShellHeader.svelte
```diff
@@ -0,0 +1,51 @@
...
+</script>
+
+<!-- @component AppShell header subcomponent. -->
+
+<header id="appShell-header" bind:offsetHeight={ctx.headerHeight} class="{rxSticky} {rxHideOnScroll} {rxHidden} {classes}">
+	{@render children()}
+</header>
+
+<style>
+	header {
```
rChaoz, line 48: Same as above

packages/skeleton-svelte/src/components/AppShell/AppShell.svelte
endigo9740: Per my other comment - styles blocks are a no-go in React 18, so another option will need to be found. Not sure I like the idea of an extra stylesheet requirement. I'm aware grid-template isn't possible with pure Tailwind, but not sure I like the idea of another `.css` import either. At some point you have nearly the same amount of steps as just implementing one of our canned custom layouts which makes me reconsider the value of this feature to users.

rChaoz: Actually.... it turns out [it is, in fact, possible](https://play.tailwindcss.com/yDDn608PWv). I'll replace the custom styles with Tailwind.

endigo9740: Oof, that's a beefy class lol. I'll have to sit on that one for a moment to decide if I like that or not.

rChaoz: Indeed it is, but I don't really see an issue with it. I though of it to, but the end of the way it provides he exact functionality we need in a cross-platform way, and the end user will never be affected by it.

At the end of the day, it's identical code as before, just in a different place.

endigo9740: I was looking at it as something we might expose as a prop for users to adjust, but I suppose if we're only using it as an implementation detail it'll be alright. Just a bit convoluted to read and maintain.

endigo9740:
@rChaoz it's just dawned on me you're implementing this as a Skeleton component feature - where as in our original discussion I mentioned this would only be allowed as a Cookbook guide. It's fine if you want to develop the feature like this, then convert it to a guide - but just want to make sure I'm explicit we will not be implementing this as a Skeleton standard feature ever again.

https://github.com/skeletonlabs/skeleton/issues/3428#issuecomment-2734702757

> Essentially, if you're willing to contribute this as a Cookbook guide, I'd be interested. If you are only willing to submit this as a component, I don't think would be valuable for the reasons stated above.

Just want to clarify as the PR indicates this as a "features" `feat/` in a few places.

rChaoz:
I'm not going to lie, I'm a bit perplexed right now. From our earlier discussions, it seemed like we were on the same track.

Regardless, I would like to ask again what exactly are the _current_ reasons for not wanting this as a component. The reasons you started earlier are:

1. > I think the flaw here is expecting users to be ok with the inflexibility of this specific example. One of the core appeals to the AppShell was you could easily create 10+ different layouts by mixing and matching slots and styles. What you're providing would be a very fixed layout, that from what I can tell doesn't include sidebars.

   This is no longer the case, this AppShell supports every use case that the previous AppShell supported, even more in fact. And if you feel like it's missing something, just let me know and I will add it. 

2. > Because the easiest way to add sidebars requires you to abstract the main to a scrollable child, but the moment you do that it detects scroll from <body>, which where the majority of problems come in. Especially in terms of a11y, printings, etc.

   This is the reason the old AppShell was removed, but the whole point of this new AppShell is that this issue was resolved.

3. > We gotta remember, Skeleton is not Svelte-only any more - features that can't or won't support other frameworks are DOA for us. So bare minimum we would need a React version for now. And the challenge with React is they do not have the same concept of named slots (or snippets if you will).

   I resolved this by removing the need of global styles and using your suggestion of subcomponents, e.g. `<AppShell.Header>`, and data sharing via context. Right now, this can be fully ported to React, Vue and whatever else without any drawbacks.

In short, I believe I've resolved all issues you've mentioned in the linked comment, but correct me if I'm wrong.

I'd also like to add the value I see in this as a component.

First, this appeals to new users a lot. As a new user, what I want from an UI library is to be able to quickly build an UI in 5 minutes, and customise it infinitely later. I think Skeleton fits this gap very well, except for a base layout, which is where AppShell comes in.

As for customisability, I've followed the Skeleton conventions of using Tailwind for all styles, allowing for pretty much infinite customisability. If this is somehow not enough, the layouts guide will not disappear, and this AppShell is incredibly easy to migrate from, as it's just top-level HTML elements, unlike the previous version. So, there is very little to lose for anyone here. 

Finally, I don't see the value of a cookbook guide here, the layout guide already does a good job of explaining how to achieve similar results, and I don't think I can add much on top of it.

endigo9740:
I felt like I was being very clear when I said this could be submitted as a Cookbook guide or not at all. You agreed, then submitted it as a component against my request. The cookbook guide should be showing folks how to ~roll their own~ AppShell component. We are not keen to re-implement, maintain, or support this as a component in the library ever again.

Trying to address the short list of issues I expressed with AppShell was not permission to bypass this explicit requirement and submit this as a component instead. I think you may have mistakenly interpreted those as being the *only issues* with this feature, but that's simply not the case.

Personally I think the AppShell is a flawed concept and has no place for the majority of users. It abstracts away too much away and becomes a black box. If users can't easily create a certain non-standard layouts with AppShell, then it's been my observation that many will assume said layout is not possible - which is devastatingly wrong.

Pair that with the accessibility and other technical issues, and that's why we're now pushing so hard for custom layouts. I'd rather "teach a man how to fish" here rather than "give a man a fish".  Just because something is easier, doesn't mean it's good or healthy to add. Skeleton is a *design system* first, not a component library. We are not trying to solve every challenge in web interfaces with components.

I've made an exception for the Cookbook recipe because it still requires users to at least look at the source and consider how it's constructed. Some level of critical thinking is required, though no doubt some will still blindly copy/paste. So this is leniency for users that feel it's necessary, not complacency in adding what I feel is a *bad* feature.

I'm really keen for folks to design their own custom apps with any and all crazy custom layout they can think of, standard or not,. And we'll provide guidance for supporting that. Not focus on limited component that boxes them in.

endigo9740:
Furthermore Chaos, I feel like we've butted heads a lot this last week or two. I feel like it's my job as the core maintainer to set the rules and ensure folks are following them. I'm one person and only human, not everyone is going to agree with my opinions and decisions, and that's ok. But I think it's important to remember you're working in a collaborative environment. So a couple things here:

1. If you don't agree with something, me and the team are open to discuss further - just bring it up. But simply bypassing requirements you don't like or agree with it's not the right way to go about this.
2. Next week we're going to have an internal team meeting where I'm going to bring up governance of the project. I'd like to try to dimension my voice, at least to the point there's a few folks that weigh decisions alongside me. So not everything rests purely on *my* shoulders. Because I'm aware I can be quite opinionated at times, and that my goal is to do what's best for the project itself.

Just want to emphasize that second point because you may not agree with my opinion and my thoughts, but also indicate that things may change going forward. But both points are worth considering because they boil down to the same thing - let's try and work together to move the project forward.

rChaoz:
I totally understand your point, I'm sorry if it seemed like I was trying to bypass permission. I just missunderstood that "AppShell as a component has x, y and z problems" meant that it _could_ be a component, if not for said problems.

> I feel like it's my job as the core maintainer to set the rules and ensure folks are following them

I definitely agree with this, without proper rules the project can easily get overloaded/messy from various random contributions that pile up. Personally, I thought an AppShell fit because it was in v2, albeit removed in v3 due to a [list of issues](https://github.com/skeletonlabs/skeleton/issues/2383) that I fixed. I'm totally okay with your decision, and didn't mean to push to add it as a component. However, a cookbook guide feels like simply duplicating the existing layouts guide, so to me, that's not a great idea either.

> Furthermore Chaos, I feel like we've butted heads a lot this last week or two.

I hope it doesn't feel that way and I'm very sorry if it does. I tend to be a tad pushy about certain things I believe in, and you should say to me _it's not going to be a component_ (like you said here), rather than _I don't think would be valuable_, to make sure I stop. Also, it's Chaoz, [with a z](https://www.newgrounds.com/audio/listen/85046) 🙂

endigo9740:
> Also, it's Chaoz, [with a z](https://www.newgrounds.com/audio/listen/85046) 🙂

Sorry about that, I'm responding on my living room PC which uses a TV a bit further away. Might be time to schedule an appointment with the eye doctor!

> I hope it doesn't feel that way and I'm very sorry if it does.

Water under the bridge, don't worry about it. I genuinely appreciate all feedback and opinions, even if they differ from my own. It's always nice to have new contributors with big ideas. but I know it can be frustrating to have those ideas and not see them acted on right away. Just know I'm always taking in feedback, just sometimes it takes a bit to process and act on.

Also know I'm usually drawing from observations made in some cases over years of time. I'm in a unique position to have seen the project from day zero to present, and have hands on a bit of everything.

AppShell just happened to be a major pain point for us for a long time, so I'm not keen to revisit or maintain it in the short term if we can avoid it. But at the same time I know folks want it, or something akin to it, so hopefully the Cookbook guide can be a happy medium.

rChaoz:
As per the reasons above. I'll leave the branch up in case anyone wants to use it as a starting point.

Issue #3457, opened 2025-03-25
Customizable Select for Chrome


endigo9740:
### Describe the feature in detail (code, mocks, or screenshots encouraged)

We should aid to adjust Select styles to adhere to these new options
https://nerdy.dev/customize-a-select

Details here:
https://developer.chrome.com/blog/a-customizable-select

### What type of pull request would this be?

Enhancement

### Provide relevant links or additional information.

Chromium based browsers represent the majority of web uses, so this is a big improvement that should hopefully work as progressive enhancement and be supported by Safari/Firefox in the future.

endigo9740:
Given this is so cutting edge and some potential issues with Svelte (linked above), I'm gonna close this and give things a bit more time to cook. Seems like a great enhancement for the future though.

PR #3459, opened 2025-03-26
fix: select option checked styling

robert-wettstaedt:
## Description

A styling for checked options is missing. It only shows the checked option(s) when the select is focused.

## Checklist

### DIFF
--- a/packages/skeleton/src/utilities/form-selects.css
+++ b/packages/skeleton/src/utilities/form-selects.css
@@ -96,5 +96,9 @@
 		line-height: --spacing(9);
 		height: --spacing(9);
 		padding: --spacing(2);
+
+		&:checked {
+			background-color: var(--color-primary-500);
+		}
 	}
 }


endigo9740:
Hey @robert-wettstaedt, while your contribution is likely well intentioned, unfortunately we have a couple other initiatives in the pipeline that overlap with this.

1. We typically don't allow "drive-by" PRs. These can often overlap with existing initiatives, as is the case here.
1. Select elements are notoriously hard to style properly, and we've actually removed similar checked style recently for that very reason. They were causing a11y issues. While this may work in your browser of choice, it'll cause issues in others.
2. I'm preparing to implement some new styling techniques made possible or [Chromium-based browsers](https://github.com/skeletonlabs/skeleton/issues/3457) this very week!

A couple other notes on this:

- Again, Select elements have historically been one of the hardest elements to style consistently cross-browser/device
- These new Chromium updates will hopefully allow for progressive enhancement, even if limited to a subset of the web
- Long term, there's an initiative to replace these with a more flexible element aptly named [customizable select](https://developer.chrome.com/blog/rfc-customizable-select), which should go even further!

So in short, thanks for your contribution, but I cannot move forward with it. I'd suggest keeping an eye on #3457 as an alternative though!

Issue #3471, opened 2025-03-29
Expose all types/props


Keagel:
### Describe the feature in detail (code, mocks, or screenshots encouraged)

Currently only some select types are exported: https://github.com/skeletonlabs/skeleton/blob/main/packages/skeleton-svelte/src/index.ts

Exporting all props/types/interfaces would allow us to extend the Skeleton components without having to manually redefine every prop of a component.

### What type of pull request would this be?

Enhancement

endigo9740:
I'm going to go ahead and close this out, as we have a definitive answer to this in the upcoming Skeleton v4 release. You should be able to directly import any component props as expected.

<img width="1104" height="290" alt="Image" src="https://github.com/user-attachments/assets/e244bab9-10b8-439c-bb1c-5165d49eaa76" />

Issue #3339, opened 2025-03-08
Proposal/RFC: Tailwind Merge pattern for native Tailwind v4


endigo9740:
@Hugos68 came across this new technique that could potentially allow us to rethink the way you apply classes to components via props.

https://x.com/diegohaz/status/1897776774206853269

## What This Does

Currently we have the concept of `base | {property} | classes`, but this would allow us to streamline this down to ONLY `classes` because base and property classes could be applied inline in the HTML template, but overwritten with classes that take precedence without the need of third-party tools like Tailwind Merge or using `!`.

So here's what that would look like from the user's perspective:

```html
<Avatar background="bg-green-500" classes="border border-white" />
```

And after the change:

```html
<Avatar classes="bg-green-500 border border-white" />
```

## What is NOT Changed

We would still need the prefix approach to target multiple elements within the component template. Like:

```
classes
imgClasses
fallbackClasses
```

## Downsides / Pitfalls

Unfortunately this does have a slight cost to the final bundle size. Every time we use a native Tailwind utility like `base:p-4`, this does not resolve down to the same instance as `p-4`. So every `base:` class we add in our components would be generate "dupes".

He's a trivial example using `bg-red-500`:

https://play.tailwindcss.com/5DRJbdc3HW

Hugos68:
I think we should move away from `classes` if we were to use this because `classes` isn't picked up by Tailwind Intellissen, if we did `class` and `<element>Class` (or imo, just use components and always use `class`) we could instruct users to only add `*Class` to their tailwind extension config.

endigo9740:
My biggest concern with that was that `class` typically requires some kind of work around because it's a reserved name. We would work around that for the root element, but the each child would implemented differently. Where as if we just keep them custom props it's consistent across the board, across all frameworks. But I do think folks would respond better to the singular version.

Hugos68:
No workarounds needed:
```html
<script>
	const props = $props();
</script>

<div class="base:bg-red-500 {props.class}">
	<div class={props.labelClass}>test</div>
</div>
```
```jsx
export default function(props) {
  return (
    <div className='base:bg-red-500 {props.class}'>
      <div className={props.labelClass}>test</div>
    </div>
  );
}
```

endigo9740:
That's 1/4 frameworks. Now we need React, Vue, and Solid too.

Hugos68:
@endigo9740 I showed you react and solid already, look again. Vue is the same aswell. Nothing changes.

[...]

Issue #3230, opened 2025-02-18
v3: Avatar component does not accept crossorigin prop or actions.


itzTheMeow:
### Describe the feature in detail (code, mocks, or screenshots encouraged)

In v2, you were able to pass an action to the `Avatar` component. This was useful for performing actions on the image after it loaded. In v3 however, this property has been removed. Also in v2, you could set additional props on the avatar such as `crossorigin="anonymous"`, as they were passed to the internal img element. In v3 this is removed.

I'd like to request these features be added back in v3.

### What type of pull request would this be?

Enhancement

itzTheMeow:
Additionally, I'd be willing to submit a PR for this feature if needed.

endigo9740:
Just keep in mind that Actions are not attributes, they are functional implementation that the Svelte compiler interfaces with. AFAIK they cannot be passed as a arbitrary prop. We have [been pondering a convention](https://github.com/skeletonlabs/skeleton/issues/2703) for passing arbitrary props to components that require them, but I'm not sure we've settled on one yet. Ideally this needs to satisfy a few requirements:

- Ideally it's a convention that works cross-framework (Svelte 5 is closer to React in this regard now)
- If possible we should be able to target arbitrary properties to a specific element. In case more than one need this. (think image + input in a single component)
- Rich Harris recently teased an alternative for Actions in Svelte that might solve some of this for us. Unfortunately I don't have a link and can't recall the name off hand. His RFC should still be somewhere on their GitHub.

I know this seems like a trivial add, but any PR must take into account the bigger picture here.

In the meantime, you could consider using the Fallback slot with your own `<img>` element. This would give you full control over it. In fact this is what we will be recommending for Image components like `enhance:image` for Svelte and the `<Image>` component from Next.js.

itzTheMeow:
The action was passed through the `action` prop on the element in v2, it wasn't arbitrary sorry.

Arbitrary props aren't completely necessary, but a crossorigin one could be implemented for avatars and other image-based components.

endigo9740:
@itzTheMeow I'll quickly reiterate that the team and I are still evaluating this and similar issues as part of a larger sweeping change to component APIs in the future. Unfortunately I don't think we're going to be able to make headway on this prior to the v3 launch [coming next week](https://github.com/skeletonlabs/skeleton/discussions/3315).

So I'm going bump this to our post-release milestone to revisit in the future. My advise in my prior message will need to suffice as the interim solution for now:

[...]

Issue #3193, opened 2025-02-04
Review VS Code Intellisense auto-complete for Tailwind classes


endigo9740:
### Describe the feature in detail (code, mocks, or screenshots encouraged)

In versions prior to Skeleton v3 we implemented a build-time process that would scan all components and generate a unique list of props that utilize the special `CssClass` type. This class simply represented a string, but by doing this we could generate a JSON file like this:

https://raw.githubusercontent.com/skeletonlabs/skeleton/master/packages/skeleton/scripts/tw-settings.json

This was then used as part of the Skeleton CSA (aka v2 CLI) to insert this file into user's Skeleton project. This would live here:

```
./.vscode/settings.json
```

By doing this, VS Code can identify all Skeleton props that should take CSS classes, allowing for Intellisense auto-complete.

### Goal

In v3 we should try to find a way to generate similar JSON files for each component framework - since prop names can differ. Then providing a manual (but optional) onboarding step for users that wish to install this in their project and gain this functionality. Perhaps look at if similar functionality is available for other popular editors as well.

Then long term, we should revisit moving this to a dedicated Skeleton VS Code extension so users no longer have to manage this on their own.

> NOTE: The CSA approach will never be possible again as we will never build tooling for scaffolding user apps again. It's just not possible when supporting multiple frameworks and meta-framework combinations.

### What type of pull request would this be?

New Feature

oatmealproblem:
Hey, I'm interested in working on this. We chatted in Discord, but I'll repeat myself a bit so the conversation is followable here.

I've implemented a script that extracts the CSS class props from Skeleton, so they can be fed into VSCode's `tailwindCSS.classAttributes` setting or prettier-plugin-tailwind's `tailwindAttributes` setting.

https://gist.github.com/oatmealproblem/3792d16e52c8b95db86cf81efdffcb52

I implemented that as a 3rd party script that works with the current Skeleton dist. It has some limitations, notably it's a bit "fuzzy" in finding the CSS class props:
- it finds exported interfaces with names ending in `Props`
- within those interfaces, it finds optional string props whose attached JSDoc comment includes the text "class", "css", or "style"

For a first party solution, that "fuzziness" can be avoided. You mention that Skeleton v2 used a custom `CssClass` type for this (simple an alias of `string`?). Is that the approach we'd like to use for v3 as well, or were there downsides to that?

Alternatives:
- use a custom jsdoc tag (eg `@css-class-prop`)
[...]

PR #3491, opened 2025-04-06
feat: experimental component API

### LINKED ISSUE #3463: Proposal: Component API Overhaul
## Goal

While we're pleased the component APIs for v3 and the recent Zag.js integration, @Hugos68 has identified a few things that could be improved that would provide a number of quality of live benefits. Unfortunately these *are* breaking changes, and as such would need to be reserved to a major release.

## Proposals

The following are based on [Hugo's suggestions](https://gist.github.com/Hugos68/7624bc59f9b7f7e62d02beefae832676).

### 1. Single Entry Point for Classes

This would utilize either the `base:` variant strategy -or- Tailwind Merge, to replace `base/property` style props with only a single entry point. This would be similar to the v3 `classes` (plural) prop, but likely resolve to the nature `class` (singular) attribute name. Essentially ALL inherent styles would be implemented in such a way that user styles always take precedence and either overwrite existing classes, or if new, extend the class list.

### 2. More Modular and Composed Components

Components would now be composed of more/smaller child components, ala [Bits UI](https://www.bits-ui.com/docs/components/accordion#quick-start). This means components are composed of smaller pieces in a template-driven manner, rather than relying on fewer/larger components that implement features such as named slots and snippets. This more closely matches the HTML template structure, which not only benefits Zag.js integration, but also minimizes the "depth" of elements/classes assigned per component.

### 3. Improved Support for "Rest" Props/Attributes

The other advantage of more modular/composed components is it would be easier to provide and spread restProps. This can include a number of common props like `style` attributes, `lazy` load attributes for images, plus tons of input-related attributes.

### 4. Container Queries for Responsive Sizing

I think this would be a great time to convert to container-based queries instead of the media query breakpoints we use now. Work with Chris if you need help getting started on this.

https://tailwindcss.com/docs/responsive-design#what-are-container-queries

## Prototype ASAP In Public

Rather than writing up a length proposal and gathering community comments, @ryceg has suggested a novel approach. Which is to implement this for one, singular, new component. Ideally a clone of an existing component to compare to. That way users test drive this in practice right away. This component would be marked with an `alpha` or `experimental` tag to make it clear it's subject to changes. If the majority of the community agrees with these changes, then this would become the focus of the v4/Next release, spearheaded by @Hugos68  

## Additional Information

Here's an example of what the changes might look like from user-land...

Before (v3.0)

```html
<Avatar src="..." background="bg-primary-500" imageClasses="greyscale" border="border border-surface-50-950" />
```

After (v4/Next)

```html
<Avatar class="bg-primary-500 border border-surface-50-950">
    <Avatar.Image src="..." class="greyscale" lazy></Avatar.Image>
</Avatar>
```

- More piecemeal component structure to more closely match the HTML template structure (ala Bits UI)
- All style props are collapsed to the singular `class` entry point prop
- The `class` doesn't care what or how many classes you pass, they always take precedence
- Functional props like `src` are moved relative to usage - so `Avatar.Image` in this case
- Demos restProps such as `lazy` ton the `` tag (a native HTML attribute)

Hugos68:
## Linked Issue

Closes #3463 

Also solves/closes:
- https://github.com/skeletonlabs/skeleton/issues/3193
- https://github.com/skeletonlabs/skeleton/issues/3471
- https://github.com/skeletonlabs/skeleton/issues/3339
- https://github.com/skeletonlabs/skeleton/issues/3230
- https://github.com/skeletonlabs/skeleton/issues/2982
- https://github.com/skeletonlabs/skeleton/issues/2811
- https://github.com/skeletonlabs/skeleton/issues/3514

## Remaining Tasks

- [x] Catch up to the latest updates on main branch (done on Aug 14, 2025)
- [x] Flush out the docs for the new Avatar/Accordions (should be equivalent to the old component docs)
- [x] Set at least one rudimentary test case per Avatar/Accordion per framework
- [x] Shorten the doc navigation category header so it doesn't wrap (open to suggestions)
- [x] Chris: Introduce a new "Fundamentals" guide showcasing the usage and advantages of the changes

NOTE: we will revisit custom animations starting this week. Nothing to share at this time.

## Review the RFC

- https://github.com/skeletonlabs/skeleton/discussions/3499

## Description

This PR adds a `composed` folder with all of our components using the new component API described in #3463.

### API
These components will become available (when this PR is merged) under an alternative export path for users who want to test the new API.

```js
import { Accordion } from '@skeletonlabs/skeleton-react/overhaul'
```
```html
<Accordion>
	<Accordion.Item value="item-1">
		<Accordion.Heading>
			<Accordion.Trigger>Item 1</Accordion.Trigger>
			<Accordion.Content>Content for Item 1</Accordion.Content>
		</Accordion.Heading>
	</Accordion.Item>
	<Accordion.Item value="item-2">
		<Accordion.Heading>
			<Accordion.Trigger>Item 2</Accordion.Trigger>
			<Accordion.Content>Content for Item 2</Accordion.Content>
		</Accordion.Heading>
	</Accordion.Item>
	<Accordion.Item value="item-3">
		<Accordion.Heading>
			<Accordion.Trigger>Item 3</Accordion.Trigger>
			<Accordion.Content>Content for Item 3</Accordion.Content>
		</Accordion.Heading>
	</Accordion.Item>
</Accordion>
```

### Features:
- Piecemeal components which allows for more fine grained (less abstraction).
- Full `restAttributes` support, components are now simply wrappers around HTML elements and allow you to define any attribute like a normal HTML element.
- Implementation of `base` styling system where we get rid of all the specific style props and introduce a single `class` (or `className` in React) prop that uses either the `base:` directive or `twMerge` (up for discussion).
- Implement the `child` snippet pattern, see: https://bits-ui.com/docs/child-snippet

### Ccomponents
| Component | React | Svelte |
| - | - | - |
| Accordion | Done | Done |
| Avatar | Done | Done |


packages/skeleton-react/src/composed/accordion/anatomy/accordion-content.tsx
```diff
@@ -0,0 +1,20 @@
+'use client';
```
endigo9740, line 1: Just a random aside here, but I think some React purists will have some strong reactions to multi-file components. I've heard many mention one of the benefits for React, for them at least, is all this defined in one file.

We should maybe look at Mantine, Shadcn, etc and follow their lead in this regard too.

Though personally I prefer multi-file too!

packages/skeleton-react/src/composed/accordion/modules/anatomy.ts
```diff
@@ -0,0 +1,16 @@
+import AccordionRoot from '../anatomy/accordion-root.js';
+import AccordionItem from '../anatomy/accordion-item.js';
+import AccordionHeading from '../anatomy/accordion-heading.js';
+import AccordionTrigger from '../anatomy/accordion-trigger.js';
+import AccordionIndicator from '../anatomy/accordion-indicator.js';
+import AccordionContent from '../anatomy/accordion-content.js';
+
+const Accordion = Object.assign(AccordionRoot, {
```
endigo9740, line 8: Kind of random thought here, but can we set or extend the `Object.assign()` to a bit more metadata? The idea being, that this may be something we can utilize to more programmatically define the `<Anatomy>` diagram on each doc page.

packages/skeleton-react/src/composed/accordion/modules/types.ts
```diff
@@ -0,0 +1,37 @@
+import * as accordion from '@zag-js/accordion';
+import type { ButtonHTMLAttributes, HTMLAttributes } from 'react';
+import type { PropsWithChildOrChildren } from '../../../internal/props-with-child-or-children.js';
+
+interface AccordionRootProps
```
endigo9740, line 5: Let's make sure we're exposing component types to resolve the recent requests for such. I believe we have a pending ticket for this too, which you might can add to to the list this change solves.

packages/skeleton-react/src/composed/accordion/modules/types.ts
```diff
@@ -0,0 +1,37 @@
...
+		Omit<HTMLAttributes<HTMLDivElement>, 'id' | 'defaultValue' | 'dir'> {}
+interface AccordionItemProps extends PropsWithChildOrChildren, accordion.ItemProps, HTMLAttributes<HTMLDivElement> {}
+interface AccordionTriggerProps extends PropsWithChildOrChildren, ButtonHTMLAttributes<HTMLButtonElement> {}
+interface AccordionHeadingProps extends PropsWithChildOrChildren, HTMLAttributes<HTMLHeadingElement> {
+	/**
+	 * The level of the heading. This is used to determine the heading level for accessibility purposes.
+	 *
+	 * @defaultValue 3
+	 */
+	level?: 1 | 2 | 3 | 4 | 5 | 6;
```
endigo9740, line 17: Just to verify, this is *only* used for the `<hX>` tag, and not for the class itself? Because Tailwind won't accept constructed classes like that.

packages/skeleton-react/src/composed/accordion/modules/anatomy.ts
endigo9740, line 8: Additionally, I might suggest we set some conventions for component names. To begin we can start with something like this...

1. We should default to whatever Zag refers to the element. This will provide the simplest option. No thinking involved.
2. If Zag doesn't name the element, we should try to find a common reference. For Accordion this could be the HTML Disclosure element. See below.

- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/summary
- https://tailwindcss.com/blog/tailwindcss-v4-1#a-bunch-of-other-new-variants

So in the case we would end up with something more like this:

```
- Accordion.Root
-- Accordion.Item
--- Accordion.Summary
--- Accordion.Content
```

packages/skeleton-react/src/internal/props-with-child-or-children.ts
```diff
@@ -0,0 +1,6 @@
+import type { PropsWithChildren } from 'react';
```
endigo9740, line 1: We want users to be able to follow our conventions for implementing their own custom components. As such we likely should limit internal-only utilities, and try to both make public, and also document how they operate.

At hoorai we have a slew of custom components and I'd want to ensure we didn't have to reinvent all these. Especially since I noticed Svelte has several additional utilities.

playgrounds/skeleton-react/src/app/composed/accordion/page.tsx
```diff
@@ -0,0 +1,28 @@
+'use client';
```
endigo9740, line 1: The playgrounds are a great starting place while we update and refine, but just want to make it clear we'll need to document in the main website docs too. I figure we create a new navigation category like:

- `/components`
- `/components-compound`

Then in the visible Navigation component, set the titles as:

```
Functional Components

...

Compound Components (Beta)

...
```

This will be temporary as don't think we can fit "Functional Components (Beta)" as is.

endigo9740 (changes requested):
@Hugos68 here's my initial review.

packages/skeleton-react/src/composed/accordion/modules/anatomy.ts
Hugos68: What metadata are we missing?

packages/skeleton-react/src/composed/accordion/modules/types.ts
Hugos68: Yep, only the tag, we can't assume people are using our typography styles, right? Or do we want to?

packages/skeleton-react/src/internal/props-with-child-or-children.ts
Hugos68: I'm unsure what you mean with this, are you talking about the PropsWithChildOrChildren helper type? Since that's really only for us. Users won't have any benefit having it exposed

packages/skeleton-react/src/composed/accordion/modules/anatomy.ts
endigo9740: Zag doesn't provide data to all portions of the template, so some of that is left up to us:

```tsx
<div {...api.getRootProps()}>
  {data.map((item) => (
    <div key={item.title} {...api.getItemProps({ value: item.title })}>
      <h3>
        <button {...api.getItemTriggerProps({ value: item.title })}>
          {item.title}
        </button>
      </h3>
      <div {...api.getItemContentProps({ value: item.title })}>
        {item.content}
      </div>
    </div>
  ))}
</div>
```

packages/skeleton-react/src/internal/props-with-child-or-children.ts
endigo9740: You can tag a directory or comments (afaik) so I tagged a random `/internal`  utility script here. But just saying in general if there's tools we're using to build the components, we should expose those to the end users. Not this file/script specifically.

packages/skeleton-react/src/composed/accordion/modules/anatomy.ts
endigo9740:
```html
<Anatomy label="Accordion" isComponent>
	<Anatomy label="Accordion.Item" isComponent>
		<Anatomy label="control" tag="button">
			<Anatomy label="lead" />
			<Anatomy label="content" />
			<Anatomy label="indicator" />
		</Anatomy>
		<Anatomy label="panel" />
	</Anatomy>
</Anatomy>
```

- `isComponent` - or some way to differentiate between components/snippets
- `tag` - to identify the semantic HTML used
- `descriptor` - which is used infrequently (see Pagination)

endigo9740: Actually come to think about it, the `Object.assign()` isn't nested, it just separate root/parent vs children. So nix this idea.

oatmealproblem:
One potential concern about this `base:` classes approach: how much will it increase the size of CSS for apps that use skeleton?

With the current v3 approach, all the classes are potentially re-used. Eg, if skeleton base styles something with `round-container` and I style something with `round-container`, we're both using the same class so that's just defined once.

But in this proposed approach, skeleton styles with `base:round-container` and I style with `round-container`, those are separate classes, so won't that essentially be duplicated? Worst-case scenario, it could double the amount of CSS needed?

packages/skeleton-svelte/src/composed/accordion/anatomy/accordion-heading.svelte
```diff
@@ -0,0 +1,15 @@
+<script lang="ts">
+	import type { AccordionHeadingProps } from '../modules/types.js';
+
+	const props: AccordionHeadingProps = $props();
+	const { level = 3, child, children, ...restAttributes } = $derived(props);
+	const tag = $derived(`h${level}`);
+</script>
+
+{#if child}s
```
oatmealproblem, line 9: You got a rogue `s` here

Hugos68:
@oatmealproblem That's an excellent question we immediately asked ourselves, so I did some investigation, you can see our findings here: https://gist.github.com/Hugos68/8a5e2ee2f53522a350ac70ae20c2dcfc

packages/skeleton-react/src/composed/accordion/modules/anatomy.ts
Hugos68: I think the current anatomy naming is on track with Zag's naming, Zag doesn't have a getHeadingProps or anything because the heading doesn't require any props, but according to the WACG spec it should be there. So `Heading` seems logical, no? The rest of the components closely follow Zag's naming.

packages/skeleton-react/src/composed/accordion/anatomy/accordion-content.tsx
Hugos68: Sure, but we can refer to other React component libraries and argue it's more consistent for us when jumping from one framework to another, it has no impact on users.

packages/skeleton-react/src/composed/accordion/modules/anatomy.ts
Hugos68: We can still attach information, it's definitely not a crazy idea! For example:

```js
const Accordion = Object.assign(AccordionRoot, {
	Item: Object.assign(AccordionItem, {
		[' metadata']: {
			tag: 'div'
		}
	}),
	Heading: AccordionHeading,
	Trigger: AccordionTrigger,
	Indicator: AccordionIndicator,
	Content: AccordionContent
});
```

Notice the empty space I put before the prop name, this removes the prop from intellisense!

image

So end users aren't aware.

But when we access it:
image

we still get type safety and extra info about each component.

oatmealproblem:
> @oatmealproblem That's an excellent question we immediately asked ourselves, so I did some investigation, you can see our findings here: https://gist.github.com/Hugos68/8a5e2ee2f53522a350ac70ae20c2dcfc
> 
> TL;DR The bundle size tradeoff is quite insignificant (Especially compared to similar solutions like `tailwind-merge).

Thanks, looks good! Glad to see that was already considered

endigo9740:
@oatmealproblem the new RFC is now live covering these changes:

- https://github.com/skeletonlabs/skeleton/discussions/3499

packages/skeleton-svelte/src/composed/accordion/anatomy/accordion-heading.svelte
```diff
@@ -0,0 +1,15 @@
...
+	import type { AccordionHeadingProps } from '../modules/types.js';
+
+	const props: AccordionHeadingProps = $props();
+	const { level = 3, element, children, ...restAttributes } = $derived(props);
+	const tag = $derived(`h${level}`);
+</script>
+
+{#if element}
+	{@render element({ attributes: restAttributes })}
+{:else}:
```
endigo9740, line 11: There's an extra colon after the else statement here. This is visible when previewing the component in the Svelte playground page.

playgrounds/skeleton-svelte/src/routes/composed/accordion/+page.svelte
```diff
@@ -0,0 +1,28 @@
+<script lang="ts">
+	import { Accordion } from '@skeletonlabs/skeleton-svelte/composed';
+</script>
+
+<Accordion>
+	<Accordion.Item value="item-1">
+		<Accordion.Heading>
+			<Accordion.Trigger>
+				{#snippet child({ attributes })}
```
endigo9740, line 9: Two issues here:

1. There's a type error, seems to be `any` type
2. The custom markup doesn't appear to actually render in the Svelte playground page

Screenshot 2025-04-09 at 1 10 41 PM

packages/skeleton-svelte/src/composed/accordion/anatomy/accordion-item.svelte
```diff
@@ -0,0 +1,33 @@
...
+
+	const rootContext = AccordionRootContext.consume();
+	const props: AccordionItemProps = $props();
+	const [itemProps, componentProps] = $derived(accordion.splitItemProps(props));
+	const { element, children, ...restAttributes } = $derived(componentProps);
+	const attributes = $derived(
+		mergeProps(
+			rootContext.api.getItemProps(itemProps),
+			{
+				class: 'base:grid gap-2'
```
endigo9740, line 15: Should we use `base:gap-2` here?

endigo9740 (changes requested):
FYI I've pushed a change that updates the playground navigation to include a "Composed (Experimental)" category. I've gone ahead and linked the Accordion for React and Svelte.

A few things though...

## 1) Avatar Playground Pages

Let's go ahead and add the Avatar playground pages. It'll be easier to review these and ensure they are working as expected. I caught a couple small issues with the Accordion just by doing this.

## Classlist

Are we forced to define `base:` classes in `mergeProps()` or can these be defined in the HTML template? I think I'd prefer the template so we get auto-suggestions from the Tailwind extension. We only abstracted styles up to the script block because they were props previously.

## Animations

We need to figure out a way to use an wrapping `#if` statement for `Accordion.Content` to we can utilize Svelte transitions. For React we'll need the `<Presence>` component ported from Zag.

Alternatively we can try to come up with some kind of universal strategy using pure CSS. Something that could work for all frameworks. But something to noodle on.

packages/skeleton-svelte/src/composed/accordion/anatomy/accordion-item.svelte
Hugos68: Yep, mb, will fix

packages/skeleton-svelte/src/composed/accordion/anatomy/accordion-heading.svelte
Hugos68: Oh wth, ill fix right away

playgrounds/skeleton-svelte/src/routes/composed/accordion/+page.svelte
Hugos68: Second issue is fixed, can't figure out why the first issue happens.

Hugos68: Finally resolve the type issue, typescript is my enemy

Hugos68: @TGlide Helped me figure it out! It was `moduleResolution` set to `NodeNext` causing issues. I've set it to `bundler` for now which works fine.

.vscode/settings.json
```diff
@@ -5,5 +5,6 @@
 	"eslint.enable": true,
 	"editor.codeActionsOnSave": {
 		"source.fixAll": "explicit"
-	}
+	},
+	"tailwindCSS.experimental.classRegex": ["(?:class|className):\\s*['\"]([^'\"]*)['\"]"]
```
rChaoz, line 9: This exists?? It's not documented anywhere but I can find it with GitHub code search.. why is this not documented, would have saved so many lives with current Skeleton like `.*Classes|.*Base`

packages/skeleton-svelte/src/composed/accordion/modules/anatomy.ts
```diff
@@ -0,0 +1,16 @@
+import AccordionRoot from '../anatomy/accordion-root.svelte';
```
endigo9740, line 1: We should implement at least one example test case for either component to verify how we'll handle these. We should not re-test Zag features. But we should test custom Skeleton changes - such as the inclusion of a new icon snippet. Or any similar change not inherent to Zag's implementation.

endigo9740 (changes requested):

.vscode/settings.json
Hugos68: Yeah well classes and base would have been fixed yeah but what about all the other properties?

sites/skeleton.dev/src/content/docs/components-composed/accordion/meta.mdx
```diff
@@ -0,0 +1,10 @@
+---
```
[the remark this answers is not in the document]
Hugos68: I've done that now, I've specifically *not* used `error` because the user isn't doing anything wrong and that would give them a sense of wrongdoing. It's merely a warning about a non stable API. So all stability warnings are of the warning color.

packages/skeleton-svelte/src/composed/accordion/modules/anatomy.ts
Hugos68: I'm having a hard time figuring out what to test, I could test that we are propegating properties like class/className accordingly?

PR #3524, opened 2025-04-24
Move Login button into nav element

neknalb:
## Description

It's about the Components Preview on <https://themes.skeleton.dev/themes/create>. The code is in `sites/themes.skeleton.dev/src/lib/components/generator/Preview/PreviewComponents.svelte`. There we have this `nav` element with some buttons.

Currently the "Login" button is placed outside the `nav` element (after it). That seems to be a bug. My PR should fix that.


endigo9740:
@neknalb thanks, but the current structure was intentional. I'm not sure this would have any benefit to end users, as this secure is purely or visual previews for end users. Appreciate the work, but we don't typically make changes unless they are impactful.

PR #3585, opened 2025-07-01
feat: add tree view

### LINKED ISSUE #3468: Add Tree View Component
### Link to the Page

- React: https://zagjs.com/components/react/steps
- Svelte: https://zagjs.com/components/svelte/steps

We recently polled the community and this was the number 3 requested component.

Given the complexity of this one, I'd suggest it be implemented by a member of the Skeleton core team, including @Hugos68 or @phamduylong, with my help for the visual interface design.

Note that our initial public release should be marked with the `beta` label to match Zag's status of this feature. This should remain until a stable release is provided by Zag.

dev-viinz:
## Linked Issue

Closes #3468

## Description

Implements the Zag.js TreeView.
In this draft I'm starting with the svelte component until it's finalized, to then "port" it over to react.
Current Progress:
- [x] Add svelte component
- [x] Svelte component done
- [ ] Add react component
- [ ] React component done
- [x] Add changeset

## Checklist

## Changsets


endigo9740:
Thanks @dev-viinz, FYI you may see some portions of the CI checks fail right now. This is a [known issue and I'm in progress on resolving this](https://github.com/skeletonlabs/skeleton/pull/3584). I just ran out of time for today, so I'll resolve this and provide an initial review of your draft PR here tomorrow. Totally understand if things are a work in progress, so I won't nitpick I promise! Just want to ensure we have a firm foundation established.

Again I'm out for a bit, but I'll try to check in later today if you respond on here or Discord.

endigo9740:
@dev-viinz see my comments in the Contributor section on Discord. Additionally, I ask that you please catch up your feature branch to the latest changes on `main`. This includes fixes that will enable the full CI/CD check process to run successfully. I'll run the process now so you'll see the failure, then once updated this should be all green.

Note that if you still see errors after merging the `main` branch, those will likely be on your end to resolve. You should be able to run `pnpm check` in the root of the monorepo project to see these, otherwise let me know and I can help diagnose what's up.

dev-viinz:
@endigo9740 It seems like the CI tasks that are not Vercel related need an approval after every push, is this intentional?

endigo9740:
@dev-viinz you can see GitHub's docs about this [here](https://docs.github.com/en/actions/how-tos/managing-workflow-runs-and-deployments/managing-workflow-runs/approving-workflow-runs-from-public-forks).

> These workflow approval policies are intended to restrict the set of users that can execute workflows in GitHub Actions runners that could lead to unexpected resource and compute consumption when using GitHub-hosted runners.

We're just using the default configuration, which is a bit more locked down. We don't get so many contributions that it would likely cause a major issue, but that's the reasoning for it.

The list of checks run can be replicated locally using the commands in the Checklist section in the original post above. I'd recommend you run through periodically during development to ensure all is well. I will verify everything is green when I do my full review though.

I'll approve this again so you can see the current status.

pnpm-workspace.yaml
```diff
@@ -3,7 +3,7 @@ packages:
   - 'sites/*'
   - 'playgrounds/*'
 
-define: &zag_version '^1.7.0'
+define: &zag_version '^1.17.4'
```
endigo9740: This is perfect. Can you do me one favor and check if this version includes this change they made for us on the Progress Circular component too:

https://github.com/chakra-ui/zag/pull/2562

This would help us kill two birds with one stone.

packages/skeleton-react/src/components/Segment/Segment.test.tsx
```diff
@@ -1,11 +1,23 @@
-import { describe, expect, it } from 'vitest';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import { render } from '@testing-library/react';
 
 import { Segment } from './Segment.js';
 
 // Segment ---
 
 describe('<Segment>', () => {
+	beforeEach(() => {
```
endigo9740: It looks like this should already be mocked globally per:

`packages/skeleton-svelte/vitest.setup.js`

Perhaps this needs to be extended or modified?

packages/skeleton-svelte/src/components/TreeView/TreeView.svelte
```diff
@@ -0,0 +1,185 @@
+<script lang="ts" module>
+	import { createRawSnippet } from 'svelte';
+
+	const chevron = createRawSnippet(() => {
```
endigo9740: Good call using a raw SVG for a placeholder. Any particular reason to create this snippet in the script tag as opposed to an HTML snippet in the template? The latter might be more legible we'll get syntax highlighting for the SVG tags and attributes.

packages/skeleton-svelte/src/components/TreeView/TreeView.svelte
```diff
@@ -0,0 +1,185 @@
...
+	import { slide } from 'svelte/transition';
+
+	const {
+		// Animation
+		animationConfig,
+		// Data
+		collection,
+		// Root
+		base = 'flex flex-col w-fit',
+		bg = '',
```
endigo9740: Just confirming you did a great job with the style prop conventions. The only one that stands out is `bg`, I think favor `background` in other components.

packages/skeleton-svelte/src/components/TreeView/TreeView.svelte
```diff
@@ -0,0 +1,185 @@
...
+		// Indent
+		indentAmount = '1.2rem',
+		// Indicator
+		indicatorOpenRotation = '90deg',
+		indicatorTransition = 'transition-transform',
+		// Snippets
+		branchIcon,
+		itemIcon,
+		branchIndicator,
+		...zagProps
```
endigo9740, line 26: Nitpick: we tend separate the `zagProps` with a comment above as well.

packages/skeleton-svelte/src/components/TreeView/TreeView.svelte
```diff
@@ -0,0 +1,185 @@
...
+		itemBase = 'flex gap-2',
+		itemBg = '',
+		itemSpaceY = '',
+		itemHover = controlHover,
+		itemBorder = contentBorder,
+		itemPadding = controlPadding,
+		itemShadow = '',
+		itemClasses = '',
+		// Indent
+		indentAmount = '1.2rem',
```
endigo9740: Functional props (basically general settings) can be grouped at top of the prop list. Just a convention, this has no impact on the functionality.

packages/skeleton-svelte/src/components/TreeView/TreeView.svelte
```diff
@@ -0,0 +1,185 @@
...
+			{/if}
+			<span {...api.getItemTextProps(nodeProps)} data-testid="tree-item-text">
+				{nodeProps.node.value}
+			</span>
+		</button>
+	{/if}
+{/snippet}
+
+<style>
+	[data-scope='tree-view'][data-part='branch-indent-guide'] {
```
endigo9740: So this is a bit of a pain point we've encountered with Zag, but we try to avoid `<style>` tags whenever possible. The team and I have struggled with this a lot internally for a few reasons:

1. Plain and simple these don't work in React until the absolute most recent major release (React 18). We have not yet dropped support for versions below 18, so this means we have to find another strategy to apply these styles.
2. Skeleton does not ship a stylesheet specifically for components, so we can't abstract these away to an external/shared stylesheet like `components.css` or similar.
3. If possible, let's try to convert to Tailwind utilities and insert within the template. If not, consider using the approach you did with `indicatorOpenRotation` and update a `style` attribute directly. The latter is the pattern we've used for few components.

packages/skeleton-svelte/src/components/TreeView/TreeView.svelte
```diff
@@ -0,0 +1,185 @@
...
+	{#if api.getNodeState(nodeProps).isBranch}
+		<!-- Branch -->
+		<div {...api.getBranchProps(nodeProps)} data-testid="tree-branch">
+			<!-- Control -->
+			<button
+				class="{controlBase} {controlBg} {controlSpaceY} {controlHover} {controlBorder} {controlPadding} {controlShadow} {controlClasses}"
+				{...api.getBranchControlProps(nodeProps)}
+				data-testid="tree-control"
+			>
+				<span
```
endigo9740: I'd encourage you to add comments denoting each "part" of the template. You've done this for the most part, but some of the lower level children could use this as well. General rule of thumb: if it has a unique Zag property name, or you would set a style prop prefix, then it should have a comment.

This is super helpful for folks like me (or end users) that come in and view the template blind. We can then identify each and every part of the template at a glance.

packages/skeleton-svelte/src/components/TreeView/TreeView.svelte
```diff
@@ -0,0 +1,185 @@
...
+				<span {...api.getBranchTextProps(nodeProps)} data-testid="tree-branch-text">
+					{nodeProps.node.value}
+				</span>
+			</button>
+
+			<!-- Content -->
+			{#if api.expandedValue.includes(nodeProps.node.id)}
+				<div
+					class="{contentBase} {contentBg} {contentSpaceY} {contentBorder} {contentPadding} {contentShadow} {contentClasses}"
+					transition:slide={animationConfig}
```
endigo9740: Interesting use of Svelte transitions. Just keep in mind we can't replicate these for React or other frameworks. We have this same dilemma for other components (ex: Accordions) so no change required on your end. Just wanted to bring this to your attention ahead of time.

packages/skeleton-svelte/src/components/TreeView/TreeView.svelte
```diff
@@ -0,0 +1,185 @@
...
+					{...api.getBranchContentProps(nodeProps)}
+					data-testid="tree-content"
+				>
+					<div {...api.getBranchIndentGuideProps(nodeProps)} style="--indent-factor:{indentAmount}"></div>
+					<div class="flex flex-col">
```
endigo9740: Even if it's structural, avoid fixed classes in the template. Go ahead and abstract these to style props. These should be `base` type since they are for the core layout.

packages/skeleton-svelte/src/components/TreeView/TreeView.svelte
```diff
@@ -0,0 +1,185 @@
...
+							{#each nodeProps.node.children as childNode, index}
+								{@render treeNode({ node: childNode, indexPath: [...nodeProps.indexPath, index] })}
+							{/each}
+						{/if}
+					</div>
+				</div>
+			{/if}
+		</div>
+	{:else}
+		<button
```
endigo9740: By default buttons use `type="submit"`. Make sure Zag is setting a `type="button"` to prevent accidental form submission. Should be as simple as inspecting the button in your browser tools and reviewing the attributes applied.

- If they are -> no further action
- If they are not -> let me know and we'll report a bug

packages/skeleton-svelte/src/components/TreeView/TreeView.test.ts
```diff
@@ -0,0 +1,99 @@
+import { describe, expect, it } from 'vitest';
```
endigo9740, line 1: Just a big thumbs up for including the test cases this early. Most folks skip these outright. This is very much appreciated!

I'll save any comments for these until my final review when the code is solidified.

playgrounds/skeleton-svelte/src/routes/components/tree-view/+page.svelte
```diff
@@ -0,0 +1,57 @@
+<script lang="ts">
```
endigo9740, line 1: Again kudos on setting up the Playground iteration. It's of course aimed to help developers like yourself test the code you're writing, but sometimes folks miss it.

endigo9740 (changes requested):
@dev-viinz FYI I'm just starting my PR review, first off I want to so great job on this. This is one of the more complex components so you've dived right into the deep end of the pool. Kudos to that!

I'm going to begin my review by asking some high level questions as you've had more hands on with the code for this component than I have. I'll also highlight individual changes in the code itself, so take a peak at those as well when you can.

## Question 1

Do you think there are ways we can split this one monolithic component into smaller sub-components? 

There are typically notable benefits to this:

1. It can be a lot easier for users to parse how components work when the structure is piecemeal.
2. This is the exact approach we're learning into with our future [component API changes](https://github.com/skeletonlabs/skeleton/discussions/3499). In fact it's required as we're retiring prefix style props in place of localized props. Each "part" within the component has it's own component and all utility classes, attributes, etc are passed verbatim to this.
3. Svelte operates fine with "monolithic" components, but React actually works a lot better with "piecemeal" components. Since we support both frameworks, we tend to favor what works in both scenarios. In other words we've lean towards the React pattern.

The only downside I could see offhand would be Branch Control/Content sections. If any state has to go upwards in the tree to inform the parent/wrapping component that can be a challenge. Though maybe Zag handles that for us here.

## Question 2

Do you think this component would benefit from adopting our new props Component APIs over the current APIs? I wanted to give you a chance to construct this component using our current production conventions that way you could accurately judge the difference. Keep in mind we will need to port this forward at some point anyways. And we're already aiming to test a [limited set of components](https://github.com/skeletonlabs/skeleton/pull/3491) with the new conventions in the near future. Not a hard requirement - just asking if you think this would make this easier to work with and better future-proof things.

## Question 3

Have you tested `multiple` mode? I know we encountered some issues with the Combobox needing template changes to support a similar feature. Just wanted to confirm if that's the case here:

- https://zagjs.com/components/svelte/tree-view#multiple-selection

## Question 4

Generally speaking, how do you feel about the way Zag handles state management for this component? There's no hierarchy to the state, it's just a flat list of IDs for open branches. I'm assuming you're contributing this as you have need for it for yourself. Would it satisfy the requirements you need it for?

dev-viinz:
First of all, thanks for all the kind words :)
To answer the questions to the best of my abilities:
# Answer to Q1:
I think so yes.
To clarify, I understand the template-driven approach to require the user to recursively add the different nodes himself correct?
If so, my biggest concern lies with making Zag cooperate with a non data-driven approach. As far as I can tell, they don't really support any other way apart from providing the entire collection at the service/api creation point. 
But I can see how we could leverage reactivity to make it work, no idea if it would be viable in terms of performance though.
I reckon I could give it a shot and we'll see from there :)
# Answer to Q2:
To me this kind of goes hand-in-hand with the previous one. If we go down the multi-component path anyways, I believe it makes a lot of sense to just go all in on the newer approaches, including the props Component APIs.
Wouldn't this require #3491 to be merged, for me to access the required types?
# Answer to Q3:
I have not, but seeing I'll be refactoring the component as mentioned above, I'll test that then. Will probably also add automated tests accordingly.
# Answer to Q4:
I don't mind it, and for the current use case at work this wouldn't be an issue. I did realize that I would need checkboxes at the lead position, so the template-driven approach might be saves this for me.
# Recap
All in all, I like the idea of splitting the component and using the new props Component APIs. So I'll get started on that.

pnpm-workspace.yaml
dev-viinz: Unfortunately not. The change is included as of version 1.18.1. Though I wouldn't see why we couldn't update to the currently latest version (1.18.2). If you want, I could keep Zag up to date in this branch, so that it's not outdated by the time this merges 🤔

packages/skeleton-react/src/components/Segment/Segment.test.tsx
dev-viinz: It is indeed mocked in `packages/skeleton-svelte/vitest.setup.js`, but not in `packages/skeleton-react/vitest.setup.js`, and the test in question is in the react package.
I'll move the mock to the react vitest config to mirror the svelte setup.

packages/skeleton-svelte/src/components/TreeView/TreeView.svelte
dev-viinz: Honestly this was only due to the Rating.svelte component doing it this way. I don't have the most hands on experience with svelte 5, as we're stuck with svelte 4 at work, so I thought this must be for performance reasons, as the module script tags only get run once afaik. I personally don't mind either approach, do you want me to make it a HTML snippet?

packages/skeleton-svelte/src/components/TreeView/TreeView.svelte
dev-viinz: I'll keep this in mind, but if we switch to the props Component APIs this becomes a non issue anyways right?

packages/skeleton-svelte/src/components/TreeView/TreeView.svelte
dev-viinz: Noted. Same line of thinking though: with the switch to the "new" way to do props, this won't exists in this way anymore.

packages/skeleton-svelte/src/components/TreeView/TreeView.svelte
dev-viinz: Noted. Same line of thinking though: with the switch to the "new" way to do props, this won't exists in this way anymore.

packages/skeleton-svelte/src/components/TreeView/TreeView.svelte
dev-viinz: When splitting into multiple components this mostly becomes obsolete, but I'll keep it in mind :)

packages/skeleton-svelte/src/components/TreeView/TreeView.svelte
dev-viinz: They seem to have slipped though, I'll get them out of there with the splitting into multiple components.

packages/skeleton-svelte/src/components/TreeView/TreeView.svelte
dev-viinz: Makes total sense, and I wouldn't have thought of that :D
Unfortunately, Zag doesn't seem to set this attribute.

packages/skeleton-svelte/src/components/TreeView/TreeView.svelte
dev-viinz: I had a go at a solution, was it something like this you had in mind? [the commit in question](https://github.com/skeletonlabs/skeleton/pull/3585/commits/02e116ea2faba6faa5e8abf3b0a891b12979dc89)

pnpm-workspace.yaml
endigo9740: Go ahead and give 1.18.2 a try. if there's no breaking changes then go for it. Otherwise let me know and I'll make this a separate task as part of that the Circular Progress bar update planned.

packages/skeleton-svelte/src/components/TreeView/TreeView.svelte
endigo9740: Let's do an HTML snippet then! Afaik performance should not be impacted.

packages/skeleton-svelte/src/components/TreeView/TreeView.svelte
endigo9740: Correct, I've reviewed everything with the current production conventions in mind.

packages/skeleton-svelte/src/components/TreeView/TreeView.svelte
endigo9740: Whatever we need to do to get rid of the `<style>` tag. As mentioned it won't carry over to the React version.

packages/skeleton-svelte/src/components/TreeView/TreeView.svelte
endigo9740: Ok, let's make sure insert the attribute ourselves for now. Then we'll need to inform Zag of the issue. It should be rare that a Tree View is inserted inside a form, but folks do weird stuff all the time :D

endigo9740:
> Answer to Q1:

Generally speaking Zag components are minimum examples and always in a monolithic format. They leave it up to the implementing party to determine how they wish to consume and divide. In our case I think we can make use of the Context API for Svelte/React. You can reference examples from other components here:

- https://github.com/skeletonlabs/skeleton/blob/main/packages/skeleton-svelte/src/components/Segment/context.ts
- https://github.com/skeletonlabs/skeleton/blob/main/packages/skeleton-svelte/src/components/Tabs/context.ts

> Answer to Q2:

Not to flip flop here but let's stay on course with the current production conventions. Honestly I'm just overly excited about the changes and wanting them all now. I do believe there are some utility features in #3491 that would be required to proceed with this. Let's not make that a blocking issue to get this out the door asap. Users would rather have a working component sooner than later I'm sure.

> Answer to Q3:

Cool, thanks.

> Answer to Q4:

Sound good and I agree.

> All in all, I like the idea of splitting the component and using the new props Component APIs. So I'll get started on that.

Yeah I'd very much like to see what that looks like. I think will also give us a leg up on the next iteration with the new approach to our component APIs

pnpm-workspace.yaml
dev-viinz: Updated, doesn't seem to break (the changelog doesn't contain anything that would suggest a breaking change either) 😄

endigo9740:
@dev-viinz things are just finally getting back to normal for me. If you don't mind, feel free to give me a status update when you can. If this is ready for another proper review, tap the "ready for review" option on here and I'll set aside time asap.

endigo9740:
@dev-viinz just saw you flagged this "ready for review". Note I'm just in the process of stepping away for the day and weekend, but I'll aim to do a pass through asap next week. On Monday if time allows. Feel free to give me a run down if there's anything I need to know ahead of time. Thanks sir!

dev-viinz:
@endigo9740 No worries, I would have flagged it earlier, but I didn't see the message.
There aren't really any big things to look out for, though I would recommend to start by looking at the usage in the playground first.
It might make the component structure easier to follow.

You will notice, that apart from the TreeView, TreeItem, and TreeBranch, there is a TreeNode component. This one is internal and doesn't get exposed to the user. Its purpose is to not repeat some code between items and branches, but I'm unsure whether it's more confusing than just having the logic in both the item and the branch, though it would result in quite a few duplicate lines.

Other than that, it's just very heavy on the context usage, but that's by design, as I had to satisfy Zag's requirement of having the entire data structure at once. So while it works, it _might_ be a bit performance hungry.
I'm not experienced in benchmarking, so I didn't do any for now. Maybe you can (out of experience) gauge how performance could be affected and how we could benchmark or straight up improve performance.
But again, I haven't noticed any problems.

PS: If possible, it's probably better to (additionally) ping me on discord, as I'll likely react **much** quicker :)

packages/skeleton-svelte/src/components/TreeView/context.ts
```diff
@@ -0,0 +1,12 @@
+// import * as tree from '@zag-js/tree-view';
+import { createContext } from '../../internal/create-context.js';
+import type { NodeContext, TreeViewContext } from './types.js';
+
+// const throwOnDefault = (): never => {
```
endigo9740: Go ahead and remove unused comment blocks like this.

packages/skeleton-svelte/src/components/TreeView/TreeView.svelte
```diff
@@ -0,0 +1,208 @@
...
+		updateNode,
+		// Base
+		background,
+		spaceY,
+		border,
+		padding,
+		shadow,
+		classes,
+		// Control
+		controlBase,
```
endigo9740: You've done a great job of distributing the logic to sub-components, but style props need the same treatment. Styles props need to be relative to the component, not passed down through context. I'll take the hit on this one as I don't think this was ever noted in the contribution docs - but is reflected in other existing components.

I know the goal is try and save folks having to repeat styles, but this actually ends up adding a lot of complexity where everything flows in from the parent down to the children. It should result in less complex Tailwind classes like this:

```
[&[data-selected]]:preset-tonal-primary /* Before */
[data-selected]:preset-tonal-primary */ After */
```

Another notable benefit is rather than having prefixes like `itemBorder` applied on the `<Treeview.Item>` component, it can follow the root prop convention and drop the prefix:

```
<TreeView.Item border="...">
```

This leads to a lot less prefixes for end users to keep up with, and puts us one step closer to the v4 conventions we're leading into in the future.

packages/skeleton-svelte/src/components/TreeView/TreeView.svelte
```diff
@@ -0,0 +1,208 @@
...
+
+	setTreeContext(treeContext);
+</script>
+
+<!-- @component A collapsible TreeView. -->
+
+<!-- Tree -->
+<div class="{base} {background} {spaceY} {border} {padding} {shadow} {classes}" {...api?.getRootProps()} data-testid="tree">
+	{#if label}
+		<h3 {...api?.getLabelProps()}>
```
endigo9740: I understand Zag's example showcases a H3 tag here, but I feel like this might be making some assumptions that could lead to invalid a11y - that being, what if there's not an H2 proceeding this?

There's also the concerns that if a label snippet is enabled, there's no props for providing the appropriate Skeleton typography class (ex: `.h3`).

So I'm going to propose two changes:

1. Implement a new prop to set a `<svelte:element>` - the default value can be h3
1. Implement the appropriate prefixed style props for `labelBase/labelClasses`, as well as a unique `labelTypography` prop that takes a Skeleton typography class - the default can be `h3`.

Let me know if you have any questions around this or the future React equivalent.

packages/skeleton-svelte/src/components/TreeView/types.ts
```diff
@@ -0,0 +1,173 @@
...
+
+export interface NodeSnippets {
+	control?: Snippet;
+	content?: Snippet;
+	item?: Snippet;
+}
+
+export type SnippetTypes = 'item' | 'control' | 'content';
+
+export interface TreeViewContext extends TreeViewProps {
```
endigo9740: See the `types.ts` for Tabs to see how we denote and categorize Context types versus Component prop types. Basically:

- Context comes first
- Components comes second
- Then we use comment blocks to visually separate them

packages/skeleton-svelte/src/components/TreeView/TreeBranch.svelte
```diff
@@ -0,0 +1,61 @@
...
+			<!-- Control -->
+			<button
+				class="{treeContext.controlBase} {treeContext.controlBackground} {treeContext.controlSpaceY} {treeContext.controlHover} {treeContext.controlBorder} {treeContext.controlPadding} {treeContext.controlShadow} {treeContext.controlClasses}"
+				{...treeContext.api?.getBranchControlProps(nodeProps)}
+				data-testid="tree-control"
+				type="button"
+			>
+				<!-- Indicator -->
+				<span
+					class="flex items-center {treeContext.indicatorTransition} {treeContext.indicatorRotationClass}"
```
endigo9740: Make sure no hardcoded classes are left orphaned in the component templates. Default (read: structural) styles should be part of the `base` styles per each component element.

https://www.skeleton.dev/docs/resources/contribute/components#style-prop-conventions

Rule of thumb: HTML templates should contain zero Tailwind classes. They should all be distributed to customizable props.

packages/skeleton-svelte/src/components/TreeView/TreeNode.svelte
```diff
@@ -0,0 +1,103 @@
+<script lang="ts">
```
endigo9740, line 1: I'm a little confused at the addition of TreeNode and how much additional logic is in place compared to the original Zag examples - all this registering/unregistering logic. It feels like we've gone way off course here. The additional of Context APIs should not have resulted this level of extra complexity. In most cases it's just a matter of passing the appropriate state up/down the tree, and letting Zag do most of the heavy lifting.

You have a better handle of the component than I do though, so I'm not saying it's not required. I just need to understand the justification for it. As this is going to significantly impact our ability to maintain this feature.

One of our goals with Skeleton components is to keep as close to the Zag source as we reasonably can, so not to introduce regressions or break a11y. Even if that means we avoid introducing new features.

packages/skeleton/src/base/theme.scss
```diff
@@ -43,6 +43,20 @@
 	--anchor-text-decoration-active: inherit;
 	--anchor-text-decoration-focus: inherit;
 
+	/* Indentation --- */
+
+	--indent-width-none: 0rem;
```
endigo9740: Unfortunately this is probably a no-go. We set the expectation that components will be isolated and not require extra CSS from the core library to function. We don't expect our components to be truly headless, but this sort of cross-requirements typically leads to bad practices. Can you help explain the goal here and maybe I can suggest an alternative?

I'm also not immediately clear what `rotation` is referring to - did you mean to use `rotate`? Don't forget Tailwind allows for the arbitrary syntax, so `rotate-[90]` is perfectly valid if `rotate-90` is not a canned option by default. And I believe both 90/180 are canned options.

packages/skeleton/src/utilities/tree-view.css
```diff
@@ -0,0 +1,25 @@
+/* Component: TreeView */
+
+@utility indent-guide {
```
endigo9740: Again, we don't allow custom CSS generated in the core package for use in components like this. Styles should be isolated to the components themselves. Let's work together to determine another way to solve for this.

.changeset/neat-bushes-thank.md
```diff
@@ -0,0 +1,8 @@
+---
+"@skeletonlabs/skeleton-svelte": minor
+"@skeletonlabs/skeleton-react": minor
+"@skeletonlabs/skeleton": patch
```
endigo9740: Make sure to remove any mention to modification of the core library per my other review comments.

endigo9740 (changes requested):
Alright @dev-viinz so there's going to be quite a lot of feedback this round. Remember this one of the most complex components, so it's important we get things right if we wish to be able to maintain this long term.

You've taken several steps in the right direction, but I have some notable concerns about the increase in complexity overall. Plus there's a few style prop conventions that aren't quite right yet.

Feel free to address this in multiple passes if need be. But let's make sure we're implementing this as succinctly as possible.

A couple other loose notes that didn't fit in elsewhere...

1. There's an alignment issue with the icons, highlighted in red here:
treeview
2. Also make sure you've caught up to the latest updates on `main` branch. There's been quite a few notable updates - including a major dependency bump. Defer to our handling of most dependencies, but of course use the latest Zag version between either side.

packages/skeleton-svelte/src/components/TreeView/context.ts
dev-viinz: hope i haven't missed any :)

packages/skeleton-svelte/src/components/TreeView/TreeView.svelte
dev-viinz: Done, though i believe we can't simplify the tailwind class, i made a change, so it's a lot more user friendly, lmk what you think :)

dev-viinz: as you mentioned, the solution won't hold up in production, see my proposal on discord

packages/skeleton/src/base/theme.scss
dev-viinz: Honestly I don't remember what made me do this, looking at your first review you advised explicitly against this... Must have been a late night session.

I'll switch these to just use tailwind classes directly, no real reason not to.

dev-viinz: Done, should be a lot cleaner now.

packages/skeleton/src/utilities/tree-view.css
dev-viinz: Done, should be a lot cleaner now.

.changeset/neat-bushes-thank.md
dev-viinz: Done, should be a lot cleaner now.

packages/skeleton-svelte/src/components/TreeView/TreeView.svelte
dev-viinz: I added my proposed change for now, just to try it out. I think it's pretty good, but I hope there is comparable syntax in React 👀

packages/skeleton-svelte/src/components/TreeView/TreeBranch.svelte
dev-viinz: Done :)

packages/skeleton-svelte/src/components/TreeView/TreeNode.svelte
dev-viinz: As discussed on Discord I'll wait on your verdict, but please tell me if there is anything I can do ^^

packages/skeleton-svelte/src/components/TreeView/types.ts
dev-viinz: I cleaned the file up, though I still have a small Zag section at the bottom.

packages/skeleton-svelte/src/components/TreeView/TreeView.svelte
dev-viinz: Also, I moved all the style props from the context to actual props on their respective components. (forgot to mention it)

dev-viinz:
> A couple other loose notes that didn't fit in elsewhere...

Fixed the alignment issue, and we're up to date :)
@endigo9740

packages/skeleton-svelte/src/components/TreeView/TreeView.svelte
dev-viinz: I took inspiration from the Combobox and adjusted it accordingly, should be as you expected (and easily portable, as I've opted for option 2).

packages/skeleton-svelte/src/components/TreeView/TreeView.svelte
endigo9740: This is better, but we have a pattern established for this that keeps the logic out of the template:
https://www.skeleton.dev/docs/resources/contribute/components#dynamic-style-props

In this case you would do something like this:

```ts
const rxSelected = $derived(treeContext.api?.selectedValue.includes(id) && selected);
```
```html
class="... {rxSelected}"
```

Looks like our documentation was updated to remove the `rx` (aka "reactive") naming convention, which indidcates the classes "swap out" - but we actively use this pattern in our components today:

- Svelte: https://github.com/skeletonlabs/skeleton/blob/main/packages/skeleton-svelte/src/components/FileUpload/FileUpload.svelte#L67
- React: https://github.com/skeletonlabs/skeleton/blob/main/packages/skeleton-react/src/components/FileUpload/FileUpload.tsx#L65

Again, sorry about the gap in the docs. In this case it previously showed this properly, then someone modified it and I didn't catch it. That's my bad.

packages/skeleton-svelte/src/components/TreeView/TreeView.svelte
endigo9740: First off, I apologize for not being more clear. I intended for you to implement BOTH changes, not an either/or situation. We need to have the semantic HTML element to ensure support for screen readers. While the `h3` class is purely cosmetic.

This is also another place we've gone "off script" and modified the element structure Zag proposed. I mentioned before, but we really want to stick as close to Zag as possible. In this case there's no wrapping `<label>`, JUST the heading:

```html
<div {...api.getRootProps()}>
  <h3 {...api.getLabelProps()}>My Documents</h3>
  <div {...api.getTreeProps()}>
    {#each collection.rootNode.children || [] as node, index}
      {@render treeNode({ node, indexPath: [index], api })}
    {/each}
  </div>
</div>
```

Let's go ahead and strip out the label and follow their lead. But ensure the H3 element/class is dynamic via props. We're just offering the smallest of extra features here to help a11y - but this is the exception, not the norm.

packages/skeleton-svelte/src/components/TreeView/TreeView.svelte
endigo9740: Note this is also tripping up the linter, so another reason to modify it:

Screenshot 2025-08-07 at 9 59 30 AM

Make sure you're running the recommended test commands in the original PR template periodically. The `pnpm lint` command should catch this.

packages/skeleton-svelte/src/components/TreeView/TreeNode.svelte
endigo9740: Ok I see what's going on here. I missed that `treeNode` was a Snippet in the original Zag example. Zag likely used this as it's the closest Svelte convention to React's convention for adding multiple components in a single file.

That said, I'm not confident in the implementation of `<TreeNode>` - it seems you've reimplemented a lot of the logic in script form rather than keeping the snippet format. So I think we have three options here:

## Option 1

What I'd expect instead would be nearly verbatim to what they have in their example, but a single component called `<TreeNode>` that implements a single snippet `treeNode` (thought the name could be changed to avoid the awkward redundancy), and that's implemented EXACTLY like Zag's example. So the component is merely a wrapper around the shared snippet.

## Option 2

Alternatively. rather than the extra overhead of a component - snippets can be implemented and exported top-level.
https://svelte.dev/docs/svelte/snippet#Exporting-snippets

But with this caveat, meaning data from the script tag (ex: `api`) would have to be passed in:
> provided they don’t reference any declarations in a non-module <script>

## Option 3

Finally, there is a way to programmatically create Snippets in Svelte - but I find the syntax very difficult to parse and it would require a notable rewrite. Let's not go down this path.

## My Recommendation

I advise we opt for option 1. It'll keep most of the code close to the original source, while allowing us to do away with all this custom update/register logic. The result would be something like this:

> NOTE: this is untested pseudo code - but should cover the concept at least...

```html
<script lang="ts">
  let { treeNode } = $props();
  const treeContext = getTreeContext();
</script>

{#snippet treeNode(nodeProps: TreeNodeProps)}
  {@const { node, indexPath, api } = nodeProps}
  {@const nodeState = api.getNodeState({ indexPath, node })}

  {#if nodeState.isBranch}
    <div {...treeContext.api?.getBranchProps({ indexPath, node })}>
      <div {...treeContext.api?.getBranchControlProps({ indexPath, node })}>
        <FolderIcon />
        <span {...treeContext.api?.getBranchTextProps({ indexPath, node })}>{node.name}</span>
        <span {...treeContext.api?.getBranchIndicatorProps({ indexPath, node })}>
          <ChevronRightIcon />
        </span>
      </div>
      <div {...treeContext.api?.getBranchContentProps({ indexPath, node })}>
        <div {...treeContext.api?.getBranchIndentGuideProps({ indexPath, node })}></div>
        {#each node.children || [] as childNode, index}
          {@render treeNode({ node: childNode, indexPath: [...indexPath, index], api })}
        {/each}
      </div>
    </div>
  {:else}
    <div {...treeContext.api?.getItemProps({ indexPath, node })}>
      <FileIcon />
      {node.name}
    </div>
  {/if}
{/snippet}
```

Let me know if you have any questions or concerns. Or just generally any struggles implementing this.

endigo9740:
@dev-viinz I've complete another review. Note my comments above - as well as my suggestion to run each PNPM command found in the original post above. There's at least 1-2 issues that can be caught and resolved due to these checks.

I think where you're landed is fairly close to what we need, just with the notable shake up to the `TreeNode` component. Note that when we implement the React version it'll be a LOT closer to the original source given components are defined in a single file. Zag is developed by the Chakra UI devs (a React component lib) so they tend to be a bit more knowledgeable about React conventions than Svelte. And in this case this is a actually a really hard problem to solve since the snippet is used recursively.

Once the suggestions above are implemented I'd like to do one more "nitpick" pass before we move onto React. Expect this to be a small collection of asks - like adjusting icons sizes, etc.

packages/skeleton-svelte/src/components/TreeView/TreeView.svelte
dev-viinz: Ah dang, also kinda on me, wasn't that ambiguous :D
Should be good now ^^

packages/skeleton-svelte/src/components/TreeView/TreeView.svelte
dev-viinz: Yeah I've seen that pattern, but apparently didn't think of applying it...
Should be good now :)

packages/skeleton-svelte/src/components/TreeView/TreeNode.svelte
dev-viinz: I'm a bit confused now...
If I understand this correctly, you're wanting to go back to an approach where the user specifies the tree through data.
My current implementation doesn't expect the user to do that. The user simply provides the markup, and the data needed for Zag gets built behind the scenes.

That's also the reason for the added complexity.
I think before I do anything, I need to understand what "interface" we want to provide to the user.
But I was under the impression, that the last time we spoke about this, you asked me to make this "template-driven", which is what this is, and I don't see a much easier solution.
If we are going back to "data-driven", you're totally right to want to simplify the entire thing.

Again, just slightly confusing as that's pretty much what I did in my first iteration.

FIY: I'm out until Sunday, won't be able to respond until then :)

endigo9740: @dev-viinz yeah some of the downsides to async communication. In this case, converting TreeNode to a component that wraps the snippet would still be Template-Driven IMO. Here's a brief comparison if it helps:

Data Driven:

```html
<TreeView data={someData} />
```

Template Driven:

```html
<TreeView>
  <TreeBranch>
    <TreeItem>...</TreeItem>
    <TreeItem>...</TreeItem>
    <TreeItem>...</TreeItem>
  </TreeBranch>
  <TreeBranch>
    <TreeItem>...</TreeItem>
    <TreeItem>...</TreeItem>
    <TreeItem>...</TreeItem>
  </TreeBranch>
</TreeView>
```

Basically the shape is defined via a markup-esque manner (like HTML) rather than passing all the data to a component and it constructs everything within. Hopefully that gets us to the same semantics/vernacular here.

To clarify our goal with all components:

1. We want to follow the Zag source as closely as possible
2. However, Zag goes out of their way to demonstrate every component using a single code block. This unfortunately doesn't scale to real world usage. This is why we need to break the component up into smaller chunks (into sub-components).
3. That said, we should not try to get clever and go way off script from how they implement. To put that literally, we want to keep the code syntax as close theirs as possible. Even if the source is broke into smaller parts/files and implementing shared context values.
4. But we should never rewrite portions of the source to facilitate any sort of extra functionality, no matter how creative or clever we think it might be.

Examples of where we're good to go away from their source:

1. Implementing `<svelte:element>` for the heading
2. Implementing the Context API when sub-dividing to multiple components

But again, we should do everything in our power to avoid diverging from their source and structure for the elements. That's getting way too clever, and we don't want to be clever. We want to be boring and predictable. This will lead us down a path where Zag does the heavy lifting, and just follow their lead. But sprinkle tiny changes (ex:  heading) when it won't rock the boat.

If users need or want more complex features or updates - then it's a request that needs to go upstream to Zag themselves. The onus for that is not on us.

> FIY: I'm out until Sunday, won't be able to respond until then :)

Totally fine!

endigo9740: Also so it doesn't come across as me saying "everything has to be this way" without context. I was originally of the mindset and did exactly what you did. Got creative and modified large portions of the Zag code to fit our needs. It took @Hugos68 to come in and bop me on the head and explain the practicality and pros/cons of that approach for me to realize why it's really necessary. Basically it comes down to this: let Zag do the heavy lifting, that way we can focus on what's unique to Skeleton (the Design System aspects)

dev-viinz: > Basically the shape is defined via a markup-esque manner (like HTML) rather than passing all the data to a component and it constructs everything within. Hopefully that gets us to the same semantics/vernacular here.

Good thing is: we are on the same page about the semantics of the two approaches.
The issue with the template driven markup is that it doesn't bode well with how Zag works. At least not without _**also**_ providing the data (the same way you would in the data driven approach).

Zag **needs** the collection data, in order to do all the heavy lifting. My funny logic with registering and unregistering and updating nodes is just that. It's updating the collection that's fed to Zag, so that it can properly set all the props etc.

It might just be me, but I don't see how to get this data into Zag without the logic.
But I would argue, that what I built isn't interfering with Zag or modifying it. It's just what's needed to make Zag go to work. After initialization my code only ever runs when the User changes the markup on the fly, which is probably negligeable, as you don't really need to do that, as you have a functioning TreeView via Zag.

What I mean to say is, the additional logic is what's making the Zag TreeView easily usable without having to think about the data that needs to be provided to Zag.

Hope this paragraph even makes sense 🫣

> Also so it doesn't come across as me saying...

Oh no worries, I'd rather have this conversation than to deliver a sub optimal end result ;)

endigo9740: @dev-viinz all good, in that case go ahead and commit any changes you made tonight and I'll go hands on tomorrow and test my idea. If it works I'll commit the change, if it doesn't I'll drop and revert and keep what you've got now. Either way we should land in a good place come early next week and begin the React version.

And again, great work on all this. I believe I said it before, but TreeView is one of (if not the) most complicated components out there. So props for building it, I'm sure folks are going to love it!

endigo9740: Note this conversation is getting buried in the fold between all the commits, so I'm going to start a new message below.

endigo9740:
Conversation continued from here: https://github.com/skeletonlabs/skeleton/pull/3585#discussion_r2263235445

@dev-viinz Ok, so I've had a chance to go hands on with this today and unfortunately I don't have great news...

After reviewing how Zag is implementing the TreeView in detail I've determined it's not well suited to a template-driven approach where you define the structure using multiple sub-components. The primary culprit for this is the way the `#each` block is used here:

```html
{#each node.children || [] as childNode, index}
    {@render treeNode({ node: childNode, indexPath: [...indexPath, index], api })}
{/each}
```

Each item to know both it's node and indexed position within the DOM tree. For example:

- 0: first node element
- 1: second node element
- 2: third node element

We ran into a similar issue with the Skeleton v2 Stepper component. And while it's not impossible to deal with it requires some really annoying work arounds.
- In your case you had to break from the source and implement all the update/register/unregister logic.
- In the last version of our v2 Stepper, we used a Context API to essentially self-register each node (and thus index) with the parent. But this pattern immediately breaks down when the DOM structure of the component change in an unpredictable manner (ex: using `#if` blocks, etc)

This is more than likely a case where the pattern works well in React (which the Chakra team are well versed in), but not so much in other frameworks like Svelte. JSX plays by a different set of rules since so much can be inlined and multiple components registered in the same file.

Your solution works, but it's broken our rules about the keeping Zag's source structure in place - which leads to extra maintenance overhead. As much as it pains me to say it, the most practical solution I can think of would be to revert back to the original monolithic single component solution. The upside is the structure and data flow becomes much simpler, while the downside is users lose access to the UX of multiple components and the isolated scope for customization. This isn't going to play well with our proposed v4 changes either.

On a final note, I'm not sure if you're aware of [Ark UI](https://ark-ui.com/docs/). Their the first-party counterpart to Zag - where they actually implement the Zag primitives into a functional component library. They DO [provide a foundation](https://ark-ui.com/docs/components/tree-view) that's a lot close to our proposed v4 solution (with many sub-components), but with the trade off of a requiring a HUGE number of components and what looks to be a ton of complexity. See the source here:

https://github.com/chakra-ui/ark/tree/main/packages/svelte/src/lib/components/tree-view

It might be worth researching what their doing, as they seem to be solving some of these issues. But I'm not going to lie, their implementation is a beast. There's SO much going on here I'm not sure that's a reasonable ask. Plus I am also a bit leery of the user-facing DX for juggling so many small parts.

What I am going to to do is suggest you push pause until you hear back from me. Then I'm going to brainstorm with the team and see if they have any other suggestions. We can resume once we have an battle plan in place - which may or may not include moving forward with your current implementation. Please be prepared for that.

endigo9740:
@dev-viinz quick follow up - I talked a bit with @Hugos68 on this. He doesn't have enough working knowledge of the TreeView source code to provide an informed response yet, but I think between us we do believe there may be a way to solve this. Especially since Ark is doing this - just with a ton of extra overhead and complexity :D

I may try to carve out some time next week to run through and better understand how they're accomplishing this. One quick item of note though - they're using a script module to share *some* state. Not sure what exactly yet. But this may be worth looking into.

Either way we can pick up on this next week. If you have any other ideas I'm all ears.

dev-viinz:
Hi @endigo9740 ,
I wasnt able to have a look yet, but I could analyze what Ark does, and build something adjacent to it. 

Bummer that my implementation isnt going to work, because I really liked the DX 🤷🏼‍♂️

But I totally understand. Again I'll have a look at ark, so you might Not Need to dive in too deep for now :)

endigo9740:
@dev-viinz My day-to-day time to work on Skeleton is pretty minimal right now, so any help digging into the Ark version would be much appreciated. I don't think we need to go to the extreme they did for the number of sub-components. But I do think they have a functional strategy for dealing with the DOM tree index issue I brought up above. If you can determine how that's working, maybe we can meet them half way. Thanks for the assist!

dev-viinz:
@endigo9740 my first impression is that ark is pretty much just doing a hybrid approach between the two I had proposed.
makes sense as they probably want to keep the experience of using ark close to zag itself.
so borrowing from that I would keep a lot of what I have, and remove all the state juggling, and let the user provide the collection as well as building the tree with our components. we'll be "close" to what ark is doing, just with less granular components, and their whole "ark" structure they have in place

Edit: I'll probably start tomorrow, but progress might be slower as I'm not home until Monday :)

endigo9740:
@dev-viinz that all sounds great. And just a heads up we have formally announced the new v4 component changes here:

- https://github.com/skeletonlabs/skeleton/discussions/3634

Feel free to give this a test and maybe consider again if the changes would be valuable enough to move straight towards this sooner than later. Especially if any notable refactoring is needed. It might be best to just wait and build this version rather than building it now, then we have to turn around and rebuild in the near future.

endigo9740:
Hey @dev-viinz, just a heads up - I've been speaking with @Hugos68 this morning and we're keen to start building towards our "next" release (aka v4). As such I'm going to make the executive decision here to close out this PR and put this version of the TreeView to rest for now.

That said, we would love to get you involved with building the v4 equivalent. I think there's a lot that can be referenced and learned this PR iteration. And I think you would be invaluable as a contributor (or at least consultant) on the next version.

We're currently working through some administrative tasks, such as prepping our branch and release structure as we split between v3/next again. We don't quite have contributor docs ready for the new component structure. That said, we have two brand new components that can [begin to review and test right away](https://www.skeleton.dev/docs/get-started/v4-early-preview). As well as reach out to Hugo and I to ask questions on Discord about how things are setup and why.

I know you expressed a desire to work on other components as well, so if you would like to maybe take on a simpler component first, then work your way up. That would be totally fine by me. Just let me know here or on Discord please!

And again, thanks for all the hard work on this PR!

dev-viinz:
Sounds good, the faster we can get v4 out the better imo. To me v3 feels a bit like a "short term" release :)
Lmk what I should get started on first ^^

endigo9740:
@dev-viinz absolutely - stay in touch with Hugo and I in the Discord `#contributor` channel. If you don't hear anything by end of week ping me and I'll give you a status update. But we'll be aiming to get everything prepared asap. In the meantime definitely give this new components a review and let us know if you have any questions about the implementation. That can help us flush out the contribution guidelines as we go.

PR #3612, opened 2025-07-15
dep bump

Hugos68:
updates all dependencies (including a major one: vite 7)


endigo9740:
@Hugos68 awesome! Looks like "code quality" is failing in the CI checks though. I'll devote some time to helping do some regression testing on this tomorrow!

pnpm-workspace.yaml
```diff
@@ -1,29 +1,29 @@
...
-  - 'packages/*'
-  - 'sites/*'
-  - 'playgrounds/*'
-
-define: &zag_version '^1.7.0'
+  - packages/*
+  - sites/*
+  - playgrounds/*
 
 catalog:
```
endigo9740, line 8: Just curious about the change from a shared/common Zag version to all of these being maintained individually. It's my understand these work in conjunction with the core Svelte/React/etc package from Zag so they must remain in sync, right?

playgrounds/skeleton-react/package.json
```diff
@@ -10,17 +10,17 @@
...
 	"devDependencies": {
-		"@types/node": "^22.13.14",
-		"@types/react": "^19.0.12",
-		"@types/react-dom": "^19.0.4",
+		"@types/node": "^24.0.14",
+		"@types/react": "^19.1.8",
+		"@types/react-dom": "^19.1.6",
 		"tailwindcss": "catalog:",
 		"typescript": "catalog:"
 	}
```
endigo9740, line 26: Just using this file to tag a comment...

But just noticed the Playground link to Progress Rings is failing because the `layout.tsx` uses `ringS` with an `s`, but the route is `ring`   without. Let's go ahead and resolve this (and double check for the Svelte playground too)

Screenshot 2025-07-16 at 10 17 10 AM

endigo9740 (changes requested):
See the Vitest setup config fix for the `<Segment>` component here, the win.Resizer` addition:
https://github.com/skeletonlabs/skeleton/blob/4c62218aea0f9a63f55472f13c48d5010d9602fb/packages/skeleton-react/vitest.setup.js#L10

Seems Zag updates something for this component in the newer version that's tripping up our test cases.

Currently this is tripping up our "Code Quality" CI check. It's an issue for Svelte, but have not had a chance to test for React - but should likely be the same.

pnpm-workspace.yaml
Hugos68: What do you mean individually? This zag_version variable sets every package and machine we consume to the specified version

Hugos68: Oh WTH I see what you mean now, I didn't do this, PNPM changed this to be all individual, good catch. I will revert this. I have no clue why PNPM did this.

endigo9740: Weird! Ok yeah let's revert that then!

PR #3615, opened 2025-07-19
feat: add disableAutoFilter prop to Combobox for custom filtering

### LINKED ISSUE #3328: Allow passing a custom filter function to the Combobox component

Similarly to the [v2 Autocomplete component](https://www.skeleton.dev/components/autocomplete?tab=properties), a custom filter option would allow defining our own custom function that fetches and filters the options to display.

Enhancement

_No response_

### LINKED ISSUE #3600: Combobox not updating with reactive data in Svelte 5 ($state)
### Current Behavior

Hi, I'm using the Combobox component from @skeletonlabs/skeleton-svelte with Svelte 5 . I'm updating the `data` prop dynamically based on user input, fetching countries from a local JSON file when the input has 2 or more characters. The `countries` array updates correctly (confirmed via console logs and rendering in a `<select>` element), but the Combobox doesn't reflect these changes until an additional input event occurs. For example, when typing "co", it shows 8 countries in logs but the Combobox remains empty; when I delete a character (to "c"), it shows the previous 8 countries instead of clearing. Using a native `<select>` works perfectly. Is there a known issue with Combobox reactivity in Svelte 5, or am I missing a specific prop/event to make it update dynamically? Here's my code:
```js
<script lang="ts">
	import { Combobox } from '@skeletonlabs/skeleton-svelte';

	interface ComboboxData {
		label: string;
		value: string;
		emoji: string;
	}

	interface CountryResponse {
		translations: { es: string };
		iso3: string;
		emoji?: string;
	}

	const COUNTRIES_DATA: CountryResponse[] = [
		{ translations: { es: 'Argentina' }, iso3: 'ARG', emoji: '🇦🇷' },
		{ translations: { es: 'Bolivia' }, iso3: 'BOL', emoji: '🇧🇴' },
		{ translations: { es: 'Colombia' }, iso3: 'COL', emoji: '🇨🇴' },
		{ translations: { es: 'Costa Rica' }, iso3: 'CRI', emoji: '🇨🇷' },
		{ translations: { es: 'Comoras' }, iso3: 'COM', emoji: '🇰🇲' },
		{ translations: { es: 'Congo' }, iso3: 'COG', emoji: '🇨🇬' },
		{ translations: { es: 'Chile' }, iso3: 'CHL', emoji: '🇨🇱' },
		{ translations: { es: 'Ecuador' }, iso3: 'ECU', emoji: '🇪🇨' },
		{ translations: { es: 'México' }, iso3: 'MEX', emoji: '🇲🇽' },
		{ translations: { es: 'Perú' }, iso3: 'PER', emoji: '🇵🇪' }
	];

	let countries: ComboboxData[] = $state([]);
	let selectedCountry = $state(['']);

	async function searchCountries(input: string): Promise<void> {
		try {
			console.log(input.length, 'size');
			if (input.length >= 2) {
				console.log('search');
				const filtered = COUNTRIES_DATA.filter(
					(country) =>
						country.translations?.es &&
						country.translations.es.toLowerCase().startsWith(input.toLowerCase())
				);
				console.log('FILTERED:', filtered);
				countries = [
					...filtered.map(({ translations, iso3, emoji }) => ({
						label: translations.es ?? '',
						value: iso3 ?? '',
						emoji: emoji || ''
					}))
				];
				console.log('Countries updated:', countries);
			} else {
				console.log('clear');
				countries = [];
			}
		} catch (e) {
			console.error('Error fetching countries:', e);
			countries = [];
		}
	}
</script>

<div style="margin-bottom: 24px;">
	<h2
		style="border-bottom: 1px solid #ccc; padding-bottom: 8px; font-size: 1.25rem; font-weight: bold; color: #333; display: flex; align-items: center; gap: 8px;"
	>
		Location
	</h2>

	<div style="display: grid; grid-template-columns: 1fr; gap: 24px; max-width: 800px;">
		<div style="margin-bottom: 16px;">
			<label
				style="display: flex; align-items: center; gap: 8px; font-size: 0.95rem; font-weight: 500; color: #444;"
				for="country"
			>
				Country
			</label>
			<Combobox
				data={countries}
				value={selectedCountry}
				onValueChange={(e) => (selectedCountry = e.value)}
				onInputValueChange={(e) => searchCountries(e.inputValue)}
				label=""
				placeholder="Select..."
			>
				{#snippet item(item)}
					<div style="display: flex; justify-content: space-between; width: 100%; gap: 8px;">
						<span>{item.label}</span>
						<span>{item.emoji}</span>
					</div>
				{/snippet}
			</Combobox>

			<select bind:value={selectedCountry} style="width: 100%; padding: 6px; margin-top: 8px;">
				{#each countries as country}
					<option value={country.value}>{country.label} {country.emoji}</option>
				{/each}
			</select>
		</div>
	</div>
</div>


```

### Expected Behavior

- The Combobox should dynamically update its dropdown to reflect changes in the `data` prop immediately.
- Typing 2+ characters (e.g., "co") should display filtered countries (e.g., "Colombia", "Costa Rica").
- Deleting to <2 characters (e.g., "c") should clear the dropdown.
- The behavior should match a native `<select>`, which renders the same data correctly.

### Steps To Reproduce

1. Set up a Svelte 5 project with `@skeletonlabs/skeleton-svelte`.
2. Copy and paste the code attached.
3. Type "co" in the Combobox input.
4. Observe that the Combobox does not display the filtered countries, despite the data updating correctly (verified via console logs and a native `<select>`).
5. Delete to "c" and observe that the Combobox shows the previously filtered countries instead of an empty dropdown.

_No response_

### Environment Information

 System:
    OS: Linux 6.11 Linux Mint 22.1 (Xia)
    CPU: (16) x64 AMD Ryzen 7 7730U with Radeon Graphics
    Memory: 8.13 GB / 15.03 GB
    Container: Yes
    Shell: 5.2.21 - /bin/bash
  Binaries:
    Node: 22.13.0 - ~/.volta/tools/image/node/22.13.0/bin/node
    npm: 10.9.2 - ~/.volta/tools/image/node/22.13.0/bin/npm
    pnpm: 9.15.4 - ~/.volta/bin/pnpm
  Browsers:
    Brave Browser: 138.1.80.115
    Chrome: 138.0.7204.92
  npmPackages:
    @skeletonlabs/skeleton: ^3.1.4 => 3.1.4 
    @skeletonlabs/skeleton-svelte: ^1.2.4 => 1.2.4 

### More Information

I noticed that when I type “Co” and the comboBox doesn't show anything if I remove the focus and put it back on the input it loads correctly, seems like a synchronization problem?

Yeison07:
## Linked Issue

Closes #3600
Closes #3328

## Description

This PR addresses the issues reported in #3600, where the Combobox component experiences desynchronization between the provided data and the filtered options. The root cause is the internal "contains" filter logic, which always executes before any external handlers (e.g., onInputValueChange). This results in the data being one step behind, causing lag in scenarios involving dynamic filtering, such as minimum length checks or asynchronous API fetches.

To resolve this, I've introduced an optional boolean prop disableAutoFilter (default: false). When set to true, the internal filtering is skipped, allowing users full control over data updates via their own onInputValueChange handler. This ensures seamless handling of custom filtering logic without interference from the component's built-in behavior.

Additionally, a reactive $effect has been added to synchronize the options with the data prop when auto-filtering is disabled, maintaining reactivity.

This approach preserves backward compatibility, as the default behavior remains unchanged.

**Alternatives Considered**
I evaluated a more flexible option, such as introducing a filterFn prop to allow custom filter callbacks. However, the disableAutoFilter prop provides a simpler and more accessible solution for users, especially beginners, while still enabling advanced customization.

## Checklist

- [X] Your branch should be prefixed with: `docs/`, `feature/`, `chore/`, `bugfix/`
- [X] Contributions should target the `main` branch
- [X] Documentation should be updated to describe all relevant changes
- [X] Run `pnpm check` in the root of the monorepo
- [X] Run `pnpm format` in the root of the monorepo
- [X] Run `pnpm lint` in the root of the monorepo
- [X] Run `pnpm test` in the root of the monorepo
- [X] If you modify `/package` projects, please supply a Changeset

## Changesets

### DIFF
--- a/packages/skeleton-svelte/src/components/Combobox/Combobox.svelte
+++ b/packages/skeleton-svelte/src/components/Combobox/Combobox.svelte
@@ -8,7 +8,7 @@
 		data = [],
 		label = '',
 		zIndex = 'auto',
-		disableAutoFilter = false,
+		filter = (item: T, value: string) => item.label.toLowerCase().includes(value.toLowerCase()),
 		// Base
 		base = '',
 		width = '',
@@ -46,7 +46,8 @@
 	}: ComboboxProps<T> = $props();
 
 	// Zag
-	let options = $state.raw(data);
+	let internalOptions = $state(data);
+	const options = $derived(filter === null ? data : internalOptions);
 	const collection = $derived(
 		combobox.collection({
 			items: options,
@@ -59,26 +60,24 @@
 	const id = $props.id();
 	const service = useMachine(combobox.machine, () => ({
 		id: id,
-		collection: collection,
+		get collection() {
+			return collection;
+		},
 		...zagProps,
 		onOpenChange(event) {
-			options = data;
+			if (event.open && filter !== null) {
+				internalOptions = data;
+			}
 			zagProps.onOpenChange?.(event);
 		},
 		onInputValueChange(event) {
-			if (!disableAutoFilter) {
-				const filtered = data.filter((item) => item.label.toLowerCase().includes(event.inputValue.toLowerCase()));
-				options = filtered;
+			if (filter !== null) {
+				internalOptions = data.filter((item) => filter(item, event.inputValue));
 			}
 			zagProps.onInputValueChange?.(event);
 		}
 	}));
 
-	if (disableAutoFilter) {
-		$effect(() => {
-			options = data;
-		});
-	}
 	const api = $derived(combobox.connect(service, normalizeProps));
 	const triggerProps = $derived(mergeProps(api.getTriggerProps(), { onclick }));
 </script>
--- a/packages/skeleton-svelte/src/components/Combobox/Combobox.svelte
+++ b/packages/skeleton-svelte/src/components/Combobox/Combobox.svelte
@@ -66,7 +66,9 @@
 		...zagProps,
 		onOpenChange(event) {
 			if (event.open && filter !== null) {
-				internalOptions = data;
+				internalOptions = data.filter((item) => filter(item, api.inputValue));
+			} else if (event.open) {
+				internalOptions = [...data];
 			}
 			zagProps.onOpenChange?.(event);
 		},
--- a/packages/skeleton-svelte/src/components/Combobox/Combobox.svelte
+++ b/packages/skeleton-svelte/src/components/Combobox/Combobox.svelte
@@ -8,6 +8,7 @@
 		data = [],
 		label = '',
 		zIndex = 'auto',
+		disableAutoFilter = false,
 		// Base
 		base = '',
 		width = '',
@@ -65,11 +66,19 @@
 			zagProps.onOpenChange?.(event);
 		},
 		onInputValueChange(event) {
-			const filtered = data.filter((item) => item.label.toLowerCase().includes(event.inputValue.toLowerCase()));
-			options = filtered;
+			if (!disableAutoFilter) {
+				const filtered = data.filter((item) => item.label.toLowerCase().includes(event.inputValue.toLowerCase()));
+				options = filtered;
+			}
 			zagProps.onInputValueChange?.(event);
 		}
 	}));
+
+	if (disableAutoFilter) {
+		$effect(() => {
+			options = data;
+		});
+	}
 	const api = $derived(combobox.connect(service, normalizeProps));
 	const triggerProps = $derived(mergeProps(api.getTriggerProps(), { onclick }));
 </script>
--- a/packages/skeleton-svelte/src/components/Combobox/Combobox.svelte
+++ b/packages/skeleton-svelte/src/components/Combobox/Combobox.svelte
@@ -8,7 +8,6 @@
 		data = [],
 		label = '',
 		zIndex = 'auto',
-		filter = (item: T, value: string) => item.label.toLowerCase().includes(value.toLowerCase()),
 		// Base
 		base = '',
 		width = '',
@@ -27,9 +26,10 @@
 		positionerBase = '',
 		positionerClasses = '',
 		// Content
-		contentBase = 'card p-2',
+		contentBase = 'card p-2 overflow-y-auto',
 		contentBackground = 'preset-outlined-surface-200-800 bg-surface-50-950',
 		contentSpaceY = 'space-y-1',
+		contentMaxHeight = 'max-h-80',
 		contentClasses = '',
 		// Option
 		optionBase = 'btn justify-start w-full',
@@ -46,8 +46,8 @@
 	}: ComboboxProps<T> = $props();
 
 	// Zag
-	let internalOptions = $state(data);
-	const options = $derived(filter === null ? data : internalOptions);
+	let options = $derived(data);
+
 	const collection = $derived(
 		combobox.collection({
 			items: options,
@@ -64,19 +64,20 @@
 			return collection;
 		},
 		...zagProps,
-		onOpenChange(event) {
-			if (event.open && filter !== null) {
-				internalOptions = data.filter((item) => filter(item, api.inputValue));
-			} else if (event.open) {
-				internalOptions = [...data];
+		async onOpenChange(event) {
+			if (zagProps.onOpenChange) {
+				zagProps.onOpenChange(event);
+				return;
 			}
-			zagProps.onOpenChange?.(event);
+			options = data;
 		},
-		onInputValueChange(event) {
-			if (filter !== null) {
-				internalOptions = data.filter((item) => filter(item, event.inputValue));
+		async onInputValueChange(event) {
+			if (zagProps.onInputValueChange) {
+				zagProps.onInputValueChange(event);
+				return;
 			}
-			zagProps.onInputValueChange?.(event);
+			const filtered = data.filter((item) => item.label.toLowerCase().includes(event.inputValue.toLowerCase()));
+			options = filtered;
 		}
 	}));
 
@@ -123,7 +124,7 @@
 				<!-- Content (list) -->
 				<nav
 					{...api.getContentProps()}
-					class="{contentBase} {contentBackground} {contentSpaceY} {contentClasses}"
+					class="{contentBase} {contentBackground} {contentSpaceY} {contentClasses} {contentMaxHeight}"
 					style="z-index: {zIndex}"
 				>
 					{#each options as option (option.label)}
--- a/.changeset/cyan-brooms-run.md
+++ b/.changeset/cyan-brooms-run.md
@@ -1,5 +1,0 @@
----
-"@skeletonlabs/skeleton-svelte": patch
----
-
-feature: Added disableAutoFilter prop to Combobox for custom filtering control.
--- a/.changeset/cyan-brooms-run.md
+++ b/.changeset/cyan-brooms-run.md
@@ -1,0 +1,5 @@
+---
+"@skeletonlabs/skeleton-svelte": patch
+---
+
+feature: Added disableAutoFilter prop to Combobox for custom filtering control.
--- a/.changeset/sharp-crabs-smash.md
+++ b/.changeset/sharp-crabs-smash.md
@@ -2,5 +2,4 @@
 "@skeletonlabs/skeleton-svelte": patch
 ---
 
-Added customizable `filter` prop to Combobox for flexible filtering, replacing `disableAutoFilter`. Supports custom filter functions, null to disable internal filtering, and maintains default behavior
-  
+chore: Ensure Comboxbox data is reactive, enable custom filtering, and set max content height
--- a/.changeset/sharp-crabs-smash.md
+++ b/.changeset/sharp-crabs-smash.md
@@ -1,0 +1,6 @@
+---
+"@skeletonlabs/skeleton-svelte": patch
+---
+
+Added customizable `filter` prop to Combobox for flexible filtering, replacing `disableAutoFilter`. Supports custom filter functions, null to disable internal filtering, and maintains default behavior
+  
--- a/packages/skeleton-svelte/src/components/Combobox/types.ts
+++ b/packages/skeleton-svelte/src/components/Combobox/types.ts
@@ -8,9 +8,8 @@ export interface ComboboxProps<T extends ComboboxItem> extends Omit<combobox.Pro
 	label?: string;
 	/** Set a positioner style for z-index, ex: "10" */
 	zIndex?: string;
-
-	/** Set to true to disable the automatic internal filtering, allowing full custom control over data updates via onInputValueChange. Defaults to false. */
-	disableAutoFilter?: boolean;
+	/** Custom filter function for items based on input value. Defaults to case-insensitive label includes. Set to `null` to disable internal filtering and manage data updates manually via `onInputValueChange`. */
+	filter?: ((item: T, value: string) => boolean) | null;
 
 	// Base ---
 	/** Set base classes for the root element. */
--- a/packages/skeleton-svelte/src/components/Combobox/types.ts
+++ b/packages/skeleton-svelte/src/components/Combobox/types.ts
@@ -50,8 +50,7 @@ export interface ComboboxProps<T extends ComboboxItem> extends Omit<combobox.Pro
 	contentBackground?: string;
 	/** Set space-y classes for the content. */
 	contentSpaceY?: string;
-	/** Maximum height of the combobox options list container. Controls vertical scrolling when there are many options. Default: 'max-h-80'.
-	 * Use Tailwind CSS classes (e.g., 'max-h-64') or a valid CSS value (e.g., '300px'). */
+	/** Set max-h classes for the content. */
 	contentMaxHeight?: string;
 	/** Provide arbitrary classes for the content. */
 	contentClasses?: string;
--- a/packages/skeleton-svelte/src/components/Combobox/types.ts
+++ b/packages/skeleton-svelte/src/components/Combobox/types.ts
@@ -9,6 +9,9 @@ export interface ComboboxProps<T extends ComboboxItem> extends Omit<combobox.Pro
 	/** Set a positioner style for z-index, ex: "10" */
 	zIndex?: string;
 
+	/** Set to true to disable the automatic internal filtering, allowing full custom control over data updates via onInputValueChange. Defaults to false. */
+	disableAutoFilter?: boolean;
+
 	// Base ---
 	/** Set base classes for the root element. */
 	base?: string;
--- a/packages/skeleton-svelte/src/components/Combobox/types.ts
+++ b/packages/skeleton-svelte/src/components/Combobox/types.ts
@@ -8,8 +8,6 @@ export interface ComboboxProps<T extends ComboboxItem> extends Omit<combobox.Pro
 	label?: string;
 	/** Set a positioner style for z-index, ex: "10" */
 	zIndex?: string;
-	/** Custom filter function for items based on input value. Defaults to case-insensitive label includes. Set to `null` to disable internal filtering and manage data updates manually via `onInputValueChange`. */
-	filter?: ((item: T, value: string) => boolean) | null;
 
 	// Base ---
 	/** Set base classes for the root element. */
@@ -52,6 +50,9 @@ export interface ComboboxProps<T extends ComboboxItem> extends Omit<combobox.Pro
 	contentBackground?: string;
 	/** Set space-y classes for the content. */
 	contentSpaceY?: string;
+	/** Maximum height of the combobox options list container. Controls vertical scrolling when there are many options. Default: 'max-h-80'.
+	 * Use Tailwind CSS classes (e.g., 'max-h-64') or a valid CSS value (e.g., '300px'). */
+	contentMaxHeight?: string;
 	/** Provide arbitrary classes for the content. */
 	contentClasses?: string;
 
--- a/sites/skeleton.dev/src/content/docs/integrations/popover/svelte.mdx
+++ b/sites/skeleton.dev/src/content/docs/integrations/popover/svelte.mdx
@@ -145,6 +145,21 @@ Triggers an anchored popover list when you tap the arrow. Includes auto-suggesti
 	</Fragment>
 </Preview>
 
+Supplying a custom filter:
+
+```ts
+let selectedValue = $state(['']);
+function customSearchFn(e) { ... }
+```
+
+{/* prettier-ignore */}
+```svelte
+<Combobox
+	onValueChange={(e) => (selectedValue = e.value)}
+	onInputValueChange={(e) => customSearchFn(e.inputValue)}
+>...</Combobox>
+```
+
 ### Anatomy
 
 <Anatomy label="Combobox" element="span" isComponent>


endigo9740:
@Yeison07 just confirming I see this is ready for review. I've been away the last week so I'm just getting back on my feet here. It may take me a bit longer to get a proper review in, but I will asap.

endigo9740:
@Yeison07 so reviewing this, I actually like your alternative idea, as we've had had a pending request for custom filtering for a while now:

- https://github.com/skeletonlabs/skeleton/issues/3328

We must preserve the default behavior for backwards compatibility (to avoid a breaking change), so here's my suggestion:

1. Remove the `disableAutoFilter` prop in favor of a `filter` prop (named that if possible)
2. Set the type to take a function of course
3. Then when this is present, it should take over responsibility for the filtering feature
4. See my comment below regarding the use of `$effect`
5. We'll need to provide guidance on the doc page on how users can utilize this

Per the `$effect` I feel like there may be a better way to handle this. I'm going to defer to @Hugos68 to see if he has any suggestions. Hugo let me know your thoughts here. The goal is to ensure `options` state remains reactive if the original `data` changes externally.

packages/skeleton-svelte/src/components/Combobox/Combobox.svelte
```diff
@@ -65,11 +66,19 @@
...
-			options = filtered;
+			if (!disableAutoFilter) {
+				const filtered = data.filter((item) => item.label.toLowerCase().includes(event.inputValue.toLowerCase()));
+				options = filtered;
+			}
 			zagProps.onInputValueChange?.(event);
 		}
 	}));
+
+	if (disableAutoFilter) {
```
Hugos68: Instead of:
```js
let options = $state.raw(data);

if (disableAutoFilter) {
  $effect(() => {
	  options = data;
  });
}
```
We should do:
```js
let options = $derived(data);
```

packages/skeleton-svelte/src/components/Combobox/Combobox.svelte
```diff
@@ -65,11 +66,19 @@
 			zagProps.onOpenChange?.(event);
 		},
 		onInputValueChange(event) {
-			const filtered = data.filter((item) => item.label.toLowerCase().includes(event.inputValue.toLowerCase()));
-			options = filtered;
+			if (!disableAutoFilter) {
```
Hugos68: We should remove the if check and do:
```js
options = data.filter(filter);
```

Hugos68: Also, the default filter right now should be the default value for the filter in our props, so :
filter = (item, value) => item.label.toLowerCase().includes(value);

Hugos68 (changes requested):

endigo9740:
@Yeison07 just checking on the status of this and the requested changes.

Yeison07:
hi @endigo9740 , checking this out, been a bit busy and couldn't, hope to finish it tonight.

packages/skeleton-svelte/src/components/Combobox/Combobox.svelte
Yeison07: I implemented $derived for options, as you suggested, to avoid using $state.raw and $effect, making the logic more declarative and reactive. Instead of using disableAutoFilter, I introduced a prop filter that allows the user to pass a custom filter function or null. If filter is null, options directly uses data with $derived(filter === null ? data : internalOptions). If filter is provided, I maintain an internal state (internalOptions) that is updated in onInputValueChange with the custom filtering. This eliminates the need for $effect and allows for a more flexible API to control filtering from the parent component.

packages/skeleton-svelte/src/components/Combobox/Combobox.svelte
Yeison07: The prop filter now uses the suggested default filter, (item, value) => item.label.toLowerCase().includes(value.toLowerCase()), maintaining compatibility with the previous behavior. Instead of options = data.filter(filter), I use an internal state internalOptions and $derived(filter === null ? data : internalOptions) to give the parent component full control over filtering. If filter is null, the component uses data directly, which allows to handle complex cases such as input length checks or asynchronous requests without interference from internal filtering.

Yeison07:
Hey @endigo9740   @Hugos68, I tested my code under three scenarios I think cover the main use cases for the Combobox:

1. **Passing filter: null**: This gives the parent full control over filtering, which I think is the best way to handle complex cases like input length checks or async API fetches. It keeps the component simple, avoids extra props, and lets the user decide how to handle the data. I went with this instead of overcomplicating the component internally, and it feels like a solid way to give users total flexibility.
2. **Not passing filter (default behavior)**: This keeps the component's original behavior, using the default filter (item.label.toLowerCase().includes(value.toLowerCase())). It's great for static lists or simple drop-in use cases where you don't need extra logic. No surprises here, just works as expected.
3. **Passing a custom filter function**: This was the trickiest case. It works well for switching filters (like contains to startsWith), but I ran into an issue: when the component first gets focus, onOpenChange loads the full data without applying preconditions (like input length checks). This happens because it resets internalOptions to data. After you clear and type again, the filter works fine, but that first focus feels inconsistent. I'm not sure if this is the expected behavior or if we should tweak onOpenChange to apply the filter with the current input value on open.

Overall, I think the filter prop with null support and internalOptions makes the component flexible, especially for dynamic cases like async fetches or custom preconditions. Since this is my first time working with Zag, I might've missed some edge cases. Let me know what you think about my approach in the latest commit and if the onOpenChange behavior is okay or needs a fix. Thanks for the feedback! :thumbsup:

endigo9740:
@Yeison07 so first things first, this looks much better.

Reviewing the original Zag implementation - which we strive to stick as close to as possible - I see filtering only occurs in `onInputValueChange()`, per their docs:

```ts
onInputValueChange({ inputValue }) {
      const filtered = comboboxData.filter((item) =>
        item.label.toLowerCase().includes(inputValue.toLowerCase()),
      )
      const newOptions = filtered.length > 0 ? filtered : comboboxData
      options = newOptions
},
```

But I would expect we could perhaps remove the need for `internalOptions` if we use [$state.snapshot](https://svelte.dev/docs/svelte/$state#$state.snapshot). Which could result in something like this:

```ts
let {
    // ...
    filter = (item: T, value: string) => item.label.toLowerCase().includes(value.toLowerCase()),
    // ...
}: ComboboxProps<T> = $props();
```
```ts
let options = $derived(data);
```
```ts
onInputValueChange({ inputValue }) {
    const filtered = $state.snapshot(options).filter((item) => filter(item, inputValue));
    options = filtered.length > 0 ? filtered : options;
},
```

This would:

- Keep the default filter as the default `filter` prop value as you have it.
- Keep a single derived instance of `options` for reactivity
- Take a snapshot of the `options` just before the filter is applied (like cloning)

To be honest I haven't used `snapshot()` heavily, so I may be making a false assumption here. And I'm not sure if there's implications to to checking the filter state, or it being explicitly `null`, for other parts of the logic such as `onOpenChange`.

I'll also again ask @Hugos68 to chime in if he has any input on your code or my suggestion here.

One final thing @Yeison07 - when we do settle on the custom filter solution. We'll need to document this alongside the Combobox itself. Likely as a small paragraph + code snippet that explains that you can replace the filtering option, and showcasing a simple use case. Let me know if you need help with this.

Also make sure to update the changset to represent the shift to the customizable `filter`

Yeison07:
Hey @endigo9740, @Hugos68,

I made some minor adjustments to the `filter` prop implementation to address the `onOpenChange` inconsistency we discussed. The changes ensure that custom filters (including `filter={null}`) work consistently with preconditions and async updates by applying the filter with the current input value on open. As per our Discord conversation, I kept `internalOptions` with `$state` since `$derived` is immutable in Svelte 5.20. Let me know if this looks good, and I can mark the related comments as resolved.

For the documentation, I’ve been testing with a local demo page that showcases the three main use cases: `filter={null}` with async search and preconditions, a custom `startsWith` filter, and the default behavior. I haven’t committed it yet since it’s a playground for internal testing, but I can share it or work with you to create the official documentation page. Would you prefer I push the demo page to the PR, or should we refine it together for the docs? I also updated the changeset to reflect the new `filter` prop.

endigo9740:
@Yeison07 I see you've modified your existing source with a number of changes.

Just curious if you tested any of my recommendations here:
https://github.com/skeletonlabs/skeleton/pull/3615#issuecomment-3155610009

My expectation was to see either some form of these changes implemented, or a heads up that they were not working as expected. Currently it feels like this feedback was overlooked.

If you need me to take a stab at testing these changes directly tomorrow I can. Likewise I am available to coordinate on the doc updates. If you can ensure all of your changes are committed by then, I can do a pass on these in the morning.

Yeison07:
Hey @endigo9740

Thanks for the feedback! I made adjustments to the `filter` prop implementation to address the `onOpenChange` inconsistency, ensuring that custom filters (including `filter={null}`) are applied with the current input value on open. This fixes the issue where all data was shown initially when using a custom filter.

Regarding your suggestion to use `$state.snapshot` and eliminate `internalOptions` (#3615 (comment)), I tested it but found that assigning to `options` (a `$derived` variable) causes an error in Svelte 5.20 due to its immutability, as we discussed on Discord. Modifying `$derived` from the child requires Svelte 5.25+, so I kept `internalOptions` with `$state` to handle mutable state while ensuring reactivity for external `data` updates. Let me know if you think we should explore another approach or if this is okay to resolve the comment.

For the documentation, I have a local demo page that I used for testing the three use cases: `filter={null}` with async search and preconditions, a custom `startsWith` filter, and the default behavior. I haven’t committed it yet since it’s a playground for internal testing, but I can push it to the PR or share it with you to refine for the official docs. Let me know how you’d like to proceed with the documentation—I’d appreciate your help with this, as you offered!

I also updated the changeset to reflect the new `filter` prop and its functionality. All changes are committed, so feel free to review them. If you want to test anything specific tomorrow, let me know how I can assist.

Thanks again! 👍

endigo9740:
@Yeison07 thanks for explaining - ok in that case let me go hands on tomorrow. Don't worry about adding any docs. I'll handle that as I go. Thanks for the quick follow up.

endigo9740:
@Yeison07 @Hugos68 so I've done a bit of research and testing this morning, and have some notable findings.

First off, per your original request Yeison, it looks like you were already aware that `onOpenChange` and `onInputValueChange` can be overwritten via the overflow props. I didn't author this component, so I wasn't aware this was possible. But IMO this is a superior alternative to the `filter` prop we've been testing. We should revert and favor that option instead.

That said, as you stated in your original issue, there is a problem with the order of operations where these custom functions merely run AFTER the internal versions, rather than replacing them outright. For that I'd suggest we consider an alternative like this - where there's a simple logic gate allowing custom props to take precedence:

```ts
onOpenChange(event) {
    if (zagProps.onOpenChange) {
        zagProps.onOpenChange(event)
        return;
    }
    options = data;
},
onInputValueChange(event) {
    if (zagProps.onInputValueChange) {
        zagProps.onInputValueChange(event);
        return;
    }
    const filtered = data.filter((item) => item.label.toLowerCase().includes(event.inputValue.toLowerCase()));
    options = filtered;
}
```

Also per how we handle `options` - I see the original implementation followed Zag's recommendation and uses:
```ts
let options = $state.raw(data);
```

I don't think we should deviate from this, as Zag has the right idea and is leaning into the performance benefits of the [raw method](https://svelte.dev/docs/svelte/$state#$state.raw). The gotcha of course being it doesn't adhere to piecemeal changes. It's more like Svelte Writables from Svelte 4 where the entire array or object within has to be reassigned. I think that's a safer default for performance reasons for most users. It also explains why some folks have reported issues where changing `data` doesn't appear to be reactive - so this may be another item to bring up in our docs.

A couple other quick takeaways:

1. This is somewhat out of scope of the original ask, but if you don't mind doing me a favor, let's add a new `contentMaxHeight` style prop to control the height of content popover. By default it's has an infinite height, where as something like `max-h-80` would be great. We'll need to add `overflow-y-auto` to the `contentBase` styles to support this as well.
2. Speaking of performance, there does appear to be some bottleneck to how many items can be fed to the combobox. But Hugo and I are discussing how we wish to proceed on this. Currently I'm leaning towards a recommended cap - to avoid poor UX where folks cram too many options into this (ex: hundreds or thousands of options is excessive)

Note I've asked Hugo to review and comment on my above findings if anything is off base. But for now, I think the logic gate approach will do what you and other folks need when it comes to custom filtering. Ping me once these changes are in place and we'll review and go from there. Thanks!

Yeison07:
@endigo9740 You're absolutely right. Since my initial solution was a flag, I was stuck in the mindset of using a prop and didn't see that we could simply overwrite the filter functions directly, haha. Let me see if I understand you correctly. The idea is not to incorporate more internal states, derivatives, or effects. I was trying a few things, and in the end, I came up with the idea that if the user overwrites the onInputValueChange function, we check if it's a promise and wait. If not, we simply use the raw static data directly. But this violates the signature of the zagProps.onInputValueChange function, and I don't think it will perform well. The only thing I can think of is, as in one of my first commits, to add an effect and condition it so that if a custom onInputValueChange function is sent, it triggers updates in options. Here's a snippet.

```js
let options = $state.raw(data);

const collection = $derived(
		combobox.collection({
			items: options,
			// Map data structure
			itemToValue: (item) => item.value,
			itemToString: (item) => item.label
		})
	);
	const service = useMachine(combobox.machine, () => ({
		id: id,
		get collection() {
			return collection;
		},
		...zagProps,
		async onOpenChange(event) {
			if (zagProps.onOpenChange) {
				zagProps.onOpenChange(event);
				return;
			}
			options = data;
		},
		async onInputValueChange(event) {
			if (zagProps.onInputValueChange) {
				zagProps.onInputValueChange(event);
				return;
			}
			const filtered = data.filter((item) => item.label.toLowerCase().includes(event.inputValue.toLowerCase()));
			options = filtered;
		}
	}));

	$effect(() => {
	     if (!!zagProps.onInputValueChange) {
			options = data;
		}
	});
```

Regarding the change in styles, I think it worked well, as you mentioned.

```js
const {
		....
		// Content
		contentBase = 'card p-2 overflow-y-auto',
		contentBackground = 'preset-outlined-surface-200-800 bg-surface-50-950',
		contentSpaceY = 'space-y-1',
		contentClasses = '',
		contentMaxHeight = 'max-h-80',
		...zagProps
	}: ComboboxProps<T> = $props();
	
	<div {...api.getPositionerProps()} transition:fade={{ duration: 100 }} class="{positionerBase} {positionerClasses}">
			{#if options.length > 0}
				<!-- Content (list) -->
				<nav
					{...api.getContentProps()}
					class="{contentBase} {contentBackground} {contentSpaceY} {contentClasses} {contentMaxHeight}"
					style="z-index: {zIndex}"
				>
```

```html
             <Combobox
		               data={asyncData}
				value={asyncSelected}
				onValueChange={(e) => (asyncSelected = e.value)}
				onInputValueChange={(e) => searchCountries(e.inputValue)}
				label="Select Country (Async)"
				placeholder="Type 3+ characters..."
				contentMaxHeight="max-h-20"
			>
				{#snippet item(item)}
					<div class="flex justify-between w-full gap-2">
						<span>{item.label}</span>
						<span>{item.emoji}</span>
					</div>
				{/snippet}
			</Combobox>
```


image

endigo9740:
@Yeison07 that all looks well, except I don't quite follow on the use of the `$effect`. It's my understanding that Zag doesn't not expect us to need this as it's not part of their original source. Does simply replacing `onInputValueChange` not suffice? Can you describe the scenario where that doesn't work?

> to add an effect and condition it so that if a custom onInputValueChange function is sent, it triggers updates in options

Just to verify again, when using `$state.raw(data)` the ONLY way to update data externally and the component render the changes is to reassign the array. You cannot update the items within the array piecemeal. This is the trade off to aid performance.

So for example:

```ts
const externalDataExample = [ ... ];

externalDataExample.push({ ... }); // this will NOT update data
externalDataExample[0].someKey = someNewValue; // this will NOT update data

// ---

externalDataExample = someNewArray; // this WILL update data
```
```html
<Combobox data={externalDataExample} />
```

So I need to verify you're updating `data` properly before we assume that an extra `$effect` is required. Aside from that all other changes look good to implement.

Yeison07:
Hey @endigo9740 , if we remove the effect, the combobox does not update correctly in cases where the list of items is updated. For example, if you try this code without the effect, nothing will load in the combobox. In this case, I am completely reassigning the data sent to the combobox with one filtered by me. If you $inspect the component, you will see that it does not register any changes in collections internally.
When reading the raw state documentation, I saw exactly what you mentioned, and in theory it should work unless I am reassigning incorrectly. What do you think?

```js

let customFilterData: ComboboxData[] = $state(
		COUNTRIES_DATA.map(({ translations, iso3, emoji }) => ({
			label: translations.es ?? '',
			value: iso3 ?? '',
			emoji: emoji || ''
		}))
	);
	
	let customFilterSelected: string[] = $state([]);
	function customFilterHandler(event: { inputValue: string }): void {
		const filtered = COUNTRIES_DATA.filter((country) =>
			country.translations.es.toLowerCase().startsWith(event.inputValue.toLowerCase())
		).map(({ translations, iso3, emoji }) => ({
			label: translations.es ?? '',
			value: iso3 ?? '',
			emoji: emoji || ''
		}));
		customFilterData = filtered;
	}
```

```js
			<Combobox
				data={customFilterData}
				value={customFilterSelected}
				onValueChange={(e) => (customFilterSelected = e.value)}
				onInputValueChange={customFilterHandler}
				label="Select Country (Custom Filter)"
				placeholder="Type to filter..."
				contentMaxHeight="max-h-80"
			>
				{#snippet item(item)}
					<div class="flex justify-between w-full gap-2">
						<span>{item.label}</span>
						<span>{item.emoji}</span>
					</div>
				{/snippet}
			</Combobox>
```

endigo9740:
@Yeison07 I'm AFK and not able to test this directly. But try pulling out and testing updating `$state.raw` using a minimal example. Just set an array, then try to update it with the `=` operator like you're doing above. You can even test this in the Svelte REPL:

https://svelte.dev/tutorial/svelte/welcome-to-svelte

If that works, then we know the issues is internal to the component. If it doesn't, then the assignment is not sufficient to trigger the reactivity update.

Let's start there.

Yeison07:
@endigo9740 Reactivity does not seem to work even if it is reassigned.

```svelte
<script>
	import Child from './Child.svelte'
	let data =$state([
		'A','E','I','O','U'
	])
	let dataStatic=$state.raw(data);
	const filter=()=>{
		let filtered=data.filter(e=> e==='E')
		data=filtered
	}

	const reset=()=>{
		data=['A','E','I','O','U']
	}
</script>

<button
	onclick={filter}
	>Filter</button>
<button onclick={reset}>Reset</button>
<p>Father data {data}</p>
<Child data={data}/>
<p>Static same componente {dataStatic}</p>
```

```svelte
<script>
	const {data=[]}=$props()

	let options = $state.raw(data);

</script>

<p>Child data {options}</p>
```



Peek 2025-08-14 14-22

endigo9740:
Thanks for confirming @Yeison07

@Hugos68 any idea what were doing wrong here? How can we ensure `$state.raw` can update?

Hugos68 (approved):

endigo9740:
@Yeison07 so a few things here. Let me explain...

1. Disregard Hugo's approval. We're not ready to merge yet. That was a miscommunication on my part.
2. Hugo explained to me that `$state.raw()` will not behave as I expected. Any performance gain would be negligible. And that we should just go ahead and use `$derived()` for `options`.
3. We currently use `$derived()` but the code is not current with all suggested changes. Remember we need to revert the custom `filter` and `internalOptions` changes. Then add the logic gates [per my recommendations here](https://github.com/skeletonlabs/skeleton/pull/3615#issuecomment-3180175084).

So once these three things are done, this PR is ready to merge:

1. The custom filtering functionality added in this PR is revertered
2. The suggest logic gates are added to those two internal methods
3. `options` is set with `$derived(data)` instead of `$state.raw(data)`

Not to rush you, but we're currently prepping some administrative changes to the repo to begin work on Skeleton v4 initiatives. As such, we need to quickly close out any remaining PRs asap. As such, let me know if you can make these changes in the next 24 hours - otherwise I will make them directly tomorrow and likely go ahead and merge the PR immediately following.

Thanks for working with me on this!

Yeison07:
Hi @endigo9740 , I'm going to upload the latest changes we discussed, but with the derived instead of the raw state.
 To avoid causing you any further delays, and given that I'm likely to be a little busy, I'd appreciate it if you could take it and adjust it as you see fit. Thank you both for your work on the library, and I'll be around if I can help with any other issues in the future. Good luck with the new version!

Hugos68 (approved):

PR #3656, opened 2025-08-26
chore: move common classes into skeleton-common

Hugos68:


.changeset/forty-badgers-clap.md
```diff
@@ -0,0 +1,7 @@
+---
+"@skeletonlabs/skeleton-svelte": patch
+"@skeletonlabs/skeleton-react": patch
+---
+
+chore: use shared classes
```
endigo9740: Two suggestions:

- Probably worth of calling this a feature, so move up to `minor`
- Let's make the description a bit more verbose

"Implements the new common style system for all framework components"

.changeset/little-files-notice.md
```diff
@@ -0,0 +1,6 @@
+---
+"@skeletonlabs/skeleton": patch
+---
+
+chore: add framework dependency sources
```
endigo9740: Not sure what this entails. And other than the extra bullet point on the changelog won't affect the version if we have multiple changelogs.

packages/skeleton-common/src/classes/accordion.ts
```diff
@@ -0,0 +1,8 @@
+export const classesAccordion = {
+	root: 'w-full grid !!gap-2',
```
endigo9740: @Hugos68 make sure to remove the test syntax here!

.changeset/forty-badgers-clap.md
Hugos68: But it isn't a feature from a public API perspective, only internals have changed. Calling this a minor goes against semver.

Hugos68: Regardless, I've renamed this to a minor

PR #3662, opened 2025-08-28
Prepare monorepo for V4

Hugos68:
This PR preps our monorepo for contributions for V4.

Including:
- Setting up framework packages properly
- Removing v3 components
- Removing v3 component docs
- Replacing `/composed` with `/` (default index now)
- Updates all dependencies


sites/skeleton.dev/src/content/docs/components/accordion/meta.mdx
```diff
@@ -1,9 +1,10 @@
 ---
 title: Accordion
 description: Divide content into collapsible sections.
-srcSvelte: '/src/components/Accordion'
-srcReact: '/src/components/Accordion'
+srcSvelte: '/src/components-composed/accordion'
```
endigo9740: If we're taking an axe to the v3 component docs and lib/components per each framework package, let's make sure we also remove any reference to `-composed`. That was only needed for the v4 preview on the v3 branch. Going forward the new components should just take the place of the old components - both in terms of packages and docs.

sites/skeleton.dev/src/content/docs/get-started/installation/nextjs.mdx
```diff
@@ -47,6 +47,7 @@ import ProcessStep from '@components/docs/ProcessStep.astro';
         @import '@skeletonlabs/skeleton';
         @import '@skeletonlabs/skeleton/optional/presets';
         @import '@skeletonlabs/skeleton/themes/cerberus';
+        @import '@skeletonlabs/skeleton-react';
```
endigo9740: Let's keep the import orders:

- Core (required)
- Framework package (required)
- Theme (optional or swappable)
- Presets (optional)

As the first couple are pretty important.

This goes for all install guides.

sites/skeleton.dev/src/examples/guides/cookbook/chat/Example.svelte
```diff
@@ -135,7 +134,7 @@
 								: 'bg-surface-hover-token'}"
 							onclick={() => (currentPersonId = person.id)}
 						>
-							<Avatar src="https://i.pravatar.cc/?img={person.avatar}" name={person.name} size="size-8" />
+							<!-- <Avatar src="https://i.pravatar.cc/?img={person.avatar}" name={person.name} size="size-8" /> -->
```
endigo9740, line 137: We might be shooting ourselves in the foot to purge all v3 components before replacing with the v4 version, as we may have to disable a wide number of components in the docs. It might be better to purge these one-by-one and do a sweep through the docs to update to the new standards.

Otherwise our docs may not be usable for a while. Depends on how many components we use. I know there's a ton of on the homepage.

Let's review on Discord.

PR #3679, opened 2025-09-02
feat: tabs

### LINKED ISSUE #3646: v4 Component: Tabs

### Requirements:

- [ ] Svelte Component
- [ ] React Component
- [ ] Test Cases
- [ ] Documentation

Enhancement

- https://github.com/skeletonlabs/skeleton/issues/3472
- https://github.com/skeletonlabs/skeleton/issues/3629

Hugos68:
## Linked Issue

Closes #3646 

## Description

Adds the Tabs component

## Checklist

## Changesets


endigo9740 (changes requested):
@Hugos68 This looks great. The implementation looks perfect. The ONLY issue I can really raise here is that we have a small bug that we fixed in the v3 version.

Namely, when you hover a tab, but you're in that small "gap" between the indicator and button it doesn't register clicks. Here's a quick visualization:

Screenshot 2025-09-03 at 11 42 12 AM

You might refer to the v3 styles, but I think we fixed this by flipping from a margin -> padding. So instead of the the inner element having an outer margin, we gave the outer element a inner padding. That allows it to register clicks.

To put it another way, EVERYTHING within the green space should be clickable.

Hugos68:
@endigo9740 This should now be fixed

endigo9740:
@Hugos68 hmm so that's changed the design. There's no longer a floating button above the indicator, it now "touches". I don't hate that, but given the `btn` style can have rounded edges this might be more noticeable the stronger the border radius is.

If we can't match the floating style ala v3, let's go ahead and implement `rounded-none` for the bottom two corners. Note that for any design changes we should likely ping Bohdan on Discord to alert him, so he can update the Figma UI Kit.

That'll provide a shape more like this:

Screenshot 2025-09-03 at 2 47 55 PM

Hugos68:
I like the floating better, but why does the area between the button and the tabs be clickable? There isn't anything there, no hover, no indication of clicking there that should work. I don't see that as a bug?

endigo9740:
@Hugos68 it was reported as a bug by users already in v3. They couldn't figure out why they would sometimes click and it wouldn't register. Turns out they were missing the button element.

But I also prefer the floating button style. So perhaps we just make the judgement call to revert and keep it like it was?

Hugos68:
> @Hugos68 it was reported as a bug by users already in v3. They couldn't figure out why they would sometimes click and it wouldn't register. Turns out they were missing the button element.

I think they thought it was a bug because in V3 there were hover styles when hovering on the so called "dead area" those hover styles aren't present in V4, there is absolutely no reason to think it's clickable since there is no visible feedback when hovering on it.

I've reverted it to floating yeah

endigo9740 (approved):

PR #3696, opened 2025-09-06
feat: popover

### LINKED ISSUE #3649: v4 Component: Popovers (Svelte Only)

### Requirements:

- [ ] Svelte Component: Popover
- [ ] Svelte Component: Tooltip
- [ ] Svelte Component: Combobox
- [ ] Svelte Component: Modal
- [ ] Test Cases
- [ ] Documentation

Enhancement

Unlike most components, note that these are restrict strictly to Svelte, as Floating UI React is available for React users and offers a broader solution.

See also:

- https://github.com/skeletonlabs/skeleton/issues/3559

Hugos68:
## Linked Issue

Closes #3649

## Description

Implements the popover component for Svelte and React

## Checklist

## Changesets


packages/skeleton-common/src/classes/popover.ts
```diff
@@ -0,0 +1,12 @@
+import { defineSkeletonClasses } from '../internal/define-skeleton-classes.js' with { type: 'macro' };
+
+export const classesPopover = defineSkeletonClasses({
+	trigger: 'btn preset-filled',
```
endigo9740: We shouldn't style this part by default. We should include the styles on the Popover itself, like:

```
<Popover.Trigger class="btn preset-filled">
```

The reason being, is this is too opinionated. We're deciding for the user that the trigger will always be a styled button. But if you're attaching the Popover to an icon, image, etc, then these styles will be annoying to override one-by-one.

packages/skeleton-common/src/classes/popover.ts
```diff
@@ -0,0 +1,12 @@
...
+	positioner: '',
+	content: 'preset-filled-surface-100-900 border-1 border-surface-200-800 p-4 rounded-container',
```
endigo9740: Same thing as the button here. We shouldn't assume folks always want a card-like interface for the content area. But we should show folks how to quickly add this.

Make sure to use something more like this:

```
class="card preset-filled-surface-100-900 shadow-xl"
```

- Card covers the rounded sizing
- We no longer do the border outline for cards (that was a v2 thing)
- And we should encourage a shadow for depth

packages/skeleton-common/src/classes/popover.ts
```diff
@@ -0,0 +1,12 @@
+import { defineSkeletonClasses } from '../internal/define-skeleton-classes.js' with { type: 'macro' };
+
+export const classesPopover = defineSkeletonClasses({
+	trigger: 'btn preset-filled',
+	positioner: '',
```
endigo9740, line 5: A note for docs: per all Popover variants (tooltips, etc) I believe the Positioner element controls the z-index, but it's not set with a class. Zag sets it with a custom CSS property like `--z-index:`. So we will need to illustrate how users can manipulate this for ALL popover variants (tooltips, etc). This is 100% required, as it's one of the top 10 support questions we receive.

packages/skeleton-common/src/classes/popover.ts
```diff
@@ -0,0 +1,12 @@
...
+	content: 'preset-filled-surface-100-900 border-1 border-surface-200-800 p-4 rounded-container',
+	arrow: '',
+	arrowTip: 'border-t-1 border-l-1 border-surface-200-800',
```
endigo9740: To match the Trigger/Container, I think we leave this unstyled by default, but show folks how to style it. As it always has to match the container styles to remain flush.

packages/skeleton-common/src/classes/popover.ts
```diff
@@ -0,0 +1,12 @@
...
+	title: 'h4',
```
endigo9740: I know, I'm repeating myself here. But I think we should leave this up to the users as well. We can't know what heading level they are going to use, so setting a default might result in a mismatch. For a11y, the HTML element and class CAN differ - but again, I think we need a doc example to explain this.

Something like this should suffice:

```html
<Popover.Title heading={1} class="h1">Title</Popover.Title>
<Popover.Title heading={2} class="h2">Title</Popover.Title>
<Popover.Title heading={3} class="h3">Title</Popover.Title>
```

playgrounds/skeleton-svelte/src/routes/components/popover/+page.svelte
```diff
@@ -0,0 +1,17 @@
+<script>
```
endigo9740, line 1: Per the doc: In addition to the full component tree's worth of components, We should also show users they can roll their own contain content as they please:

```html
<Popover>
	<Popover.Trigger>Open</Popover.Trigger>
	<Popover.Positioner>
		<Popover.Content>
			<div class="class preset-surface-100-900">
				<h2>Hello World</h2>
				<p>I'm controlling all this.</p>
				<Popover.CloseTrigger class="btn preset-filled">Close</Popover.CloseTrigger>
			</div>
		</Popover.Content>
	</Popover.Positioner>
</Popover>

```

playgrounds/skeleton-svelte/src/routes/components/popover/+page.svelte
```diff
@@ -0,0 +1,17 @@
...
+	import { Popover } from '@skeletonlabs/skeleton-svelte';
```
endigo9740, line 2: Per the docs: demonstrate programmatic control as well without a trigger in place.

```html
<Popover open={isOpen}>
	<Popover.Positioner>
		<Popover.Content>...</Popover.Content>
	</Popover.Positioner>
</Popover>
```

packages/skeleton-svelte/src/components/popover/anatomy/root.svelte
```diff
@@ -0,0 +1,33 @@
...
+	const api = $derived(connect(service, normalizeProps));
+
+	PopoverRootContext.provide({
+		get api() {
+			return api;
+		}
+	});
+</script>
+
+{@render children?.()}
```
endigo9740, line 27: Per all popover variants (tooltips, etc), we should consider wrapping the root with a `<span>`. The reason being is that if you have something like this:

```
<div class="space-y-4>
  <Popover>...</Popover>
</div>
```
The trigger and container may be affected by the extra vertical spacing. Because there's nothing in the DOM intercepting this flow.

The reason for span over div, is that span is inline by default. Meaning it won't make the contents stretch to fill the width. Which is typically the preferred behavior.

Keep in mind, if we do this, we'll need to implement a custom `class` prop/attribute.

Alternatively, if you feel that's too opinionated. We should point out this issue to end users in the docs, and explain that they can wrap the Popover themselves to avoid this issue.

playgrounds/skeleton-svelte/src/routes/components/popover/+page.svelte
```diff
@@ -0,0 +1,17 @@
...
+</script>
+
+<Popover>
+	<Popover.Trigger>Open</Popover.Trigger>
+	<Popover.Positioner>
+		<Popover.Content>
+			<Popover.Arrow>
```
endigo9740: Nitpick: Floating UI shows the arrow at the bottom of the container in their examples. I suggest we match that for readability.

Additionally, make sure we include a doc example showcasing how to implement and style and arrows, and of course indicate that it's optional.

packages/skeleton-common/src/classes/popover.ts
endigo9740: NOTE: I have a demo for this setup - but it's not working

sites/skeleton.dev/src/content/docs/components/popover/svelte.mdx
```diff
@@ -0,0 +1,97 @@
...
+		<Arrow client:visible />
+	</Fragment>
+	<Fragment slot="code">
+		<Code code={ArrowRaw} lang="svelte" />
+	</Fragment>
+</Preview>
+
+## Z-Index
+
+By default we do not take an opinionated stance regarding z-index stacking. The result is the component can sometimes be occluded beneath other elements with a higher index. The Z-Index can controlled by applying a utility class to the `Positioner` component part.
```
endigo9740, line 46: @Hugos68 the example was updated but not the docs. So I've modified this. Including a warning about needing `!`. I've updated both Svelte and React docs here - but you may need to carry this over to the Tooltip, Modal, and Combobox docs as well.

endigo9740 (approved):
@Hugos68 one thing of note here, you'll need to update the other components. But this one is now approved with said changes.

PR #3740, opened 2025-09-15
feat: tags input

### LINKED ISSUE #3647: v4 Component: Tags Input

### Requirements:

- [ ] Svelte Component
- [ ] React Component
- [ ] Test Cases
- [ ] Documentation

Enhancement

_No response_

Hugos68:
## Linked Issue

Closes #3647

## Description

Implements the TagsInput component for React and Svelte

## Checklist

## Changesets


packages/skeleton-svelte/src/components/tags-input/anatomy/item-delete-trigger.svelte
```diff
@@ -0,0 +1,41 @@
...
+
+	const attributes = $derived(
+		mergeProps(tagsinput().getItemDeleteTriggerProps(itemProps()), rest, {
+			class: classesTagsInput.itemDeleteTrigger,
+		}),
+	);
+</script>
+
+{#snippet x()}
+	<X class="size-4" />
```
endigo9740: Consider replacing the Icon with:

```html
<strong>&times;</strong>
```

This is how we [handled it in v3](https://github.com/skeletonlabs/skeleton/blob/v3/packages/skeleton-svelte/src/components/TagsInput/TagsInput.svelte#L81) and it creates a smaller/bolder X symbol that looks a bit nicer IMO:

Screenshot 2025-09-15 at 11 04 48 AM

I wouldn't mind this for the FileUpload file "close" button icons too. Either way we should probably settle on a convention here.

playgrounds/skeleton-svelte/src/routes/components/tags-input/+page.svelte
```diff
@@ -0,0 +1,25 @@
+<script lang="ts">
+	import { TagsInput } from '@skeletonlabs/skeleton-svelte';
+</script>
+
+<TagsInput>
+	<TagsInput.Label>Label</TagsInput.Label>
+	<TagsInput.Control>
+		<TagsInput.Input placeholder="Enter value..." class="w-full" />
```
endigo9740: Note I've made three changes here:

1. Added a placeholder (rule of thumb: let's always do this for inputs)
2. Add `w-full`
3. Moved this above `<TagsInput.Context>`

This helps replicate the v3 style of display for the component. I don't think this *has* to be the default display, but we should show folks they can both replicate the v3 style, as well as use the "inline" tags ala Zag's defaults.

Maybe implement this as an alternative design example for folks that want to match the v3 styling?

packages/skeleton-common/src/classes/tags-input.ts
```diff
@@ -0,0 +1,18 @@
+import { defineSkeletonClasses } from '../internal/define-skeleton-classes' with { type: 'macro' };
+
+export const classesTagsInput = defineSkeletonClasses({
+	root: 'space-y-2 gap-2',
+	label: 'label-text',
```
endigo9740, line 5: Note that for ALL components that implement a `<label>` element. We need to append `.label-text` to ensure consistent styling.

Technically in Skeleton we handle labels like this:

```html
<label class="label">
    <span class="label-text">Number</span>
    <input type="number" class="input" placeholder="Enter Age" />
</label>
```

But, since Zag opts for labels to be single line and not wrapping, we can't do the split `label/labe-text`. But the `label-text` is important for styling the visible text. It'll make the text slightly smaller.

This is useful when folks need to override this globally, they can just do:

```css
.label-text { /* some overrides */ }
```

(again - this may need to be adjusted retroactively to other existing components; sorry I kept meaning to bring it up but forgot!)

packages/skeleton-common/src/classes/tags-input.ts
```diff
@@ -0,0 +1,18 @@
+import { defineSkeletonClasses } from '../internal/define-skeleton-classes' with { type: 'macro' };
+
+export const classesTagsInput = defineSkeletonClasses({
+	root: 'space-y-2 gap-2',
```
endigo9740: Note that on a couple components I've see you do `flex flex-col gap-2` on the root element. But unfortunately this has a side effect of making the items within stretch to fill the width. This is why the "clear files" button was so wide.

By switching to `w-full space-y-2` we get the same look, without the side effect.

I've already applied this to the FileUpload as well - but keep in mind for other components in the future.

sites/skeleton.dev/src/examples/components/tags-input/react/default.tsx
```diff
@@ -0,0 +1,27 @@
+import { TagsInput } from '@skeletonlabs/skeleton-react';
```
endigo9740, line 1: As always, I recommend viewing the v3 docs for guidance on various examples to showcase:

https://www.skeleton.dev/docs/components/tags-input/svelte

endigo9740 (changes requested):
I've pushed some styles adjustments and made some changes in the Svelte playground to use as reference. Most of these are notes, with the exception of a few suggest examples to add to the docs.

endigo9740:
@Hugos68 still seeing a CI issue tripped here. We'll want to resolve that before we merge or all subsequent PRs will likely result in the same.

Also, I want to do one more pass through the styling for this tomorrow. After that though we can look to merge. Shouldn't take long.

Hugos68:
@endigo9740 Yeah I'm working on the type issue, just a bit frustrating. We won't merge until that's resolved.

PR #3777, opened 2025-09-21
fix: avoid dead loop while transforming ts files in v3 migration script

Shaokun-X:
## Linked Issue

None

## Description

When I was migrating my project from skeleton v2 to v3, the migration script gets stuck (hanging for minutes, console not responding) while transforming some ts files. I believe this is due to the fact that nodes inside `forEachDescendant` at https://github.com/skeletonlabs/skeleton/blob/v3/packages/cli/src/commands/migrate/migrations/skeleton-3/transformers/transform-module.ts#L47 could create new AST nodes identical to the original, causing `forEachDescendant` to revisit them indefinitely.

I am targeting main as this template asked, but maybe v3 is more relevant. Let me know if you want me to switch the target branch. 

## Checklist

## Changesets

### DIFF
--- a/.changeset/purple-tigers-happen.md
+++ b/.changeset/purple-tigers-happen.md
@@ -1,0 +1,6 @@
+---
+"skeleton": patch
+---
+
+Avoid dead loop while transforming ts files in v3 migration script
+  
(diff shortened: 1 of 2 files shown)


Hugos68:
@Shaokun-X Could you provide a test that would fail without this PR? That also would give me a better idea of when it fails.

Shaokun-X:
> @Shaokun-X Could you provide a test that would fail without this PR? That also would give me a better idea of when it fails.

After some investigation, I found it was not actually a "dead loop," but rather a performance issue with `node.replaceWithText` when handling large files. In my project, I have a generated GraphQL schema file with around 26k lines, and for this file each iteration of `forEachDescendant` takes about 500 ms on my machine.

In that case do you still want to me create a test?

Hugos68:
@Shaokun-X All good, we won't have any performance benchmarks but all tests are passing and you indeed same some compute by checking if something has actually changed, thanks!

PR #3859, opened 2025-10-04
feat: individual llms.txt

nullpointerexceptionkek:
## Linked Issue

Closes #{i forget number i will update this later}

## Description

{description}

## Checklist

- [ ] Your branch should be prefixed with: `docs/`, `feature/`, `chore/`, `bugfix/`
- [ ] Contributions should target the `main` branch

## Changesets

### DIFF
--- a/sites/skeleton.dev/src/lib/generate-llm.ts
+++ b/sites/skeleton.dev/src/lib/generate-llm.ts
@@ -95,14 +95,30 @@ async function processPreviewBlocks(content: string, language: string): Promise<
 		if (!importPath) {
 			continue;
 		}
-		const resolvedPath = importPath.replace('@examples', './src/examples');
+		const resolvedPath = importPath.replace('@examples', './src/examples').replace(/^@\/examples/, './src/examples');
+
+		// Try multiple extensions if the path doesn't have one (Currently there are no ts, js, and html files)
+		const extensions = ['.tsx', '.svelte', '.astro', '.ts', '.js', '.jsx', 'txt', 'html', ''];
 		let fileContent = '';
-		try {
-			fileContent = await fs.readFile(path.resolve(resolvedPath), 'utf8');
-		} catch (error) {
-			console.error('Error reading file:', resolvedPath, error);
-			fileContent = '// Error loading file';
+		let foundFile = false;
+
+		for (const ext of extensions) {
+			try {
+				const pathToTry = resolvedPath + ext;
+				fileContent = await fs.readFile(path.resolve(pathToTry), 'utf8');
+				foundFile = true;
+				break;
+			} catch (error) {
+				// Try next extension
+				continue;
+			}
 		}
+
+		if (!foundFile) {
+			console.error('Error reading file (tried all extensions):', resolvedPath);
+			fileContent = '// Error loading file, please report this issue.';
+		}
+
 		const replacement = `\`\`\`${language}\n${fileContent}\n\`\`\``;
 		content = content.replace(previewBlock, replacement);
 	}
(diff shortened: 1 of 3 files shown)


sites/skeleton.dev/src/lib/generate-llm.ts
```diff
@@ -95,14 +95,30 @@ async function processPreviewBlocks(content: string, language: string): Promise<
 		if (!importPath) {
 			continue;
 		}
-		const resolvedPath = importPath.replace('@examples', './src/examples');
+		const resolvedPath = importPath.replace('@examples', './src/examples').replace(/^@\/examples/, './src/examples');
+
+		// Try multiple extensions if the path doesn't have one (Currently there are no ts, js, and html files)
+		const extensions = ['.tsx', '.svelte', '.astro', '.ts', '.js', '.jsx', 'txt', 'html', ''];
```
Hugos68: I think instead of having an extensions array and all that we should use a glob matcher like this:
await glob(path.replace(extension, ".*")) and then jsut pick the first option that returns from that match. This would support infinite extensions and would result in less fs.readFile calls. (glob is a lot faster). You can use the `tinyglobby` package for this.

sites/skeleton.dev/src/pages/docs/[...slug]/llms.txt.ts
```diff
@@ -0,0 +1,30 @@
...
+	const { page } = props as { page: Awaited<ReturnType<typeof getCollection<'docs'>>>[number] };
+
+	// Check if this is a component or integration that has a meta entry
+	const hasMeta = ['components/', 'integrations/'].some((id) => page.id.startsWith(id));
+	const metaEntry = hasMeta ? await getEntry('docs', page.id.replace(/\/[^/]*$/, '/meta')) : undefined;
+
+	// Generate the documentation
+	const content = await generateSinglePageDocumentation(page, metaEntry);
+
+	return new Response(content, {
```
Comment (Hugos68) [inline sites/skeleton.dev/src/pages/docs/[...slug]/llms.txt.ts]:
```suggestion
	return Response.text(content);
```

This helper function already sets the correct headers.

sites/skeleton.dev/src/lib/generate-llm.ts
```diff
@@ -291,6 +307,52 @@ async function processIntegrations(framework: Framework): Promise<string> {
...
+	let bodyContent = docEntry.body ?? '';
+
+	// Apply transformations based on document type
+	const language = framework === 'react' ? 'tsx' : framework === 'svelte' ? 'svelte' : 'html';
+	bodyContent = await processPreviewBlocks(bodyContent, language);
+	bodyContent = await processApiTables(bodyContent, docEntry.id);
+
+	content += bodyContent;
+
+	// Final cleanup
```
Hugos68: Could you maybe add a comment to what this does? It's not very descriptive.

Hugos68 (changes requested):

sites/skeleton.dev/src/pages/docs/[...slug]/llms.txt.ts
Comment (louisqli) [reply sites/skeleton.dev/src/pages/docs/[...slug]/llms.txt.ts]: When I write this function I get an error saying it does not exist

Comment (Hugos68) [reply sites/skeleton.dev/src/pages/docs/[...slug]/llms.txt.ts]: Oh my bad, apparently Response.json exists, but not Response.text, weird. Just leave it then.

sites/skeleton.dev/src/lib/generate-llm.ts
```diff
@@ -95,14 +95,30 @@ async function processPreviewBlocks(content: string, language: string): Promise<
 		if (!importPath) {
 			continue;
 		}
-		const resolvedPath = importPath.replace('@examples', './src/examples');
+		const resolvedPath = importPath.replace('@examples', './src/examples').replace(/^@\/examples/, './src/examples');
```
Hugos68: @examples no longer exist, and if you want to resolve the path correctly you must replace `@` with `./src`, you can see the tsconfig.json on how it works.

PR #3891, opened 2025-10-08
docs: universal alerts

Hugos68:
This PR adds the `Card` component for the docsite to standardize info/warning cards.

Before:
image
After:
image

Before:
image
After:
image

This <Card> components uses the Skeleton design system and is more consisten than our current system.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -334,8 +334,8 @@ catalogs:
       specifier: 1.0.7
       version: 1.0.7
     vite-plugin-transform-lucide-imports:
-      specifier: 0.2.0
-      version: 0.2.0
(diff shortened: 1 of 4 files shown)


endigo9740:
@Hugos68 these are Alerts, can we name them as such? I might also consider using the styles from the Cookbook section of the same name. The filled background is hard to work with when links are included.

Hugos68:
Sounds good, will do, just thought we should have a universal component for this instead of winging it everytime.

Hugos68:
@endigo9740 Done

Hugos68:
@endigo9740 Can this be merged?

endigo9740:
@Hugos68 LGTM

PR #3895, opened 2025-10-09
feat: collapsible

Hugos68:
## Linked Issue

Closes -

## Description

Implements the Collapsible component

## Checklist

## Changesets

### DIFF
--- a/sites/skeleton.dev/src/components/examples/components/collapsible/react/controlled.tsx
+++ b/sites/skeleton.dev/src/components/examples/components/collapsible/react/controlled.tsx
@@ -6,7 +6,7 @@ export default function Controlled() {
 
 	return (
 		<Collapsible open={open} onOpenChange={(details) => setOpen(details.open)}>
-			<Collapsible.Trigger className="btn preset-filled">{open ? 'Open' : 'Close'}</Collapsible.Trigger>
+			<Collapsible.Trigger className="btn preset-filled">{open ? 'Close' : 'Open'}</Collapsible.Trigger>
 			<Collapsible.Content>
 				Lorem ipsum dolor sit amet consectetur, adipisicing elit. Velit pariatur error soluta voluptate, accusantium modi repudiandae omnis
 				eligendi, suscipit repellat impedit architecto neque sequi mollitia autem dicta quae ipsum et?


endigo9740:
@Hugos68 so far everything looks good. I'll revisit for styling tomorrow. We'll probably go fairly minimal on built-in styles to keep this one flexible as it can be applied to a lot different use cases. But not completely headless like overlay components.

sites/skeleton.dev/src/components/examples/components/collapsible/react/dir.tsx
```diff
@@ -0,0 +1,13 @@
+import { Collapsible } from '@skeletonlabs/skeleton-react';
+
+export default function Dir() {
+	return (
+		<Collapsible dir="rtl">
```
endigo9740, line 5: I don't believe `dir` is working as expected. Neither your default styles, or anything I threw at this was working. Essentially any sort of default alignment should be flipped.

So the root is full width, and the trigger inline (with button styles) which means it'll be left aligned on LTR, and should flip to right align on RTL.

I've removed the flex styles, but even adding an explicit `flex flex-col items-start` did not flip when RTL was set.

We might want to report this upstream.

sites/skeleton.dev/src/content/docs/components/collapsible/svelte.mdx
```diff
@@ -0,0 +1,57 @@
...
+
+<Preview client:load>
+	<Fragment slot="preview">
+		<Default client:visible />
+	</Fragment>
+	<Fragment slot="code">
+		<Code code={DefaultRaw} lang="svelte" />
+	</Fragment>
+</Preview>
+
```
endigo9740, line 27: It would be great to have a "Controlled" example like other components. Both frameworks of course.

sites/skeleton.dev/src/content/docs/components/collapsible/svelte.mdx
```diff
@@ -0,0 +1,57 @@
...
+<Preview client:load>
+	<Fragment slot="preview">
+		<Indicator client:visible />
+	</Fragment>
+	<Fragment slot="code">
+		<Code code={IndicatorRaw} lang="svelte" />
+	</Fragment>
+</Preview>
+
```
endigo9740, line 49: It would be great to have an "Alignment" example where we show folks how to center the trigger over the content area.

This would just require a few classes on the root:

`flex flex-row items-center`

packages/skeleton-common/src/classes/collapsible.ts
```diff
@@ -0,0 +1,8 @@
+import { defineSkeletonClasses } from '../internal/define-skeleton-classes' with { type: 'macro' };
+
+export const classesCollapsible = defineSkeletonClasses({
+	root: 'w-full data-[state=open]:space-y-2',
```
endigo9740, line 4: FYI I've set `data-[state=open]:space-y-2` on the root. As this appears to be the only realistic way to add a default vertical space between the trigger and content. 

General idea is we ONLY want the spacing when in the open state, otherwise it leaves a small gap at the bottom of the trigger that's always on. Because the "content" element appears to always be loaded in the DOM, just not just visible.

The problem is, the data attribute only seems to work in Svelte. React is ignoring this. I have no idea why - would love a sanity check on your end.

sites/skeleton.dev/src/components/examples/components/collapsible/react/dir.tsx
Hugos68: Already reported and fixed upstream, next update to zag will fix

PR #4133, opened 2025-11-20
task: use direct package accessing during development

Hugos68:
This PR aims to setup our monorepo so that Vite apps can directly access our packages instead of through a dev script running during development. This will improve hot reloading and simplifies the monorepo structure and config. We do this using `clean-package` which will swap our package.json before publishing.


Hugos68:
@copilot In order for this to work we need to get rid of all usage of `path` aliases in the `packages/*` directories. Can you go into each package inside that directory, remove the `paths` option from `tsconfig.json` and replace all usage of `paths`, mainly `@/` in the `src` and `test` of each packages and replace it with the correct relative path? You can run `pnpm check` in the monorepo to keep checking wether it's fixed or not.

PR #4135, opened 2025-11-21
perf: monorepo improvements

Hugos68:
## General Improvements

- We're using [Rolldown](https://vite.dev/guide/rolldown) now to speed up bundling across our apps and packages
- We're using [Vitest Browser Mode](https://vitest.dev/guide/browser/) which runs our tests in actual browsers, not a simulated environment to increase accuracy, as a bonus we're able to remove all testing library code and dependencies as vitest supports react and svelte first party.
- We're using [`clean-package`](https://github.com/roydukkey/clean-package) to enable direct package access during development for `skeleton-react` and `skeleton-svelte`
- We're using [`@prettier/plugin-oxc`](https://prettier.io/blog/2025/06/23/3.6.0#javascript) and [`--experimental-cli`](https://prettier.io/blog/2025/06/23/3.6.0#cli) for much faster formatting.

## Performance Metrics

| Task | Time Before (in seconds) | Time After (in seconds) |
| - | - | - |
| Installing Dependencies (`pnpm i`) | 20 | 8 |
| Vercel Deployment (skeleton.dev) | 187 | 163 |
| Vercel Deployment (themes.skeleton.dev) | 54 | 31 |
| Formatting Project (`pnpm format`) | 17 | 10 |

Non-measurable, but honorary mentions:
- Developing (dev mode) is *much* more responsive due to less packages needing to run in parallel.

### DIFF
--- a/playgrounds/skeleton-react/package.json
+++ b/playgrounds/skeleton-react/package.json
@@ -5,7 +5,7 @@
 	"scripts": {
 		"dev": "vite dev",
 		"check": "tsc",
-		"postinstall": "tsr sync"
+		"postinstall": "tsr generate"
 	},
 	"dependencies": {
 		"@tanstack/react-router": "catalog:",


Hugos68:
@copilot Can you create a PR against this branch that removes all uses of `@/` paths alias inside all `packages/*` projects, this PR removes the aliases as prep for some upcoming stuff. Just grep @/ inside import paths in pacakges/* and replace them with the correct relative path instead.

Hugos68:
@copilot This PR moves from testing library to vitest browser mode, please update every test to use vitest browser mode setup:
test('UserProfile handles loading and data states', async () => {
  const { getByText } = render(<UserProfile userId="123" />)

  // Test loading state
  await expect.element(getByText('Loading...')).toBeInTheDocument()

  // Test for data to load (expect.element auto-retries)
  await expect.element(getByText('User: John')).toBeInTheDocument()
})

Hugos68:
@copilot Your last PR migrated to vites tbrowser mode test setup which is great, but the render function is async, please make all tests that utilize `render` async and await the render call.

Hugos68:
@copilot The svelte API actually doesn't need async, so remove any unneccesary ascyn or await keywordsa in the svelte package

Hugos68:
@copilot Per the test packages, instead of doing `accordion/index.test.{ts,tsx}` I want to simplify everything, let's do:
accordion.test.{ts,tsx}


Then add the sibling test component with the name of the component so accordion/test.{svelte,tsx} becomes accordion.{svelte,tsx}. Can you do this for both skeleton-svelte and skeleton-react? Make a PR against this branch.

PR #4162, opened 2025-11-24
task: style to tailwind

Hugos68:
This PR moves from all CSS variable usage from `style` to tailwind native `class`/`className` expressions, aligning more with how Tailwind works and simplifying what users need to do themselves otherwise.


Hugos68:
@copilot Can you leave a review? --spacing() is an actual tailwind function for your information.

PR #4165, opened 2025-11-25
task: vitest rolldown-vite fix

Hugos68:
Overrides `vitest>vite` to our own catalogged version (rolldown-vite).

(Shoutout to @dominikg for helping figure this out on Discord!)

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,8 @@
 	},
 	"pnpm": {
 		"overrides": {
-			"vite": "catalog:"
+			"vite": "catalog:",
+			"astro>vite": "^6.4.1"
 		}
 	},
 	"packageManager": "pnpm@10.23.0"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -35,5 +35,10 @@
 		"node": "22.21.1",
 		"pnpm": "10.23.0"
 	},
+	"pnpm": {
+		"overrides": {
+			"vitest>vite": "catalog:"
```
dominikg: would recommend unconditional `"vite"` selector here, unless you have a need for different versions elsewhere

dominikg:
```suggestion
			"vite": "catalog:"
```

Hugos68: Done that, thanks! Didn't even know you could do that unconditionally

Hugos68: Actually, reverting this, because this will cause Astro to use `vite-rolldown` which completely breaks our builds.

PR #4174, opened 2025-11-27
docs: redirect v3 Figma page to new etesie.dev location

mmailaender:
## Description

The `v3.skeleton.dev/figma` page still pointed to the old Polar checkout, while the new `etesie.dev/figma` page uses our updated Creem setup. Some users still land on the v3 page via Google, and today someone ended up buying the kit twice because of that.
To avoid this confusion, I’ve updated the v3 page to redirect directly to the new one — this should prevent unnecessary trouble both for users and for support/refunds.

## Checklist

- [ ] Your branch should be prefixed with: `docs/`, `feature/`, `task/`, `bugfix/`
- [ ] Contributions should target the `main` branch

## Changesets


endigo9740:
Sorry @mmailaender but we have a policy not to update older docs or packages after a new version is released. Only the newest version remains maintained. If we made an exception here, we would have to do the same for others. And that's not a path we want to go down.  That said, not a lot of folks use the older docs, and they are clearly marked as an archive.

PR #4175, opened 2025-11-27
task: simplify skeleton common

Hugos68:


Hugos68:
@copilot Can you change all the component imports from `import { classesAccordion } from '@skeletonlabs/skeleton-common'` to `import * as classes from '@skeletonlabs/skeleton-common/classes'?

PR #4182, opened 2025-11-29
feature: native base layer

Hugos68:
This PR moves our component styles from tailwind utilities to css, this allows us to move away from the `skb` layer and use the native `@layer base` directly, this has several benefits, notable ones are:

1. Remove `@custom-variant skb` (simplification)
2. Only ships utilities once instead of twice (skb: and non skb:)
3. Skip the build step and thus have instant hot reload during development
4. Simplify contributing, just write some css, no Tailwind-In-JS or imports/exports 😎 
5. Tailwind intellisense [without our previous hack](https://github.com/skeletonlabs/skeleton/blob/main/packages/skeleton-common/intellisense.css)

### DIFF
--- a/packages/skeleton-common/src/components/toast.css
+++ b/packages/skeleton-common/src/components/toast.css
@@ -21,7 +21,6 @@
 		border: 1px solid var(--color-surface-200-800);
 
 		@apply card preset-filled-surface-50-950;
-		
 
 		&[data-state='closed'] {
 			transition:


Hugos68:
@copilot Inspect how I converted the `accordion` styles from the ts format to css, apply this to all components.

Hugos68:
@copilot Now please remove all the imports from @skeletonlabs/skeleton-common/classes inside the component packages, you can remove that because the CSS is now imported seperately.

endigo9740:
@Hugos68 everything looks good so far. The only thing of note here is that rewriting all our styles does mean we need to do a lot of manual regression testing. Even one small style being off could cause a regression. This will take more time than I have today. But we can continue to review over the next couple days as times allows.

Hugos68:
Totally understand that. We need to be very thorough as there were already some instances where to due to specificity differences styles needed slight tweaking to behave like before. I already did a somewhat thorough pass, but you can never be sure enough.

endigo9740:
@Hugos68 I've run through the checking styling for all these. There's two small issues to report, both visible on the Toast component docs page:

1. None of the examples are stacking vertically as multiple messages are queued (both frameworks)
2. The Overlap example has visible code showing (Svelte only; present in production too)

Hugos68:
@endigo9740 

> None of the examples are stacking vertically as multiple messages are queued (both frameworks)

This was already the case, we can fix it in this PR though!

> The Overlap example has visible code showing (Svelte only; present in production too)

I was not aware, will aim to fix 👍

PR #4188, opened 2025-12-01
task: remove skeleton build step

Hugos68:
This PR removes SCSS and thus the build step introducing major DX improvements, it's a bit more verbose, but our design system really doesn't change, if at all. And it's still managble. Tailwind does the same with their themes, just a giant list of theme variables, no build step either. I think this is the right approach.

### DIFF
--- a/.changeset/modern-ideas-hide.md
+++ b/.changeset/modern-ideas-hide.md
@@ -1,0 +1,6 @@
+---
+"@skeletonlabs/skeleton": patch
+---
+
+task: remove sass build step
+  


endigo9740:
@Hugos68 overall I like the idea. It does simplify things for values that haven't changed since introduction.

The only oddity I've seen in the Theme menu/pages don't show the surface color for each theme. Not sure the cause, but these are definitely working in the production.

Screenshot 2025-12-01 at 11 26 24 AM

endigo9740:
@Hugos68 LGMT, just note the minor merge conflict.

PR #4223, opened 2025-12-08
Feature: Navigation Anchor and Button

### LINKED ISSUE #4063: Navigation component should include an anchor/button part

https://www.skeleton.dev/docs/components/navigation/svelte

Currently the navigation component includes parts for many of the visual elements that make up the component structure. However, it feels odd that buttons and anchors but be implemented and managed by users. By introducing a component we can likely simplify the process for handling responsive styling as the layout is changed.

If possible it would be great to utilize a single component for this that can act both as a `<button>` or `<a>`.

- `<Navigation.Trigger>`

But this may pose an issue with prop/attribute type suggestions. So we may consider introducing this as two unique component parts:

- `<Navigation.Anchor>`
- `<Navigation.Button>`

Enhancement

_No response_

endigo9740:
## Linked Issue

Closes #4063

## Description

Adds two new component parts, that automatically set styles and adhere to layout changes:

- `Navigation.Trigger` -> an button tag
- `Navigation.TriggerAnchor` -> an anchor tag
- `Navigation.TriggerText` -> the inner button/anchor label text

## AI Disclosure

## Checklist

## Changesets

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
 		}
 	},
 	"engines": {
-		"node": "24.11.1",
+		"node": "24.x",
 		"pnpm": "10.24.0"
 	},
 	"packageManager": "pnpm@10.24.0"


packages/skeleton-common/src/components/navigation.css
```diff
@@ -74,6 +75,48 @@
...
+		&[data-layout='sidebar'] {
+			justify-content: start;
+			padding-inline: --spacing(2);
+			width: 100%;
+		}
+	}
+
+	&[data-part='trigger-text'] {
+		&[data-layout='bar'] {
+			font-size: 10px;
```
Hugos68: We should not use `px` for text, this is inaccessible for folks requiring a certain zoom level to be set for readable text (iOS zoom, chrome zoom etc). We should use em/rem, in our case, --spacing should suffice, you can use comma values with --spacing so if --spacing(1) is too big you can do --spacing(0.75)

Hugos68: Actually, on further testing, `px` does appear to scale, but still, whenever somebody in their theme will set a bigger base font size, this component part won't adhere to the set base font size. I still think there is a reason to use --spacing here.

endigo9740: I didn't realize spacing could use decimal values. In that case, I'll opt for that.

packages/skeleton-common/src/components/navigation.css
```diff
@@ -74,6 +75,48 @@
...
+			width: 100%;
+		}
+	}
+
+	&[data-part='trigger-text'] {
+		&[data-layout='bar'] {
+			font-size: --spacing(2.5);
+		}
+
+		&[data-layout='rail'] {
```
Hugos68, line 114:
```suggestion
		&[data-layout='bar'],
		&[data-layout='rail'] {
```

packages/skeleton-common/src/components/navigation.css
```diff
@@ -74,6 +75,48 @@
...
+		}
+
+		&[data-layout='sidebar'] {
+			justify-content: start;
+			padding-inline: --spacing(2);
+			width: 100%;
+		}
+	}
+
+	&[data-part='trigger-text'] {
```
Hugos68, line 109: Perhaps add `pointer-events: none` to prevent that weird text cursor when hovering over a Trigger

Edit: Or a less destructive pointer: default or w/e, I just dislike the cursor effect when hovering over a trigger

Hugos68 (changes requested):

packages/skeleton-common/src/components/navigation.css
endigo9740: I understand this would be more efficient, but I've kept this separately intentionally for code readability. I want all three states explicit.

packages/skeleton-common/src/components/navigation.css
endigo9740: This is a bug. I'm not sure the cause, but it's not something we should add an explicit class for to override. We should try to find the source of the issue and resolve it.

Currently I've only seen it in the Astro Docs > Navigation > Toggle example. But we should check the Svelte/React playgrounds to see if it carries over there.

PR #4250, opened 2026-01-12
fix: ComboBox selected item visibility bug

### LINKED ISSUE #4239: ComboBox selected element using wrong color
### Current Behavior

I've noticed both in my app and on the Skeleton UI site that the colour scheme used by the combo box for a selected element when reopening the combobox is quite hard to read. This issue is present in both light and dark mode on multiple themes (the screenshots in this issue were taken on Cerberus).

As far as I can tell, the issue is the selected item's background colour is being set to the background colour used for highlighting items whilst using the text colour for a selected item is used. This becomes more obvious when looking at a multi select ComboBox which appears to apply the highlight colour to the first selected element.

The same poor colour selection applies when a user delibretly mouses over a selected elemement

The below screenshots were taken from https://www.skeleton.dev/docs/svelte/framework-components/combobox

Image

Image

### Expected Behavior

The selected element shouldn't be using the highlight background color. 
Also, when hovering the mouse over a selected item, a lest contrasty colour should be used (or even no highlighting at all)

### Steps To Reproduce

To replicate, just select an element in the combobox, then reopen the dropdown.

_No response_

### Environment Information

_No response_

### More Information

_No response_

AdeboyeDN:
## Linked Issue

Closes #4239

## Description

This addresses issue #4239 where selected items in the ComboBox were 
hard to read when reopening the dropdown.

## AI Disclosure

## Checklist

## Changesets

### DIFF
--- a/.changeset/tangy-eels-pick.md
+++ b/.changeset/tangy-eels-pick.md
@@ -1,0 +1,7 @@
+---
+"skeleton": patch
+"@skeletonlabs/skeleton-common": patch
+---
+
+task: Default styling for disabled menu items
+  
--- a/packages/skeleton-common/src/components/menu.css
+++ b/packages/skeleton-common/src/components/menu.css
@@ -40,4 +40,9 @@
 	&[data-part='separator'] {
 		@apply hr;
 	}
+
+	&[data-disabled] {
+		pointer-events: none;
+		opacity: 0.5;
+	}
 }
--- a/playgrounds/skeleton-react/src/routes/components/menu/index.tsx
+++ b/playgrounds/skeleton-react/src/routes/components/menu/index.tsx
@@ -48,6 +48,9 @@ function Page() {
 						<Menu.Item value="export">
 							<Menu.ItemText>Export</Menu.ItemText>
 						</Menu.Item>
+						<Menu.Item value="disabled" disabled={true}>
+							<Menu.ItemText>Disabled</Menu.ItemText>
+						</Menu.Item>
 					</Menu.Content>
 				</Menu.Positioner>
 			</Portal>
--- a/sites/skeleton.dev/src/components/examples/design/typography/lists.astro
+++ b/sites/skeleton.dev/src/components/examples/design/typography/lists.astro
@@ -26,11 +26,11 @@
 	<!-- Ordered List -->
 	<section class="space-y-4">
 		<p class="text-lg font-bold">Ordered List</p>
-		<ul class="list-inside list-decimal space-y-2">
+		<ol class="list-inside list-decimal space-y-2">
 			<li>Id maxime optio soluta placeat ea eaque similique consectetur dicta tempore.</li>
 			<li>Repellat veritatis et harum ad sint reprehenderit tenetur, possimus tempora.</li>
 			<li>Lorem ipsum dolor sit amet consectetur adipisicing elit harum ad sint.</li>
-		</ul>
+		</ol>
 	</section>
 
 	<hr class="hr" />
--- a/sites/skeleton.dev/src/components/examples/framework-components/menu/react/disabled.tsx
+++ b/sites/skeleton.dev/src/components/examples/framework-components/menu/react/disabled.tsx
@@ -1,0 +1,28 @@
+import { Menu, Portal } from '@skeletonlabs/skeleton-react';
+
+export default function Disabled() {
+	return (
+		<Menu>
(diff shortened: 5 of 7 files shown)


.changeset/plain-tools-ring.md
```diff
@@ -0,0 +1,6 @@
+---
+"@skeletonlabs/skeleton-common": patch
+---
+
+fixed issue with ComboBox selected element using wrong color
```
endigo9740: Note we use a convention to prefix the changeset description with the same value used for the branch name. In this case it would need: `bugfix: ...`

Covered here:
https://www.skeleton.dev/docs/svelte/resources/contribute/get-started#changesets

packages/skeleton-common/src/components/combobox.css
```diff
@@ -68,6 +68,7 @@
 
 		&[data-highlighted] {
 			@apply preset-tonal;
+			color: var(--color-on-surface);
```
endigo9740: Unfortunately there's no way this change will address the issue:

1. The present provided should already cover contrasting color values
2. We *never* hardcode a single color value, as themes need to adapt to light/dark mode
2. `--color-on-surface` hasn't been a valid Skeleton color property since v2 (we're on v4)

You can find our modern documentation on color values here:
https://www.skeleton.dev/docs/svelte/design/colors

And the modern color properties here:
https://www.skeleton.dev/docs/svelte/get-started/core-api#colors

If for some reason the preset is not working, we would need a color pairing here.

endigo9740 (changes requested):
@AdeboyeDN typically when we see contributions that are using outdated information, it's due to AI contribution. We ask that users volunteer their use of LLMs as part of the PR submission process here:
https://github.com/skeletonlabs/skeleton/pull/4250#issue-3802417084

So either your LLM is using outdated information, or you have somehow draw information from an old documentation source. Make sure you're always using the current/modern docs at this URL:

https://www.skeleton.dev/
(Older docs have `v2`, `v3`, specified in the URL)

If you are making use of LLMs, that's fine, just make sure you:

1. Disclose this information per our request (you can edit the PR post)
2. Are feeding it the correct LLM source, which is available here:

https://www.skeleton.dev/docs/svelte/resources/llms
  
I'll go ahead and give you a chance to rectify this, but if the issue is not solved, and correct properties and protocols used, then I will have to deny and close this PR.
 
 Thanks for your understanding.

AdeboyeDN:
Hi @endigo9740 , apologies, I’ve addressed your comments. Let me know what you think.

packages/skeleton-common/src/components/combobox.css
```diff
@@ -68,6 +68,7 @@
 
 		&[data-highlighted] {
 			@apply preset-tonal;
+			color: var(--color-surface-contrast-950);
```
endigo9740, line 71: Hey @AdeboyeDN thanks for the quick turn around here.

While this is a step in the right direction. I had mentioned the issue with using a fixed color and trying to address this in the combobox styles prior. For one, your solution improves dark mode - but here's light mode now:

Screenshot 2026-01-15 at 8 40 55 AM

Your current fix is aimed at addressing the issue localized to the combobox. If we were going to do that, we would need to use a color pairing to address both modes like so:

```css
color: var(--color-surface-contrast-950-50);
```

Which would use this value:

https://github.com/skeletonlabs/skeleton/blob/main/packages/skeleton/src/base/theme.css#L388

Unfortunately this is not a broad enough change. The core issue is that `preset-tonal` *should* in fact handle the color values automatically. This means the issue is in the preset itself. If we take a look here, we can see what I mean:

https://github.com/skeletonlabs/skeleton/blob/main/packages/skeleton/src/utilities/presets.css#L331

Unlike the other Tonal presets, the generic `preset-tonal` doesn't have `color` values set. So essentially we need to fix the color in the preset, not in the combobox styling:

```cs
@utility preset-tonal {
	background-color: color-mix(in oklab, light-dark(var(--color-surface-950), var(--color-surface-50)) 5%, transparent);
	color: var(--color-surface-contrast-950-50); /* <--- add this */
}
```

This will not only fix the combobox issue, but should ensure the preset works as expected in many different use cases. If you could please, remove the combobox style change, add the change to the preset, and update your changeset. That would be great. In this case the changeset will need two changes:

- Switch from `skeleton-common` to `skeleton`
- Update the description to describe the change to the preset

Please and thanks!

endigo9740:
Closing due to inactivity. I'll have a PR available in a moment that addresses this issue as required.

PR #4268, opened 2026-01-23
Add index paramter to toast for proper overlap styling

### LINKED ISSUE #4264: Toast “overlap” option not working as expected
### Current Behavior

Skeleton’s toast has an `overlap` option in which the toasts are stacked on top of each other with no indication there are multiple toasts. When clicking on the toast, then they are expanded and you can see all of them.

Image

### Expected Behavior

On Zag’s doc site, their implementation of the `overlap` option actually creates a “stack” of toasts, so you can tell that there are multiple toasts.

Image

The expected behavior would be to mirror what Zag does, or at least have some indication of multiple toasts when using the `overlap` option.

### Steps To Reproduce

1. Using Skeleton’s doc site, navigate to the toast, specifically the overlap example
2. Click the button to create multiple overlapping toasts
3. Observe toasts are created, but no indication that there are multiple until you click on the toast

_No response_

### Environment Information

_No response_

### More Information

_No response_

cmintey:
## Linked Issue

Closes #4264

## Description

Toast overlap styling from Zag relies on the use of an `--index` CSS variable, which is created by Zag when the toast is supplied with an `index` parameter. This parameter was left out of Skeleton's implementation of the Zag toast.

This PR adds an optional `index` parameter to a Toast in order to maintain backwards compatibility within the major version. Docs were updated and the Playground was also updated to allow for dynamically changing the toast settings for testing.

While this parameter is optional, I added it to all the docs since it doesn't hurt to be explicit.

## AI Disclosure

## Checklist

## Changesets

### DIFF
--- a/.changeset/twelve-symbols-wave.md
+++ b/.changeset/twelve-symbols-wave.md
@@ -3,5 +3,5 @@
 "@skeletonlabs/skeleton-svelte": minor
 ---
 
-Support proper Toast styling when using the `overlap` option
+bugfix: Support proper Toast styling when using the `overlap` option.
   


packages/skeleton-svelte/src/components/toast/anatomy/group.svelte
```diff
@@ -35,8 +35,8 @@
 	{@render element(attributes)}
 {:else}
 	<div {...attributes}>
-		{#each api.getToasts() as toast (toast.id)}
-			{@render children?.(toast)}
+		{#each api.getToasts() as toast, index (toast.id)}
+			{@render children?.(toast, index)}
```
endigo9740: @cmintey so good news and bad news here.

The good news - it looks like you have found the core issue and provide a solution for this.

The bad news - the current implementation of the fix does not adhere to our design patterns for Skeleton components. The being, leaning into the tools Zag provides us. And trying to avoid going outside their patterns.

While I don't think it's inherently bad to implement the `index` like this. I think we could save a lot of changes here if we simply expand the `toast` instance (which is already an object) to include the `index` as well.

This would look like this:

```
{@render children?.({...toast, ...index}}
```

This would result in the toast object providing this:

```json
{
  "id": "toast:4",
  "removeDelay": 200,
  "type": "info",
  "title": "Toast",
  "description": "This is a toast message.",
  "action": {
    "label": "Undo"
  },
  "stacked": true,
  "gap": 16,
  "message": "PAUSE"
  "index": "..." <-- this would now be available
}
```

Then it would be as simple as consuming this externally like so:

```html
{#snippet children(toast)}
			<Toast {toast}>
```

You should not need to expand the `<Toast>` (root) props to support this either. As we already allow arbitrary attributes to be passed/spread to the template here:
https://github.com/skeletonlabs/skeleton/blob/main/packages/skeleton-svelte/src/components/toast/anatomy/root.svelte#L24

From that point, all that is needed is to ensure the types support the new `index` key.

Another benefit here - users updating to the new package release would just get the fix automatically. No new props to setup for their existing Toast implementation!

Given this would remove many of the changes you made in this PR. I'd understand if you want to close and recreate this PR. No qualms with that. But let's try to keep the impact surface area as small as possible.

If you have any questions about this, let me know.

cmintey: Yeah that makes a ton of sense. I had originally thought of doing that, but didn't want to mess with the existing types. But sounds like it wouldn't be an issue.

I'm happy to try to implement that change!

packages/skeleton-svelte/test/components/toast.svelte
```diff
@@ -5,8 +5,8 @@
 </script>
 
 <Toast.Group {toaster} data-testid="group">
-	{#snippet children(toast)}
-		<Toast {toast} data-testid="root">
+	{#snippet children(toast, index)}
```
endigo9740: Hey @cmintey sorry for the delay. It's been a crazy week.

Looking over your changes here. It looks like you were able to follow most of my suggestions. I'm curious though why the index is still passed as second parameter for the snippet. It's not immediately clear to me.

Did you run into some reason why this could not be folded into `toast`?

My concern here is backwards compatibility.

1. If we don't fold `index` into the `toast` object, the users updating Skeleton have to come in and manually update their code to receive this fix.
2. If we do fold it in, then users updating just immediately get access to the change. It just works out of the box from that point forward.

The former is not a breaking update, as the feature was already broken. But feels like something we should avoid if at all possible.

Just let me know, and thanks for the effort on this!

cmintey: There was no issue with folding the `index` into the `toast`, I had just forgot to revert my previous changes to the test files. Should be all good now!

playgrounds/skeleton-svelte/src/routes/components/toast/+page.svelte
```diff
@@ -1,33 +1,51 @@
...
-				<Toast.Description>{toast.description}</Toast.Description>
-			</Toast.Message>
-			{#if toast.action}
-				<Toast.ActionTrigger>{toast.action.label}</Toast.ActionTrigger>
-			{/if}
-			<Toast.CloseTrigger />
-		</Toast>
-	{/snippet}
-</Toast.Group>
+{#key toaster}
```
endigo9740, line 36: First, the `#key` here - this is a fairly destructive feature to use as it force re-renders this entire component region. I know it's just the playground, but someone might accidentally carry that over to the proper docs by mistake

Do you mind dropping this?

.changeset/twelve-symbols-wave.md
```diff
@@ -0,0 +1,7 @@
+---
+"@skeletonlabs/skeleton-react": minor
+"@skeletonlabs/skeleton-svelte": minor
+---
+
+Support proper Toast styling when using the `overlap` option
```
endigo9740: And second, I know it may come across as a nitpick. But the Changeset description is shown verbatim in our changelog like so:

https://github.com/skeletonlabs/skeleton/blob/main/packages/skeleton-svelte/CHANGELOG.md

You'll note that all non-automated PRs include a prefix for the description. So this would be the change:

```
bugfix: Support proper Toast styling when using the `overlap` option=
```

Per: https://www.skeleton.dev/docs/svelte/resources/contribute/get-started#changesets

playgrounds/skeleton-svelte/src/routes/components/toast/+page.svelte
cmintey: I had to add the key block here because the toast options I added were not taking effect. Wrapping the `createToaster` call in the `$derived` wasn't enough to force a refresh, so the key block was used.

.changeset/twelve-symbols-wave.md
cmintey: Sure thing, I missed that in the contributing docs

PR #4326, opened 2026-04-05
fix(floating-panel): fix body height overflow issue

### LINKED ISSUE #4298: Floating Panel Body's height is too high
### Current Behavior

This rule sets the height of the FloatingPanel.Body to 100%:

```css
[data-scope=floating-panel][data-part=body] {
    background:var(--color-surface-100-900);
    height:100%;
    padding:calc(var(--spacing) * 4);
    overflow-y:auto
  }
```

This causes FloatingPanel.Body to take up 100% of the content area height without
  accounting for the space occupied by FloatingPanel.Header, resulting in vertical overflow. The body content spills outside the panel's visible/resizable bounds.

Image

### Expected Behavior

FloatingPanel.Content should lay out its children in a flex column so that FloatingPanel.Body fills only the remaining vertical space after the header. The body should be scrollable within that remaining space without overflowing the panel.

### Steps To Reproduce

1. Use the FloatingPanel component with Header, Body, and ResizeTrigger children inside Content
2. Add enough content in Body to exceed the panel's default height
3. Open the panel
4. Observe that the body overflows the panel bounds

_No response_

### Environment Information

Chrome: 145.0.7632.160
Edge: Chromium (140.0.3485.54)
Firefox: 145.0.2

### More Information

The fix would likely be adding `display: flex; flex-direction: column;` to the internal styles of FloatingPanel.Content, and `flex: 1; min-height: 0; overflow-y: auto;` to FloatingPanel.Body. This ensures the body respects the header's height within the resizable container.

cyphercodes:
## Description

Fixed FloatingPanel.Body overflowing the panel bounds when content exceeds the default height.

### Problem
The FloatingPanel.Body had "height: 100%" which caused it to take up 100% of the content area height without accounting for the space occupied by FloatingPanel.Header, resulting in vertical overflow.

### Solution
- Changed content wrapper to use flex column layout (`display: flex; flex-direction: column`)
- Changed body from `height: 100%` to `flex: 1; min-height: 0`
- This ensures the body fills only the remaining vertical space after the header

### Changes
- Modified: `packages/skeleton-common/src/components/floating-panel.css`

Fixes #4298

### DIFF
--- a/packages/skeleton-common/src/components/floating-panel.css
+++ b/packages/skeleton-common/src/components/floating-panel.css
@@ -2,6 +2,8 @@
 	&[data-part='content'] {
 		overflow: hidden;
 		border: 1px solid var(--color-surface-300-700);
+		display: flex;
+		flex-direction: column;
 
 		@apply card shadow-lg;
 	}
@@ -46,7 +48,8 @@
 	}
 
 	&[data-part='body'] {
-		height: 100%;
+		flex: 1;
+		min-height: 0;
 		background: var(--color-surface-100-900);
 		padding: --spacing(4);
 		overflow-y: auto;


Hugos68:
@cyphercodes Please supply a changeset

PR #4353, opened 2026-04-13
Transition to v5 theme format

### LINKED ISSUE #3417: Documented classes don't actually exist

### LINKED ISSUE #3517: Default border/ring/divider not applying to Skeleton themes
### Current Behavior

Themes, whether generated or hand-coded, don't apply `--default-border-width` to the `border` utility, or `--default-ring-width` to the `ring` utility. It works as expected if you use `@theme {--default-border-width: 20px;}` rather than `[data-theme='cerberus'] {--default-border-width: 20px;}` however. This can also be seen on [https://themes.skeleton.dev/themes/create](https://themes.skeleton.dev/themes/create) if you change the Border Width value. Unsure if there is an issue with --default-divide-width as well.

### Expected Behavior

Expected the border and ring width options on the create theme page at [https://themes.skeleton.dev/themes/create](https://themes.skeleton.dev/themes/create) to change the border and ring widths, both in the preview and in your own app (assuming you use the `border` and or `ring` utility classes and not a set width).

### Steps To Reproduce

1. Go to the [create theme page](https://themes.skeleton.dev/themes/create).
2. Set the default border width or default divider width to a larger size.
3. The borders and dividers stay the same width, despite the login and card examples both using the `border` utility.

For the --default-ring-width:
1. Set `--default-ring-width: 6px` from the theme create page or your theme css file
2. Apply the `ring` class to an element
3. The ring stays the same width

_No response_

### Environment Information

 System:
    OS: Linux 6.11 Ubuntu 24.04.2 LTS 24.04.2 LTS (Noble Numbat)
    CPU: (8) x64 Intel(R) Core(TM) i7-6700 CPU @ 3.40GHz
    Memory: 5.70 GB / 15.57 GB
    Container: Yes
    Shell: 5.9 - /usr/bin/zsh
  Binaries:
    Node: 22.11.0 - ~/.local/share/mise/installs/node/22.11.0/bin/node
    npm: 10.9.0 - ~/.local/share/mise/installs/node/22.11.0/bin/npm
    pnpm: 10.8.0 - ~/.local/share/pnpm/pnpm
  Browsers:
    Brave Browser: 135.1.77.97
  npmPackages:
    @skeletonlabs/skeleton: ^3.1.2 => 3.1.2 
    @skeletonlabs/skeleton-svelte: ^1.2.1 => 1.2.1

### More Information

Visual example of the problem on the create theme page. Notice that all default widths are set to 6px, but cards with `border` applied still have a 1px border, with the inspector showing that the `.border` class has a declaration of `border-width: 1px;`.

Image

### LINKED ISSUE #4333: Update theme format to match v5 requirements
Per the proposal: https://www.notion.so/Skeleton-v5-RFC-33466a04713c809abb60dad12826a140

## Goal

Update the built-in preset Skeleton themes in the core package to adhere to all required changes.

## Requirements

- [x] Resolve non-standard Tailwind `@theme` property names (ex: `--base-font-color` > `--color-*`)
- [x] Determine a solution for Edge properties: `border | ring | divider | outline`
- [x] Investigate initial support for custom color properties
- [X] Investigate support for a "brand color" property; for use with `accent-color` and more
- [x] Investigate support for the `corner-shape` (used for buttons, cards, etc)

> NOE: these requirements may expand as part of our initial audit and research phase

## Remove Redundant Classes

As part of updating to standardized Tailwind `@theme` properties, this will render many custom implemented utility classes redundant. These should be removed and the [Core API](https://www.skeleton.dev/docs/svelte/get-started/core-api) documentation updated

## Improve Core API Docs

Alongside these other updates, we should aim to improve the display and consumption of the [API reference data](https://www.skeleton.dev/docs/svelte/get-started/core-api). Ideally showing exactly exact rather than summarized variables whenever possible. This will be beneficial to people and LLMs alike, which sometimes fail to parse this information properly. We can keep links to the more detail documentation (ex: presets) as needed though.

##  Use Brand Colors

As part of implementing the `accent-color` property, we should take care to update our Form element examples, as well as consider the impact for component styling.

endigo9740:
## Linked Issue

Closes #4333

Resolves these issues:

Closes #3417
Closes #3517

## Description

Transition to the v5 Theme Format.

> NOTE: this skips the CLI and legacy Theme Generator

## AI Disclosure

## Checklist

## Changesets


packages/skeleton/src/base/theme.css
```diff
@@ -1,44 +1,45 @@
...
-	--anchor-font-weight: inherit;
-	--anchor-font-style: inherit;
-	--anchor-letter-spacing: inherit;
-	--anchor-text-decoration: inherit;
-	--anchor-text-decoration-hover: underline;
-	--anchor-text-decoration-active: inherit;
-	--anchor-text-decoration-focus: inherit;
+
+	/* Typography — Base */
+	--font-typo-base: system-ui;
```
Hugos68, line 7: What's the reasoning of abbreiviating `typography`? Just curious, I keep reading it like "typo" as in a typing mistake.

packages/skeleton/src/base/theme.css
```diff
@@ -1,44 +1,45 @@
...
 
 	/* Radius */
 	--radius-base: 0.25rem;
 	--radius-container: 0.25rem;
 
 	/* Edges */
-	--default-border-width: 1px;
-	--default-divide-width: 1px;
-	--default-ring-width: 1px;
+	--default-border-width: var(--default-border-width);
```
Hugos68, line 36: Isn't this a circular reference, what's the point of this?

packages/skeleton/src/base/theme.css
```diff
@@ -226,9 +227,38 @@
 	--color-surface-contrast-800: var(--color-surface-contrast-light);
 	--color-surface-contrast-900: var(--color-surface-contrast-light);
 	--color-surface-contrast-950: var(--color-surface-contrast-light);
+
+	/* Brand */
+	--color-brand-light: var(--color-primary-500);
```
Hugos68, line 232: Why? Seems like a bit of an opinionated addition, especially because it's just another var, so just an alias?

packages/skeleton/src/base/theme.css
```diff
@@ -226,9 +227,38 @@
...
+	--color-brand-dark: var(--color-primary-500);
 }
 
 @theme inline {
+	/* Typography — Base */
+	--font-typo-base: var(--typo-base--font-family);
```
Hugos68, line 240: What's the point of all these aliases if the variable is already available?

packages/skeleton/src/base/theme.css
endigo9740: `typography` is a long word. Most of the other canned names Tailwind uses for properties are rather short. I thought about `text`, but that would have had some conflicts with particular `@theme` properties. *Shrugs* naming things is hard heh.

I also kind of got a giggle about `typo` being a mistake. But if you can think of a better fit I'm all ears.

packages/skeleton/src/base/theme.css
endigo9740: Addressed here: https://github.com/skeletonlabs/skeleton/issues/4333#issuecomment-4200157518 (See the Edges category)
And solves this bug: https://github.com/skeletonlabs/skeleton/issues/3517

Per the prototype I linked, this was the only way I could get it to work as expected.

packages/skeleton/src/base/theme.css
endigo9740: Malte and I had a conversation about this the other day:
https://discord.com/channels/1003691521280856084/1010694301883318326/1491973089850818792

See this post and the one immediately following.

TL;DR: this provides quick means to set the expected singular most important brand color for use as the default for our TW/Functional components. Currently we always rely on primary-500, which doesn't scale for all company branding (sometimes you need a non-500 value) and locks you in for light/dark mode (this doesn't)

It'll be easier to see this in practice, so give me a bit of time to cook on this.

packages/skeleton/src/base/theme.css
endigo9740: - The `@theme` value is for the default Skeleton theme (the monochrome default theme)
- The `@theme inline` value is actually consuming the `data-theme` variable

This is a new change so I'm not 100% sure it'll work as I expect. But it's on my list to test tomorrow.

Hugos68: Perhaps the base/theme.css should be moved to src/themes/base.css? or mono.css? Would be less confusing between whats a default value and actually theme logic.

endigo9740: A few things here:

1. The default theme *is* just the default values; so it's not a true standalone theme like the others.
2. In a sense when you're running the default theme, you're running *theme-less*
2. When the new generator service is built, all Skeleton themes in the core package will move to that. No `.css` theme files will remain in the core package. This will help standardize the process for setting up themes, as all themes are then inserted from an external source. So `/themes` will go away before the v5 launch.

It's also really handy to have `@theme` and `@theme inline` defaults in the same place. It makes them easier to read as one builds on the other.

Hugos68 (approved):

PR #4370, opened 2026-04-17
Global style updates

### LINKED ISSUE #4334: Review and reduce global styles in the core

Currently Skeleton introduces a number of `@base` styles by default when importing and using Skeleton's core package.

## Goal

Find ways to remove or reduce the number of global styles.

~~For any styles that remain, consider implementing a opt-in approach to enabling these. This may introduce a new onboarding step, but this is likely a necessary change.~~ (we've opted to forego this)

## Task: Update Documentation

Update the relevant [Core API](https://www.skeleton.dev/docs/svelte/get-started/core-api#base) and [Onboarding](https://www.skeleton.dev/docs/svelte/get-started/installation) docs to account for this change.

### Provide relevant links, wireframes, prototypes, or additional information.

- Docs: https://www.skeleton.dev/docs/svelte/get-started/core-api#base
- Source: https://github.com/skeletonlabs/skeleton/blob/main/packages/skeleton/src/base/globals.css

endigo9740:
## Linked Issue

Closes #4334

## Description

- Updates and cleans up the `global.css` styles in the core package.
- Implements a new Design System > Globals documentation section covering these styles.

## AI Disclosure

## Checklist

## Changesets


.changeset/dull-chairs-teach.md
```diff
@@ -0,0 +1,6 @@
+---
+"@skeletonlabs/skeleton": minor
+---
+
+task: updates to the global styles
```
Hugos68: I think you should be a little more clear what's changed, this changeset will end up in the big list of V5, this doesn't really add anything so you can either remove it or explain it I think.

packages/skeleton/src/base/globals.css
```diff
@@ -1,8 +1,12 @@
...
 
 @layer base {
 	:root {
+		/* Properties --- */
+		/* Defines form field placeholder text styling. */
+		--global-field-placeholder: var(--color-surface-700-300);
+
 		/* Color Scheme --- */
 		/* https://tailwindcss.com/docs/color-scheme */
 		color-scheme: light;
```
Hugos68, line 12: Just a general wondering, why do we need to set this?

sites/skeleton.dev/src/content/docs/design/globals.mdx
```diff
@@ -0,0 +1,137 @@
...
+	background-color: color-mix(in srgb, var(--color-brand-light) 30%, transparent);
+	@variant dark {
+		background-color: color-mix(in srgb, var(--color-brand-dark) 30%, transparent);
+	}
+}
+```
+
+### Disabled State
+
+Implements defaults and utility class for handling the [disabled](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/disabled) status for elements. Used most commonly for buttons.
```
Hugos68: > Implements defaults


This reads a bit weird to me, although I'm a non native English speaker, should `defaults` be plural here?

Hugos68 (changes requested):

packages/skeleton/src/base/globals.css
endigo9740: Color pairings in Skeleton rely on a `light-dark()`, which is what [Color Scheme](https://tailwindcss.com/docs/color-scheme) controls (basically use left or right value). So we need to define a default color scheme for light and dark modes via the default/variant setting. Then, regardless of which mode you're using, the correct `light-dark()` values will be selected

sites/skeleton.dev/src/content/docs/design/globals.mdx
endigo9740: I agree. This would be better:

```
Implements the default disabled state and provides an optional utility class.
```

I'll keep "disabled" as the URL here.

.changeset/dull-chairs-teach.md
endigo9740: It's kind of hard to explain as most of it was just rearranging the order, pruning unused comments, fine tuning the existing styles. I also distributed the placeholder value to each Tailwind Component definition

So maybe:

```
task: refined global styles, redistributed placeholder color to Tailwind components
```

packages/skeleton/src/base/globals.css
Hugos68: Alright, I think this goes a bit beyond what I know, it just looks weird to have a `light` definition with `@variant dark`. But I don't think I quite understand the color scheme stuff exactly.

sites/skeleton.dev/src/content/docs/design/globals.mdx
Hugos68: Sounds good

.changeset/dull-chairs-teach.md
Hugos68: Fair if it's barely any changes, but just making sure it's not going to end up being a list of "task: fix up some small bugs" typa list.

packages/skeleton/src/base/globals.css
endigo9740: @Hugos68 think of it like this:

- `light-dark` takes in two color params: `light-dark({lightValue}, {darkValue})`
- Color scheme is the "switch" for controlling the active param
- In Tailwind's light mode (default) we explicitly set scheme to `light`
- In Tailwind's  dark mode we explicitly set the scheme to `dark`

All this is doing is augmenting Tailwind's existing light/dark mode system to play well with color schemes.

Schemes are useful because you can control them at any scope:

```html
<div>Inherits the global scheme</div>

<div class="scheme-light">Always LIGHT</div>

<div class="scheme-dark">Always DARK</div>
```

Hugos68: Ahh okay, so a native replacement for tailwinds v3 `dark` class?

endigo9740: TL;DR: It probably should be, but ONLY for color. The `light-dark()` function is ONLY intended to work with colors. While Tailwind can support *any* property because it's affecting a -mode- instead of a color theme. That's where it gets confusing!

Here's the long answer...

Color Scheme is a native CSS feature, but it didn't exist when Tailwind rolled their option. That's probably why Color Scheme support is limited right now. Adam and his team were likely trying to feel it out for v4.

Meanwhile, Tailwind's dark mode is a collection of features that work together to handle "mode" switching:

- Handle mode based on prefers-media, aka OS setting (this is the default behavior)
- Handle mode based on a class: `.dark`
- Handle mode based on a data-attribute: `[data-theme=dark]`
- Then provides a variant to handle overrides based on mode

Unlike Color Scheme, Tailwind has no concept of an explicit light theme. Default colors are light theme, only dark variant values override when dark mode is enabled.
 
 The Variant is the important bit. By default it uses: [prefers-color-scheme](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@media/prefers-color-scheme). Here's the source:
 https://github.com/tailwindlabs/tailwindcss/blob/main/packages/tailwindcss/src/variants.ts#L1143

But by implementing a custom variant of the same name, you can override it and make it use either the class-based, attribute-based, or combine both with a extra system option:
https://tailwindcss.com/docs/dark-mode#toggling-dark-mode-manually
https://tailwindcss.com/docs/dark-mode#using-a-data-attribute
https://tailwindcss.com/docs/dark-mode#with-system-theme-support

So the end goal is similar, basically handling styling based on light/dark theme.

- Color scheme is native, superior at colors, can handle isolated scope, but doesn't cover all properties
- Tailwind mode is native-ish (sans variant), cannot handle scope, and has a uniform system for all properties

So Skeleton actually uses BOTH. You control the overall theme with mode, but color scheme stays in sync. That's what the global styles you highlighted actually do.

Hugos68 (approved):

PR #4376, opened 2026-04-19
feature: Drizzle + Better Auth

Hugos68:
This PR converts our auth + database to drizzle and better auth rather than the Supabase SDK. Reasons being:
- Remove vendor lock: we can freely move to any postgres database without being stuck to Supabase.
- Database is code first, meaning we define the model in typescript and push to the remote database.


sites/plus.skeleton.dev/src/lib/features/auth/sign-in.remote.ts
TazorDE, line 1: just something to think about: I typically like to add another layer of folders in my remote functions folder to better sort them. Something like `remote > authentication > sign-in.remote.ts`

sites/plus.skeleton.dev/src/lib/infrastructure/database/schemas/auth-schema.ts
TazorDE, line 1: I'd like to make the recommendation to also sort the schema definitions into thematically appropriate folders and files.
You can just export them from the regular schema file like so:
```ts
export * from '../../../../auth-schema';
export * from './foo/bar.schema';
```

sites/plus.skeleton.dev/src/hooks.server.ts
```diff
@@ -1,26 +1,18 @@
-// @ts-ignore - prevent CI from failing due to missing env variables
-import { PUBLIC_SUPABASE_URL, PUBLIC_SUPABASE_PUBLISHABLE_KEY } from '$env/static/public';
-import { createServerClient } from '@supabase/ssr';
+import { auth } from '$lib/server/auth';
+import { svelteKitHandler } from 'better-auth/svelte-kit';
+import { building } from '$app/environment';
 import type { Handle } from '@sveltejs/kit';
 
 export const handle: Handle = async ({ event, resolve }) => {
```
TazorDE, line 6: We also might want to automatically run migrations on the database everytime the server starts up. I don't know how this works with preview environments, so thats something to figure out I guess.

TazorDE (commented):
I only have minor or suggestion comments. Looking good from the code side of it, but I've not checked out or run the code. 

If you have any more questions feel free to ping me here or on discord. (Discord gets faster response times :D)

sites/plus.skeleton.dev/src/lib/features/auth/sign-in.remote.ts
Hugos68: Yeah agreed, done

sites/plus.skeleton.dev/src/lib/infrastructure/database/schemas/auth-schema.ts
Hugos68: I do need to pass the full `schema` around, do you have another file that combines and exports the different schema files?

sites/plus.skeleton.dev/src/hooks.server.ts
Hugos68: Since it's a serverless environment we'd want to run it once per production deployment instead I think. Or just handle it manually at first.

sites/plus.skeleton.dev/src/lib/infrastructure/database/schemas/auth-schema.ts
TazorDE: I'm not quite sure what you mean, but each file you export from the central `schema.ts` is only a partial schema. You could have a user schema, a theme schema, a code block schema, and more, all with the relevant schema parts just for them. 
Just export them from the central file and everything should work as if they were written in the same file

Hugos68: Ah yeah I misread your original comment, that's probably wise idea to keep things structured, I've done so.

PR #4394, opened 2026-04-27
feature(skeleton-{react,svelte}): `QrCode` component

Hugos68:
## Linked Issue

Closes #{issueNumber}

## Description

Adds:
- QrCode component
- QrCode tests
- QrCode playground

## AI Disclosure

## Checklist

## Changesets


packages/skeleton-svelte/src/components/qr-code/anatomy/root.svelte
```diff
@@ -0,0 +1,37 @@
...
+		id: id,
+	}));
+
+	const attributes = $derived(mergeProps(qrCode().getRootProps(), rest));
+
+	RootContext.provide(() => qrCode());
+</script>
+
+{#if element}
+	{@render element(attributes)}
```
endigo9740, line 32: I know this component doesn't need a wrapping element to functionally work. But I'd suggest we add one to make it easier to handle a uniform width. Since size is not controlled automatically, then folks are going to reach for this intuitively, and be confused at why it's not working:

```html
<QrCode ... class="w-56">...</QrCode>
```

I've vote either a `<figure>` or `<div>` here.

Also make sure every demo has an explicit width defined in user-land as this should be convention. Never let it go full bleed. This will be ridiculously huge if we do!

packages/skeleton-common/src/components/qr-code.css
```diff
@@ -0,0 +1,19 @@
+[data-scope='qr-code'] {
+	&[data-part='root'] {
+		display: grid;
+		gap: --spacing(2);
+	}
+	&[data-part='frame'] {
```
endigo9740, line 6: Make sure to include an explicit example of modifying the color of the frame/pattern parts, as they require different and specific properties.

Maybe brand color for the frame, brand contrast for the pattern.

playgrounds/skeleton-svelte/src/routes/components/qr-code/+page.svelte
```diff
@@ -0,0 +1,11 @@
+<script lang="ts">
+	import { QrCode } from '@skeletonlabs/skeleton-svelte';
+</script>
+
+<QrCode value="https://github.com/skeletonlabs/skeleton">
+	<QrCode.Frame>
+		<QrCode.Pattern />
+	</QrCode.Frame>
+	<QrCode.Overlay>Overlay</QrCode.Overlay>
```
endigo9740, line 9: Follow Zag's lead on demoing the overlay feature, just use a Lucide icon + square background styling to help separate from the busy pattern. All in user-land of course.

Screenshot 2026-04-30 at 11 02 04 AM

Don't follow their lead on the poor choice of color contrast though :D

packages/skeleton-common/src/components/qr-code.css
```diff
@@ -0,0 +1,19 @@
...
+	}
+	&[data-part='pattern'] {
+		fill: var(--color-surface-950);
+	}
+	&[data-part='overlay'] {
+		background-color: var(--color-surface-900-100);
+		color: var(--color-surface-100-900);
+	}
+	&[data-part='download-trigger'] {
+		@apply btn preset-filled;
```
endigo9740, line 17: I was meaning to talk to you about the use of presets in our components. While I think `btn` is fine for adding form, I think presets may have gotchas. The reason being is preset classes don't "swap" like the used to, they override.

If you were to apply apply `class="preset-outlined-primary-500"` on this button as a user, you would expect a transparent background with a border edge of the colors defined. But what you get instead is a filled button with an outline added.

One of two things needs to happen:

1. We need to prune presets from the stylesheet and keep those in user-land only. A breaking change, but my recommendation.
2. We need to add an implicit `bg-transparent` on the outline presets.

I'd have real concern for number 2 as it may unintentionally break this pattern:

```
preset-tonal preset-outlined-surface-200-800
```

This would result in a tonal fill but outlined edge. Aka what we used to call "ghost" styling in v2. But if we add an explicit transparent bg, it may render this unviable.

playgrounds/skeleton-svelte/src/routes/components/qr-code/+page.svelte
```diff
@@ -0,0 +1,11 @@
...
+	<QrCode.DownloadTrigger mimeType="image/png" fileName="skeleton-qr-code">Download</QrCode.DownloadTrigger>
```
endigo9740, line 10: Just a consideration for another example - folks might request a version where you can click on the QR code to download it. Maybe try a variation with the frame/pattern wrapped by this button?

.claude/skills/create-framework-component/SKILL.md
```diff
@@ -0,0 +1,73 @@
...
+  - `modules/*` = context/providers/namespace composition
+  - `index.ts` = type exports + namespace export
+- Use `PropsWithElement` + `HTMLAttributes` in part prop interfaces.
+- Build attributes with `mergeProps` and render with element override fallback.
+- React anatomy files are `.tsx`; `index.ts` type exports reference `.jsx` paths.
+- Svelte anatomy files are `.svelte`; interfaces are in `<script lang="ts" module>`.
+- Namespace composition uses `Object.assign(Root, { ...parts })` in `modules/anatomy.ts`.
+- For static parts, include `data-scope` and `data-part` attributes.
+
+## Prompt flow
```
endigo9740, line 142: This looks awesome. Only bit of feedback is perhaps prompt users for a Zag documentation URL. Then notify the LLM here how to convert to the machine-friendly version:

- From: https://zagjs.com/components/{framework}/{component}
- To: https://zagjs.com/api/mdx/components/{framework}/{component}

You might even try and include the Github source for the component using a similar pattern, so it can have a deep level of understanding of how this operates.

Just to make this an explicit part of the user<->llm transaction upfront.

.claude/skills/create-functional-component/SKILL.md
```diff
@@ -0,0 +1,73 @@
...
+- Namespace composition uses `Object.assign(Root, { ...parts })` in `modules/anatomy.ts`.
+- For static parts, include `data-scope` and `data-part` attributes.
+
+## Prompt flow
+
+Ask one at a time, confirm each, write nothing until all are answered and the user confirms.
+
+1. **Component name** — kebab-case folder and PascalCase namespace. Recommend both, accept overrides.
+2. **Type** — static/simple or machine-backed (Zag). This resolves required module files.
+3. **Parts list** — root-only or full anatomy list (`root`, `trigger`, `content`, etc.).
```
endigo9740: Step 3 and 4 feels like something the LLM could likely suggest for us if provided the documentation mentioned above. Just accept user overrides.

.claude/skills/create-functional-component/SKILL.md
```diff
@@ -0,0 +1,73 @@
...
+2. Keep export ordering consistent with neighboring component entries.
+3. Add tests in both frameworks:
+
+- React fixture: `packages/skeleton-react/test/components/<component>.tsx`
+- React suite: `packages/skeleton-react/test/components/<component>.test.tsx`
+- Svelte fixture: `packages/skeleton-svelte/test/components/<component>.svelte`
+- Svelte suite: `packages/skeleton-svelte/test/components/<component>.test.ts`
+
+4. For each anatomy part, include a render assertion (`toBeInTheDocument`).
+5. If the component should be previewable in the playgrounds, add matching demo entries in both frameworks and regenerate the React route manifest if a new route was added.
```
endigo9740: Nothing prompts the user if playground should be created, so might be worth being a default on, user opt-out?

.claude/skills/create-playground-entry/SKILL.md
```diff
@@ -0,0 +1,79 @@
...
+Reference files:
+
+- Simple: [playgrounds/skeleton-react/src/routes/components/accordion/index.tsx](playgrounds/skeleton-react/src/routes/components/accordion/index.tsx), [playgrounds/skeleton-svelte/src/routes/components/accordion/+page.svelte](playgrounds/skeleton-svelte/src/routes/components/accordion/+page.svelte)
+- Multi-section: [playgrounds/skeleton-react/src/routes/components/carousel/index.tsx](playgrounds/skeleton-react/src/routes/components/carousel/index.tsx), [playgrounds/skeleton-svelte/src/routes/components/carousel/+page.svelte](playgrounds/skeleton-svelte/src/routes/components/carousel/+page.svelte)
+
+## Prompt flow
+
+Ask one at a time, confirm each answer, write nothing until all are answered and the user confirms.
+
+1. **Component slug** - kebab-case route segment (for example `floating-panel`).
```
endigo9740, line 61: When folks are using LLMs they're going to likely use natural language and write "Floating Panel" or "floating panel" or " I want to add a floating panel component". So allow any input, but let the LLM coerce to specific kabab format here. Again user confirmation is good.

Same deal for the other skill btw.

endigo9740 (changes requested):
@Hugos68 Overall everything looks great. Just a few suggestions here. Let me know if you have any questions or feedback. Thanks bud!

packages/skeleton-svelte/src/components/qr-code/anatomy/root.svelte
Hugos68: It's already a `div`.

packages/skeleton-common/src/components/qr-code.css
Hugos68: Let's settle this in a different issue/PR, otherwise this PR will go beyond it's scope.

Hugos68: I've created an issue here: https://github.com/skeletonlabs/skeleton/issues/4466, resolving this comment.

sites/skeleton.dev/src/content/docs/framework-components/qr-code.mdx
```diff
@@ -0,0 +1,127 @@
+---
+title: Qr Code
+description: Render scannable QR codes with a frame, overlay, pattern, and download trigger.
+---
```
endigo9740, line 5: Missing `stability: beta`. Best to have this at launch for a little wiggle room in case something goes wrong.

sites/skeleton.dev/src/components/examples/framework-components/qr-code/svelte/clickable-download.svelte
```diff
@@ -0,0 +1,11 @@
+<script lang="ts">
+	import { QrCode } from '@skeletonlabs/skeleton-svelte';
+</script>
+
+<QrCode value="https://skeleton.dev">
+	<QrCode.DownloadTrigger fileName="skeleton-dev" mimeType="image/png">
```
endigo9740, line 6: Shows as extremely small, some kind of sizing issue (both frameworks)

I might suggest wrapping in a div as the Preview styles to center elements may be affecting this. That or give this a fized size rather than responsive.

sites/skeleton.dev/src/components/examples/framework-components/qr-code/svelte/frame-colors.svelte
```diff
@@ -0,0 +1,10 @@
+<script lang="ts">
+	import { QrCode } from '@skeletonlabs/skeleton-svelte';
+</script>
+
+<QrCode value="https://skeleton.dev">
+	<QrCode.Frame class="size-full max-size-36 bg-brand-contrast-dark">
```
endigo9740, line 6: Shows extremely small, doesn't look like size 36 (both frameworks)

sites/skeleton.dev/src/components/examples/framework-components/qr-code/svelte/overlay.svelte
```diff
@@ -0,0 +1,13 @@
+<script lang="ts">
+	import favicon from '@/assets/favicon.png';
+	import { QrCode } from '@skeletonlabs/skeleton-svelte';
+</script>
+
+<QrCode value="https://skeleton.dev">
+	<QrCode.Frame class="size-full max-size-36">
```
endigo9740, line 7: Only shows the icon, no QR code (both frameworks)

endigo9740 (changes requested):
Make sure to add a changeset too!

endigo9740 (approved):

PR #4410, opened 2026-05-08
task: modularize the component doc generation

Hugos68:
This PR moves all the generation logic for our types to a seperate package to remove all the clutter from the main docsite. This also preps our documentation for possible MCP use cases.
image


Hugos68:
@endigo9740 I'm assuming you don't mind this change, which just creates a more clear layer of seperation of concerns and lays the foundation for future generated documentation (the script in the docsite was becoming really large and unmaintainable).

All I really would like to know from you is what name we give this package, I used `@skeletonlabs/docs` as it provides .json files that act as documentation. But if you want to keep that name reserved for a future project we can adjust it.

